PR #381, opened 2024-07-10
Watch deposited assets

### LINKED ISSUE #344: Add token contract address via EVM wallet
As a user, I want to receive a notification via Metamask (via prompt send) to add the token contract to my wallet for any tokens interacted with on the Hemi Network.

Current token list to be integrated in issue:

- [ ] USDC.e
- [ ] USDT.e
- [ ] DAI
- [ ] ETH (tunneled ETH)
- [ ] tBTC (tunneled tBTC)

gabmontes:
After a successful deposit of ERC-20 tokens using the tunnel, the wallet will be asked to add that asset. So this change closes #344.

And, as an encore, the ESLint rule [`@shopify/prefer-early-return`](https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/docs/rules/prefer-early-return.md) was activated. It may be added to `eslint-config-bloq` soon.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
       ],
       "devDependencies": {
         "@commitlint/cli": "18.4.3",
+        "@shopify/eslint-plugin": "^45.0.0",
         "@types/mocha": "10.0.6",
         "@typescript-eslint/eslint-plugin": "6.16.0",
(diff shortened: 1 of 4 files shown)


gabmontes:
@gndelia we should close this PR, right?

gndelia:
Yes, closing

PR #445, opened 2024-07-31
Replace DetailedHTMLProps type usage

gndelia:
I just learned about `ComponentProps` which works for both native and user components, and it is more readable than using `DetailedHTMLProps`  for accessing props for HTML native components... So this tiny PR updates the few usages of it.

(The only "downside" is that it seems it is not directly accessible from `React.ComponentProps` - it requires a named export)

No visual changes for the user

### DIFF
--- a/ui-common/components/button.tsx
+++ b/ui-common/components/button.tsx
@@ -1,3 +1,5 @@
+import { ComponentProps } from 'react'
+
 const sizes = {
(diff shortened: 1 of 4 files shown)


ArturDolzan (approved):
woow, much better to read

gabmontes (approved):

Issue #456, opened 2024-08-08
Recovering past withdrawals fails to retrieve their status


gndelia:
The blame is on me! In this commit https://github.com/hemilabs/ui-monorepo/commit/c714419d3fdf4e49036b32ad56065d64a0895fa8 I removed the line `!w.status` because that would exclude `status: 0` (Which is correct), but I didn't consider that `status` may be undefined. So we need to consider this scenario.

When `status` is undefined? When reloading the past history by bringing past withdrawals.

So the check for `status === undefined` must be added for the `WithdrawalsStatusUpdater`

Issue #455, opened 2024-08-08
Read contracts from hemi-viem


gndelia:
[hemi-viem](https://github.com/hemilabs/hemi-viem/blob/master/src/chains/hemi-sepolia.ts) has the list of contracts we currently have in our `.env`.

We should remove them from the `.env` and read them from `hemi-viem` (although allowing configuration per environment variables is valuable for working with forks)

Depends on https://github.com/hemilabs/hemi-viem/issues/16

Issue #457, opened 2024-08-08
Switching page shows a grey loading icon in the Transaction History


gndelia:
Just switch any page in the Transaction history, and there's a big loading section (And the scrollbar becomes visible briefly)

https://github.com/user-attachments/assets/dbb75582-380b-425d-b074-ba26b8bc6e95

note: this is only visible in the prod build, not when running locally with the unmodified version

Issue #345, opened 2024-06-13
BTC tunnel transactions history


rmilrad:
As a user, I want to see all the tunnel transactions (deposits & withdrawals) from within the transaction history tab in chronological order of completion.

PR #458, opened 2024-08-08
Several changes for the Transaction History refactor

gndelia:
This PR is easier to review by commit. I took some changes I'm doing as part of refactoring the Transaction History to welcome the Bitcoin history 😄  and I also fixed some stuff along the way.

Related to #345 

## Changes

- 26e7ac57b021dac420a6bcb5427b0834a7561a62 The code used to create the providers was moved into its own file  under the `utils` folder
- a144e0c28700aeda395bdab0bb4fbed01710addd I added the `useConnectedToSupportedChain` hook. This will be used in a future PR
- 0d70c5550892aeaf50620ad97514bd4164d89526 The configuration for syncing chains (with MinBlockNumber or blockWindowsSize) was moved into its own file
- 8e2879c1a86d74a411e164dc01d5706e12ca8fee Adds a method to check if a chain is an `L1` chain. This will be used in a future PR
- 41e4f28e1e4abf22a10148bf237ddbd51f2ecb8c fixes a bug in the withdrawal status updater. When recovering past transactions, its status is `undefined`, and I accidentally ignored those in [this commit](https://github.com/hemilabs/ui-monorepo/commit/c714419d3fdf4e49036b32ad56065d64a0895fa8). This commit fixes it again. Closes https://github.com/hemilabs/ui-monorepo/issues/456
- ac148be7f407adea36efe547c514b0b54c6fd730 This commit removes some code that made no sense: When resetting the state of the queries after a deposit/withdrawal was successful, we were calling `invalidateQueries` on the `mutationKey`, which had no effect because exactly before that call, we were calling `mutation.reset()` which already clears that information. The correct call would be calling `invalidateQueries` on the `queryKey` from `useWaitForTransactionReceipt.`. This is now fixed and removes all the  `mutationKeys` that were exported in the hooks as they are now are useless.
- 47de62ce9538db8880da1ba06d58bd7a584b88d0 Reads the tunneling contracts from `hemi-viem`, but allows to override them through environment variables when needed. Once we stop using forks, we could drop this feature as well. Because of this change, we can now drop the addresses from the `.env` file. Closes https://github.com/hemilabs/ui-monorepo/issues/455

I also created https://github.com/hemilabs/ui-monorepo/issues/457 as part of this PR

No changes visible to the user


gabmontes (approved):
uACK

alvarengathomas (approved):

Issue #464, opened 2024-08-12
Toggling the Tunnel form breaks the app


gndelia:
(This only happens on staging)

## Steps to reproduce

1. Enter the app
2. You should see the Deposit operation by default. Try to toggle to enter "Withdraw" mode
3. The app breaks
 
<img width="1900" alt="image" src="https://github.com/user-attachments/assets/b9976bc6-01f8-4b03-b7df-58d2ad5e642b">

This issue seems to have been introduced by https://github.com/hemilabs/ui-monorepo/commit/47de62ce9538db8880da1ba06d58bd7a584b88d0

PR #465, opened 2024-08-12
Add workaround for accesing l1ChainId

gndelia:
As part of removing many hardcoded chain Ids, there is a scenario where react renders an invalid state. In order to prevent the app from breaking, this commit adds a workaround until a proper fix is found

Related to https://github.com/hemilabs/ui-monorepo/issues/464

This issue became evident after https://github.com/hemilabs/ui-monorepo/commit/47de62ce9538db8880da1ba06d58bd7a584b88d0

App is broken in staging only


ArturDolzan (approved):
utACK

PR #466, opened 2024-08-13
Replace useQueryParams with nuqs

### LINKED ISSUE #464: Toggling the Tunnel form breaks the app

gndelia:
Closes #464 

#### TL;DR

The implementation of `useQueryParams` was buggy, it worked at the beginning but as the app grew, it started to fail and render inconsistent states. Replacing it with [nuqs](https://nuqs.47ng.com/) (which is more robust) fixes the problem

#### The longer version

The tunnel form used a query string (`operation`) to determine which flow (either `deposit` or `withdraw`) the user sees. From that, each corresponding component was rendered (`Deposit.tsx` or `Withdraw.tsx`).  
In addition, a `tunnelState` is implemented in  a `useReducer` which holds all the state for the form: The selected tokens, the amounts, selected networks, and all that.  
When clicking the `toggle` option, that would trigger a callback that would do something like (simplified):

```tsx
// Assuming the user was on operation 'Deposit'
updateQueryString('operation', 'withdraw')
toggleStateOnReducer()
```

`updateQueryString` would replace the query string with `?operation=deposit` to `?operation=withdraw`
`toggleStateOnReducer` would invert the state on the reducer: The `fromToken` would now become the `toToken`, and the original `toToken` is now the new `fromToken`. The same is applied to the `fromNetwork` and `toNetwork` properties.

The `operation` lived on the query string, and was read across the app through `useTunnelOperation` - a custom hook that returned the `operation` value from the query string.

However, it seems that due to the way `React` batches the updates, the state from the `useReducer` updated before the query string changed (even though this function ran first!)

This caused the app to render invalid states: for one or two renders, `operation` would equal `deposit` (as if `updateQueryString` hadn't run), but the state was already updated to the one that `withdraw` expected! (as if `toggleStateOnReducer` did run before `updateQueryString`) After those one or two renders, `operation` would be updated to `withdraw` and everything went smoothly.

For most of the time, this wasn't a problem because we weren't using the (wrong) network or tokens for anything during the render body of the `withdraw.tsx` component - they were only used on handlers reacting to user actions - so they would eventually be correct (once React rendered the proper state). 

.... Until this commit https://github.com/hemilabs/ui-monorepo/commit/47de62ce9538db8880da1ba06d58bd7a584b88d0 was merged.

This caused the App to use `fromNetwork` during the render (to get the tunnel contracts for reading the allowance of ERC20 tokens). This was only needed on `Deposit` (where we need to approve ERC20 tokens before deposit), but it was trying to get the TunnelContracts from the `FromNetwork` that the `withdraw` state had, which was Hemi.

#### Enter nuqs

React usually batches state updates together to avoid invalid render states. So when doing

```tsx
// Assuming the user was on operation 'Deposit'
updateOperationOnState('operation', 'withdraw')
// React doesn't do anything here
toggleStateOnReducer()
// Now that there are no more updates ,React renders with both state changes!
```

the invalid state isn't rendered at all. The trick was now to handle the operation through the state (so updates are batched together), and then somehow sync it to the query string (so the state actually lives there as it was before)

Instead of updating `useQueryParams`, I found then `nuqs` that provided all the functionality we had (with parsing and TS support), it has better support, and allowed me to handle this as a state (and automatically sync the state to the query string). This way, update states are batched together, and no invalid states are rendered 🎉


ArturDolzan (approved):

gabmontes (approved):
uACK

PR #467, opened 2024-08-13
Fix allowance key not clearing

gndelia:
A tiny bug I found while working with #466   
When approving an ERC20 token, the `allowance` should be cleared. However, that never happened because I was incorrectly checking the status of the Transaction Receipt 🤦🏽


ArturDolzan (approved):

gabmontes (approved):
😅

Issue #462, opened 2024-08-12
Improve handling of tesnet/mainnet Ids


gndelia:
With the (eventual) coming of mainnet, there are lots of places that have a `hemi.id` hardcoded that comes from an import statement. However, there a high chance that the determination to use a `tesnet` or `mainnet` comes from a switch/toggle button, so both environments must live together. Because of that, we need to read the proper `chainId` instead of directly importing it.

Issue #390, opened 2024-07-12
Research web workers as an alternative to update transaction history


gndelia:
Currently, we have some components (with no UI) that check the current history and update the status of the Transactions. 

These requests take place in the background of the user UI and make the app constantly work even though the user is not doing anything. The updates are written to local storage.

One option could be to use web workers so all the work and processing could be handed off to a background process, that can periodically post messages to the user app to record the new information in local storage. As web workers are running in a separated thread of the browser, they will remove part of the heavy lift of querying tons of APIs.

Review if this approach is possible, as it may  separate the process of syncing transactions of the proper app (Another benefit: This will run outside of React, so it will reduce part of the complexity greatly)

gndelia:
Web workers were implemented to sync past deposits/withdrawals for Sepolia, and Deposits for bitcoin. This is also in production, so I'm closing

Issue #326, opened 2024-06-03
Sometimes the tx-history resyncs all over


gndelia:
Tx-history should sync from the current block to the last synced block. However, and particularly about the withdrawals, sometimes it just resyncs everything. As data is still kept, it doesn't erase the history, but it walks through blocks it has already reviewed.

gndelia:
This is no longer happening. It must have been fixed by PR #468

PR #468, opened 2024-08-13
Sync deposits and withdrawals (TX-History) in the background with Web Workers

gndelia:
Unfortunately, this PR extended more than I intended, but as I was working on it, some fixes were needed and I was deep enough in the rabbit hole that I couldn't split it into different PRs. Regardless, I hope the explanation below gives a general idea of the changes.

## Description

As part of preparing the terrain for syncing the history of past Bitcoin deposits and withdrawals, and also for Hemi Mainnet, this PR refactors the code used to sync deposits and withdrawals in the pair Sepolia/Hemi Testnet.

The main change is in the architecture, as the previous code (Based on the implementation in Pure Finance) ran everything in a custom hook. This caused several issues, such as being coupled to `useEffect` functions which made it difficult to follow, as well as being really difficult to test.

The new approach (following the description of the issue https://github.com/hemilabs/ui-monorepo/issues/390) intends to move all the logic used to sync past operations to a process running in the background, using [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers). This implementation would free up the main thread where the Portal (And the tunnel) are running.

The following diagram represents the implementation

worker-refactor

## Benefits over the previous implementation

- The code is now decoupled from the UI - it is easier to follow and should be easier to test with some automated tests
- The code doesn't depend on `useEffects` and its weird quirks
- The code is split into several different functions with different responsibilities (saving to storage, calling providers, syncing in blocks, etc)
- Users can now switch chains (from Sepolia to Hemi Testnet or the other way around), and the sync process won't get lost, and it will continue syncing as usual! 
- Related to that item, with the OG implementation, there was a strange bug in which it would first sync all deposits, and then all withdrawals, or first all withdrawals, and then all deposits, depending on which chain the user was connected. Now, they are both synced in parallel.

There are some improvements applied as part of this change (that weren't a strict part of using Web Worker, but it was facilitated by the mere refactor)

- Throttling is now shared between deposits and withdrawals because they use the same CrossChainMessenger instance
- Users can now sync their history as soon as they connect their wallet, instead of needing to explicitly visit on the transaction history page. So now they can navigate through the portal and sync their info! (With the prior implementation, users had to be on the Transaction History page and it would only sync if the user was there, which was weird)
- I also think I've fixed the issue in which history would resync all over (See https://github.com/hemilabs/ui-monorepo/issues/326). I'm leaving it open for further testing, but it should be fixed now.
- Better debug information in the console.

With all that in mind, this is a summary of some decisions made along this PR

- I created the `sliding-block-window` package, as mentioned. I added tests for this package (yay!), but I faced an issue: The implementation code was on Typescript, but I wanted to write tests on Javascript because it allows me to write simpler tests, as I don't need to send all the arguments the type requires (only the info I need for specific tests).  `mocha` (the test runner we've long used),  doesn't have native support, and requires to use `ts-node` and a special configuration for compiling the imported Typescript on the test on the fly... but that also requires converting the package into an ESM module (by adding `"type": "module"` in the package.json). But doing so breaks `nextjs`'s automatic transpiling of other packages! For me, that meant updating the next config to transpile it to JavaScript as a build step. It seemed that was a lot. Particularly, I didn't want to update any consumer of the package (We can convert it to an ESM module if we publish the package because in that case, the build step happens at pre-publish time). So I solved it by switching to `vitest` which had native support for mixing Typescript and Javascript code (You can check the [trend comparison against mocha](https://npmtrends.com/mocha-vs-vitest). The only downside is that its expectations API is similar to `Jest`'s, which is not what we may be used to using. Other than that, it worked very well. If we publish this package in another repo, we could rollback to mocha if needed.

- Instantiation of `crossChainMessenger` was moved onto its own file, and combined with `p-queue`, as throttling can only be set per method at most (The class `CrossChainMessenger` uses some prototype implementation which made it difficult to use `p-throttle` on instance level, but It does work on method level). I also tried using Proxies but they did not work with the Promises (they are better suited for get/set sync accesses)

- The local storage key used to save the sync info was updated: It now includes both the L1 and L2 Ids. This was needed because there would be no easy way to distinguish between Hemi Testnet and Hemi Mainnet (as we only had the L1 id). There could be ways to assume that (for example: Sepolia can't deposit on Hemi Mainnet!), but this makes it easier. A fallback was implemented so the previously stored info goes into the newest key.

- That also means that in each Deposit / Withdrawal, I am storing the L1ChainId and L2ChainId to later track easily where an operation belongs to

- Web workers were added, and I added an env variable to enable logs if needed (they don't work by default). A web worker is created per pair L1-L2. So for instance, a worker is created to sync Sepolia/Hemi, and another worker will be created to sync Bitcoin/Hemi

- The Sync status is nos simplified because I merged the `depositSyncStatus` and `withdrawSynsStatus` into just `status` - I don't need both.

- The `useSyncInBlockChunks` hook was dropped 🎉 

- With the reducer implementation it was easier to have generic functions as `addDeposit` or `updateDeposit`, as part of the optimistic UI changes when sending TXs - we no longer need `addEvmDeposit or `addBtcDeposit` which were functions separated per chain (That was because the `useSyncInBlockChunks` used different states and made it difficult to merge them)

Issues created/found during the development of this PR:

- https://github.com/hemilabs/ui-monorepo/issues/464 (Workaround opened in https://github.com/hemilabs/ui-monorepo/pull/465)
- https://github.com/hemilabs/ui-monorepo/issues/462
- https://github.com/hemilabs/ui-monorepo/issues/457

## Further improvements

- After long periods, the RPCs may face some rate limiting, which ends up on rejected requests. Ideally, a dynamic throttling could be applied, in which older blocks could be requested with a larger block range size between requests, as that information is probably less useful for the user - they do care about the latest information. 

- Perhaps more RPCs per chain could be added, to have fallbacks when a provider fails, or splitting the requests among them.

- The `sliding-block-window` package could be published and used in PF to also sync token revokes. The main idea of this package is allowing to track the range of blocks and run an async function in each block range. It doesn't care if you're running this on a Script, React or how you store the information. It's just a small package to help you sync information in a range of blocks.
- Once the sync process is finished, we could poll for incoming new blocks and keep syncing in real-time.

## Final notes

I'm not closing the issue  https://github.com/hemilabs/ui-monorepo/issues/390 because I think it could be used to also move the [withdrawal status updater](https://github.com/hemilabs/ui-monorepo/blob/main/webapp/context/tunnelHistoryContext/withdrawalsStateUpdater.tsx) and the [bitcoin deposits status updater](https://github.com/hemilabs/ui-monorepo/blob/main/webapp/context/tunnelHistoryContext/bitcoinDepositsStatusUpdater.tsx) into web workers.


gabmontes (approved):
uACK

ArturDolzan (approved):
utACK

PR #472, opened 2024-08-19
Sync bitcoin history deposits

gndelia:
Depends on #468 being merged first
Related to #345 

This PR allows us to bring past Bitcoin deposit operations as part of resyncing in the transaction history. A change in comparison to #345 is that I am now tracking the sync status (ready, syncing, finished, etc) per chain, instead of as a global enum, but I also calculate a global sync status by checking all chains.

I also had to use Mempool Api, which uses pagination per Tx [See docs here](https://mempool.space/docs/api/rest#get-address-transactions) - which means I cannot paginate the Deposits with the From/To blocks strategy.

### Conceptual implementation

To bring past deposits, what I do is bring all TXs involved with the BTC addresses of the vault (so far there's one, in the future it should include all vaults), and filter the TXs where that custody address is:

- The destination (And not the origin)
- it has a UTXO which is OP_RETURN and
- and I check that the above UTXO has a script pub key of `OP_RETURN OP_PUSHBYTES_40 ${Hemi-Address}`, where the HemiAddress needs a few transformations (it gets converted into uppercase, then from string to hex, removing the `0x` part)

For those, we can calculate the deposit amount by filtering the destination UTXOs (`vout` field), which have the Script pubKey address set to the bitcoin custody address, and add up all those satoshis.

From there, the rest is to track the last TX used in the query, so the next time we don't request them all again.

Similar to the previous PR, the worker just sends batches of newly found TXs and these are offloaded into Local Storage.

The Next step is to sync withdrawals.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28573,6 +28573,7 @@
         "next-recaptcha-v3": "1.3.0",
         "nuqs": "1.17.8",
         "p-all": "5.0.0",
+        "p-do-whilst": "2.0.0",
         "p-queue": "8.0.1",
         "p-throttle": "6.1.0",
         "promise-mem": "1.0.2",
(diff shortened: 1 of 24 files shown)


gabmontes (approved):
uACK

ArturDolzan (approved):

Issue #105, opened 2024-03-06
Enable "strict" option in Typescript


gndelia:
Because it is recommended and needed for getting typings right in the hook that converts ethers' provider to Wagmi. However, doing so, requires review a few changes that I can't do right now 😄  I'm creating the issue as part of working with #37

gndelia:
I think we could improve each package `tsconfig.json` by using [tsconfig/bases](https://github.com/tsconfig/bases) as a reference. So we can drop our [tsconfig.base.json](https://github.com/hemilabs/ui-monorepo/blob/main/tsconfig.base.json)

PR #475, opened 2024-08-20
Minor improvements strict mode

gndelia:
This PR adds a few minor fixes in Typescript that appear if we compile with `strict: true` in the `tsconfig.json` file (See https://github.com/hemilabs/ui-monorepo/issues/105).  These fixes are in ad3f6cd3ef6742c5b5640348d494b7bc10ac185d

In addition to that,  I am also removing in 2dbbce09195ffdade592a5121176adfb9dfc5193 `@radix-ui/react-slot` as we weren't actually using it - it was just written as a conditional prop that was never actually used. This will (Very) slightly improve installation times, compilation times, and reduced bundle size (barely noticeable, but it saves a few KBs)

### DIFF
--- a/btc-wallet/context/globalContext.tsx
+++ b/btc-wallet/context/globalContext.tsx
@@ -10,7 +10,12 @@ type GlobalState = {
   setCurrentConnector: (connector: WalletConnector | undefined) => void
 }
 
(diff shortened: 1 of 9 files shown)


gabmontes (approved):
LGTM

ArturDolzan (approved):

Issue #376, opened 2024-07-04
Remove hardcoded L1ChainIds values


gndelia:
As part of enabling Bitcoin deposits (See #342 and #345), we had to add the `chainId` field to the Transaction History stored in local Storage. This field wasn't part of it originally, so reading that data would cause an error if the code assumes it was defined. As a fix for that was implemented in TBD PR, we left a series of defaults when this value was defined. As just entering the Tx history for every user should automatically fix this problem (saving all deposits + withdrawals with the corresponding chainId), we can remove these defaults in a few weeks, or in the worst case, just pre-launch.

TL;DR remove hardcoded default chainIds where this issue was referenced

PR #481, opened 2024-08-23
Get hemi id from querystring

gndelia:
As part of preparing the code for `mainnet`, this PR removes importing `hemi` in many places and adds a new hook which, reading from the query string, returns the appropriate `hemi` instance.

There are a few scenarios where `hemi-sepolia` is directly used as getting the hook could be complicated and there are temporary solutions anyway (See https://github.com/hemilabs/ui-monorepo/issues/376 )

The next step is getting all the networks / evm networks and other chain arrays based on this setting. 

Related to https://github.com/hemilabs/ui-monorepo/issues/462

### DIFF
--- a/webapp/app/[locale]/demos/page.tsx
+++ b/webapp/app/[locale]/demos/page.tsx
@@ -1,13 +1,33 @@
 'use client'
 
-import { hemi } from 'app/networks'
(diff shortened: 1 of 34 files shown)


gabmontes (approved):
uACK

ArturDolzan (approved):

PR #484, opened 2024-08-26
Get networks query string

gndelia:
Closes https://github.com/hemilabs/ui-monorepo/issues/462

Follow-up PR of #481, this PR updates the codebase so the list of networks (and bitcoin) is calculated based on `testnet` or `mainnet` defined by query string. For that, a few hooks were added

- `useNetworks()` returns the different combination of networks given the query string (`remoteNetworks`, `evmRemoteNetworks`, etc)
- `useBitcoin` returns the instance of bitcoin (testnet or mainnet)
- `useChainIsSupported`, based on `useNetworks`, returns true or false depending on the chain being supported or not

Some type definitions (like `EvmChain` and `RemoteChain`) were moved into `types/chain`, as well as some utilities into `utils/chains` (For the cases where they couldn't be solved with hooks)

Because of that, `networks.tsx` was renamed to `network.ts` (Removing the `x`, as there were no longer React components rendered - the Hemi Logo was moved to `useHemi`)

This PR may be easier to review per commit.

There are no visible changes for the user


gabmontes (approved):
uACK

PR #488, opened 2024-08-27
Improve cross chain messenger concurrency

gndelia:
This PR is easier to review by commit.

After some testing in production with the web workers, this adds some tweaking to avoid overloading RPC calls when syncing past operations:

- 8d2a6445f704cf6020563328cd63c079cbbe9514 Adds better throttling when syncing EVM operations, tweaks a little bit the concurrency numbers after some testing, and also renames the functions used to instantiate `cross-chain-messenger`. Plus, now the hooks to deposit/withdraw do not share the global queue that syncing uses, so depositing/withdrawing is not delayed by syncing operations. (syncing will be a bit slower with the throttling, but as it happens in the background while the user is on the page, there are no hurries...). For the future, throttling should probably be solved at RPC level - See https://github.com/hemilabs/ui-monorepo/issues/489)
- bee0abaf88b32f5e03bd443161c3cf7baba32629 Updates some ticket references to move more sync code to web workers
- e192e9d853dba4548c3f1fb43f87d43e4c4c467b Updates references from old issues to the appropriate ones!

### DIFF
--- a/webapp/app/[locale]/tunnel/_hooks/useTunnelState.ts
+++ b/webapp/app/[locale]/tunnel/_hooks/useTunnelState.ts
@@ -199,7 +199,7 @@ const getDefaultNetworksOrder = function ({
   if (!['claim', 'deposit'].includes(operation)) {
     // for non-deposits, the withdrawals work ok
     // Needs to be updated once btc withdrawals are enabled
-    // https://github.com/BVM-priv/ui-monorepo/issues/343
+    // https://github.com/hemilabs/ui-monorepo/issues/343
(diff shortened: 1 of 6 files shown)


ArturDolzan (dismissed):

gabmontes (dismissed):
uACK

gndelia:
@gabmontes @ArturDolzan review requested again due to rebase to solve conflicts

gabmontes (approved):

ArturDolzan (approved):

Issue #440, opened 2024-07-30
Allow users to resync the Transaction history manually


gndelia:
Under certain circumstances (Such as a chain reorg), a user may end up with operations in the TX-history that are no longer valid. We need to find a way to remove those operations from the history, as they add noise. One option is to resync everything, so invalid operations are automatically discarded.

rmilrad:
@gndelia Do we need a bit of UI here to complete this issue? (cc @dvnahuel )

gndelia:
yes, that's why Nahuel is assigned

dvnahuel:
What I suggest and discussed with @gndelia is to add the 'Reload' CTA in the transaction history table. This will allow users to reload the items, and in doing so, the invalid operations will be automatically discarded. After the reload button is clicked, a toast notification will inform users that some of the items in the history may have disappeared, indicating they were invalid (text still WIP).

gndelia:
@dvnahuel @rmilrad is "text sill WIP" as indicated in the last comment? Otherwise, I could take this issue

dvnahuel:
> @dvnahuel @rmilrad is "text sill WIP" as indicated in the last comment? Otherwise, I could take this issue

Indeed! designs [here](https://www.figma.com/design/9A3yGxPovTEjn3OQ3Uterc/Hemi-Marketing?node-id=4529-6040&t=jPph30BDVqf77u0v-1)

PR #492, opened 2024-08-29
Allow resync tx history

gndelia:
Closes https://github.com/hemilabs/ui-monorepo/issues/440

This PR adds a button to resync back the history. This is useful when, for some reason, the history is in a bad state and needs to be solved again.

This PR may be easier to review commit by commit:

- 1599ebb20f5ef0a473de3b442e583b6f1a9f4caf The reducer of `useSyncHistory` is moved into its own file (`reducer.ts`)
- 789896c54f28a3041c5493b20003cce640e28dee moves the definition of the Tunnel History context inside of `useSyncHistory`. After all, all the functions shared in the `TunnelHistoryContext` are based on the state/dispatch from the `useReducer` in `useSyncHistory.ts`
- 97e652136b8c1101bd01ac971fa76093e58e25f5 adds the code that allows for resync. The fallbacks were removed because they would conflict when clearing the local storage. It is not a problem because, after all
- 0cfccdabb7aa4303a906ad8df49a53b8040da1d6 applies the UI changes to resync the history

Resync can be performed even when the TX history hasn't finished resyncing.

See video:

https://github.com/user-attachments/assets/7dd93790-8dc4-4f44-8468-2b9405ee0b4f

### DIFF
--- a/webapp/app/[locale]/tunnel/transaction-history/_components/reloadHistory.tsx
+++ b/webapp/app/[locale]/tunnel/transaction-history/_components/reloadHistory.tsx
@@ -1,0 +1,29 @@
+import { useTunnelHistory } from 'hooks/useTunnelHistory'
+import { useTranslations } from 'next-intl'
+
+const ReloadIcon = () => (
+  <svg fill="none" height={14} width={14} xmlns="http://www.w3.org/2000/svg">
+    <path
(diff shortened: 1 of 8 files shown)


ArturDolzan (approved):
LGTM

PR #504, opened 2024-09-06
Fix Drawer not closing when clicking outside

gndelia:
While developing the new design, I just happened to notice that the drawer doesn't close when clicking outside. This is due to a bad check when trying to run the `onClose` function because I check if any of the Rainbow Kit modals are opened (to prevent double-closing). However, for that check, I was checking the variable `openConnectModal`, which is a function... so it is always true 🤦🏽  The proper variable to check was `connectModalOpen`

Now the drawer will be closed when clicking outside, or when pressing `ESC` key

https://github.com/user-attachments/assets/b7563564-fab4-4aa0-9d23-c3b6294bdb0b


gabmontes (approved):
uACK

ArturDolzan (approved):

PR #508, opened 2024-09-10
Deploy redesign on test url

gndelia:
With this temporal github action, we should be able to deploy the redesign page as we merge it, so we can test it.


ArturDolzan (approved):
utACK

gabmontes (approved):

mbonfoster (approved):
uACK

PR #512, opened 2024-09-11
Reorder imports for customizing chain with rpc

### LINKED ISSUE #510: Allow customizing RPC urls per env variables
The sepolia RPC is stuck and users can't see their balances updated. The RPC Url comes from the viem's chain definition. We need to provide a mechanism to override RPC URLs through env variables

gndelia:
Closes #510

The fix for customizing a RPC URL for Sepolia applied in #511  broke in web workers - apparently due to some circular reference issues. This PR fixes it.


jcvernaleo (approved):
OK

ArturDolzan (approved):

mbonfoster (approved):

Issue #518, opened 2024-09-16
Use "infinite scrolling" in the Transaction History table


gndelia:

gndelia:
Closed by #522

Issue #516, opened 2024-09-14
Update transaction-history UI


gndelia:
See design on [figma](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=4204-9233&node-type=frame&m=dev)

gndelia:
Closed by #522

Issue #521, opened 2024-09-18
Apply custom scrollbar design to the transaction history table


gndelia:
The new design breaks the rounded borders (See bottom of the image)

<img width="143" alt="image" src="https://github.com/user-attachments/assets/6b6d332f-c41c-4476-ac84-e46882668c1c">

dvnahuel:
Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=10646-15687&t=2UNyD1ntJUKESVxK-1)

Issue #515, opened 2024-09-14
Update "No transactions" and "Not connected" states in transaction history


gndelia:
See designs on figma for ["no transactions"](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=5077-38403&node-type=frame&m=dev) and for ["not connected"](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=5267-5787&node-type=frame&m=dev)

gndelia:
Closed by #524

PR #522, opened 2024-09-18
Transaction history redesign

gndelia:
Closes  https://github.com/hemilabs/ui-monorepo/issues/516 (Transaction History page redesign)
Closes https://github.com/hemilabs/ui-monorepo/issues/518 (Apply infinite scrolling in the TX history table)

Pending to implement:
- https://github.com/hemilabs/ui-monorepo/issues/515 (not connected / no transactions available states)
- https://github.com/hemilabs/ui-monorepo/issues/521 (Apply custom design on scrollbars)

--------------------

This PR applies the redesign to the transaction history page. Some notes about the changes made:

- Getting the proper height of the table container was hard. This was because of the way the library used to implement infinite scrolling works. The library dynamically renders all rows, one below the other using `position: absolute` + `translate-y`. This requires knowing the height of each row (which was fixed, so this was ok), and a fixed height on the container of the table. We couldn't just do `height: 100%` because it would otherwise, render all the rows! Once the fixed height is set, it uses the `IntersectionObserver` API to know which rows are truly visible in the container and only shows those.
However, per the designs, we needed a fixed space around the page (some padding-top/bottom), and made the table to fill the rest of the space (so it wasn't fixed, but depending on the screen size). To solve this, I used `calc` to calculate "the remaining space" and set it as a fixed value. This was done assuming the rest of the items had a fixed size (Which was the case, but they were set among different layers of files). It's not the best CSS I've ever written, but it does work and I hope the comments are clear enough for future changes if needed.
As complex as this code was, I resorted to writing it down in a `css` file so the component is cleaner to read. In addition to that, as every column is now also "dynamic" (And they are re-ordered depending on the viewport), I had to use `flex-basis` + `flex-grow:1` and `flex-shrink: 0` to set some minimum width on every cell... This is also set on the very same `.css` file
 
- From the layout defined in previous PRs (For the `/tunnel` page), I had to make minor tweaks to the layout.

- Note also that the entire row is clickable. It currently takes you to the Review modal, which will be replaced with Drawers in the new design (not implemented yet)

- Libraries used for virtualization and table were bumped to their latest version

Some screenshots:

image

image

image

image

And a nice video

https://github.com/user-attachments/assets/900ba531-5ece-4f22-bd90-d3072d17bbc6


gabmontes (approved):
uACK

webapp/app/[locale]/tunnel/transaction-history/_components/topBar.tsx
```diff
@@ -0,0 +1,57 @@
...
+type Props = {
+  filterOption: FilterOptions
+  onFilterOptionChange: (filter: FilterOptions) => void
+}
+
+export const TopBar = function ({ filterOption, onFilterOptionChange }: Props) {
+  const t = useTranslations('tunnel-page.transaction-history.top-bar')
+  return (
+    <div className="flex w-full flex-wrap items-center gap-x-2 gap-y-2 px-3.5 py-2 md:flex-nowrap md:px-3">
+      <span className="order-1 flex-shrink-0 flex-grow basis-2/5 md:flex-grow-0 md:basis-auto">
```
Comment (ArturDolzan) [inline webapp/app/[locale]/tunnel/transaction-history/_components/topBar.tsx:19]: Perhaps it's better to use h1 or h2 here, thinking about web semantics. What are your thoughts on this?

webapp/components/button.tsx
```diff
@@ -1,4 +1,9 @@
+import Link from 'next-intl/link'
 import { ComponentProps } from 'react'
+import { isRelativeUrl } from 'utils/url'
+
+const commonCss = `text-ms box-content flex items-center justify-center
```
ArturDolzan, line 5: Do we need backticks here? There is no string interpolation.

webapp/components/button.tsx
```diff
@@ -1,4 +1,9 @@
...
+  rounded-lg border border-solid px-3 py-1.5 font-medium leading-5 disabled:opacity-40`
 
 const variants = {
   primary: `border-orange-700/55 from-orange-500 to-orange-600 text-white hover:border-orange-700/70
```
ArturDolzan, line 9: Same here

ArturDolzan (commented):
A few minor things otherwise LGTM

webapp/components/button.tsx
gndelia: it is needed to use multiline strings!

image


as it is a long string, I felt it read better in two lines. (It's a 37 length string(

webapp/components/button.tsx
gndelia: Same as [here](https://github.com/hemilabs/ui-monorepo/pull/522#discussion_r1765570797) - even more here, as it is longer 😅

ArturDolzan: ahh my mistake was just a quick look and I didn't notice

webapp/app/[locale]/tunnel/transaction-history/_components/topBar.tsx
Comment (gndelia) [reply webapp/app/[locale]/tunnel/transaction-history/_components/topBar.tsx]: I updated it to use a `h5` [here]([27ca8ba](https://github.com/hemilabs/ui-monorepo/pull/522/commits/27ca8bae49cd783182d8c8cea905080c732d4000)). As this is a filter bar, this is in pair with the lowest headings we've used so far across the app (We haven't used `h6`)

ArturDolzan (approved):

Issue #539, opened 2024-09-23
Set a fixed gasLimit for tunnel operations in Sepolia


gndelia:
As there are a high number of fees in Sepolia, it seems the OP SDK over-estimates the gas limit units needed to Deposit, Prove, and Claim, causing several operations to fail. After @max-sanchez analyzed several transactions in Sepolia, we're setting gas limits to `2_000_000` units.

Leaving this issue to see if we later need to revert this change

gndelia:
@max-sanchez  agreed we can remove this. I'll do so as part of #37

PR #540, opened 2024-09-23
Hardcode gasLimit for deposit/prove/claim

gndelia:
Related to #539

Currently, there are high prices in Sepolia. In addition to that, it seems there's an over-estimation of the gas units needed to deposit/claim/prove, which causes crazy amounts of fees that are later not paid at all. However, this prevents some users from submitting transactions.

@max-sanchez  analyzed several transactions, and he concluded that setting a fixed gas limit of `2.000.000` should work for all the operations in Sepolia.

For doing so, all the OP-SDK methods offer a parameter named `overrides` where consumers of the SDK can override the `gasLimit` value. This worked well for `proveMessage` (prove step) and `finalizeMessage` (claim step).

However, this caused an error message `Execution Reverted` when trying to deposit. For that, I had to patch the `cross-chain-messenger` in the OP SDK.

The change is simple: you can see [here](https://github.com/ethereum-optimism/ecosystem/blob/2ab98949a55a2cb6fd50502eb71f9888071b9a50/packages/sdk/src/cross-chain-messenger.ts#L2380) that the SDK overrides the `gasLimit` provided by the consumer (it overrides my overrides 😄 ), and there's no way to avoid that because to skip the `gasEstimation` that would eventually override my `gasLimit` setting, it checks the parameter [isEstimatingGas](https://github.com/ethereum-optimism/ecosystem/blob/2ab98949a55a2cb6fd50502eb71f9888071b9a50/packages/sdk/src/cross-chain-messenger.ts#L2369), which is defaulted to `false`... and it is not provided when depositing (see [here](https://github.com/ethereum-optimism/ecosystem/blob/2ab98949a55a2cb6fd50502eb71f9888071b9a50/packages/sdk/src/cross-chain-messenger.ts#L1962)).  
The same principle applies to `depositERC20`.

The patch, then, changes that default. value so it allows me to override the value

After merging this PR, there are no changes visible to the user, except that now fees are calculated a bit differently (and the gas limit of `2.000.000` is visible in MM)

https://github.com/user-attachments/assets/5014fce0-8750-43bd-a856-e9353b34977d

### DIFF
--- a/patches/@eth-optimism+sdk+3.2.1.patch
+++ b/patches/@eth-optimism+sdk+3.2.1.patch
@@ -14,9 +14,27 @@ index 743d4b4..6eafe8d 100644
      waitForMessageReceipt(message: MessageLike, opts?: {
          fromBlockOrBlockHash?: BlockTag;
 diff --git a/node_modules/@eth-optimism/sdk/dist/cross-chain-messenger.js b/node_modules/@eth-optimism/sdk/dist/cross-chain-messenger.js
-index 01ac898..bc760d9 100644
+index 01ac898..fcaeb3c 100644
 --- a/node_modules/@eth-optimism/sdk/dist/cross-chain-messenger.js
 +++ b/node_modules/@eth-optimism/sdk/dist/cross-chain-messenger.js
+@@ -113,7 +113,7 @@ class CrossChainMessenger {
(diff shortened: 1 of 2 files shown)


max-sanchez (approved):
LGTM

gabmontes (approved):
LGTM

PR #542, opened 2024-09-24
Update modal UI

### LINKED ISSUE #535: Update "Token Selector" modal
See designs on [figma](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=4662-4828&node-type=frame&m=dev)

gndelia:
Closes #535 

This PR updates the token selector modal.

As part of these changes, I updated how the Overlay is calculated, so it's injected into the proper container instead of the whole body, making it easier to "expand" to the entire container.

I also moved some components outside of `ui-common` (as I've been doing) as that package made sense when there were multiple sites in this repo, but now there's only one. So it just adds extra noise.

https://github.com/user-attachments/assets/f3ca1650-17ec-42b2-9f07-a85f41646071

Uploading Screen Recording 2024-09-24 at 12.11.54 PM.mov…


gabmontes (approved):
uACK

ArturDolzan (approved):

Issue #549, opened 2024-09-30
Deposit time may be missing if the user leaves the tunnel before completing


gndelia:
## Steps to reproduce

1. Initiate a deposit of Sepolia ETH
2. After confirming in Metamask, close the tunnel app before the transaction gets confirmed
3. Enter back in the app and go to the transaction history. You will see the "Time" field empty

#### Workaround

If the user re-syncs the transaction history, the timestamp loads properly.

There is a [similar process](https://github.com/hemilabs/ui-monorepo/blob/fcdb38af6783e8b910f944eb81b70a85e25d437c/webapp/context/tunnelHistoryContext/withdrawalsStateUpdater.tsx#L101) for withdrawals, and for [bitcoin deposits](https://github.com/hemilabs/ui-monorepo/blob/fcdb38af6783e8b910f944eb81b70a85e25d437c/webapp/context/tunnelHistoryContext/bitcoinDepositsStatusUpdater.tsx#L27). We could either generalize it or extend the EVM one. Related to #486 and #487

Issue #529, opened 2024-09-19
Replace Deposit / Withdraw modals with Drawers


gndelia:
This is implementing the redesign version of the portal UI for bitcoin withdrawals.

PR #552, opened 2024-10-01
View past deposits drawer

gndelia:
Related to #529

This PR partially replaces the Modal for EVM Deposits (ETH native and ERC20 tokens). After this PR, when users click a deposit in the transaction history, the drawer appears there on that page, letting you see the operation. I cannot complete my testing for replacing the "deposit in progress" due to the high fees... I have some pending changes that I will either push to this PR or send a new one to replace the Dialog when depositing fully.

As a side note, deposits are now added to the history as soon as the user confirms in the MM wallet the transaction, instead of waiting for confirmation as it used to do before. This may cause the confirmation to be missed if the user leaves the page or refresh. (See #549 for that)

As of now, the "approval" transaction hash will be saved in memory, if an approval was done, so it can be shown if available when viewing past operations. Note that it can't be recovered if the user re-syncs their history because there's no easy way to link an approval to a deposit.

The next step is updating the withdrawals, whereas all UI components are built as part of this PR, it should be a significantly smaller PR. Note that once all dialogs are replaced, many components will be removed. I'm keeping them so I can merge these features without breaking the app.

https://github.com/user-attachments/assets/2599feeb-751a-4a9a-98fc-558da4381c41

https://github.com/user-attachments/assets/9f08f2d1-7257-47f4-8aa4-0d397c4c0f6a

An example of a transaction when the approval Tx hash was saved

image

image

Examples of transactions in progress (Mocked, as I can't fully test deposits for the time being)

<details>
  <summary>Desktop</summary>

#### Approval in progress

image

#### Approval confirmed, user is prompted in MM to confirm the deposit

image

#### Deposit in Progress

image

#### Deposit confirmed

image

</details>

<details>
  <summary>Mobile</summary>

#### Approval in progress

image

#### Approval confirmed, user is prompted in MM to confirm the deposit

image

#### Deposit in Progress

image

#### Deposit confirmed

image

</details>


ArturDolzan (approved):

gabmontes (approved):
Re the approvals, there is no need to store them anywhere else than in memory. Ok with that.

PR #555, opened 2024-10-01
Use esplora-client

gabmontes:
Update the bitcoin API logic to use [`esplora-client`](https://github.com/hemilabs/esplora-client) instead of custom logic to query mempool.space or blockstream.info.

### DIFF
--- a/webapp/.env
+++ b/webapp/.env
@@ -1,6 +1,7 @@
 NEXT_PUBLIC_BTC_INPUTS_SIZE=105
 NEXT_PUBLIC_BTC_OUTPUTS_SIZE=25
 # Once mainnet is published, these values below may go into .env.development as they belong to hemi's testnet
+NEXT_PUBLIC_BITCOIN_NETWORK=testnet
 NEXT_PUBLIC_MEMPOOL_API_URL="https://mempool.space/testnet/api"
 NEXT_PUBLIC_RAINBOW_APP_NAME=Hemi
 NEXT_PUBLIC_RAINBOW_PROJECT_ID=THE_PROJECT_ID
--- a/package-lock.json
+++ b/package-lock.json
@@ -13693,6 +13693,14 @@
         "node": ">=0.10"
       }
     },
+    "node_modules/esplora-client": {
+      "version": "1.1.0",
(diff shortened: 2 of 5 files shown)


ArturDolzan (approved):

webapp/.env
```diff
@@ -1,7 +1,8 @@
 NEXT_PUBLIC_BTC_INPUTS_SIZE=105
 NEXT_PUBLIC_BTC_OUTPUTS_SIZE=25
 # Once mainnet is published, these values below may go into .env.development as they belong to hemi's testnet
+NEXT_PUBLIC_BITCOIN_NETWORK=testnet
```
gndelia, line 4: While this works for now, in the redesigned version this will be decided  from a navbar

image

which after choosing either `testnet` or `mainnet`, updates a query string and shows the appropriate chains from the environment.

This dropdown is not available now in the UI, but the support for multichain already exists in `main`. So you could implement this not to use the `.env` and directly instantiate the appropriate instance depending on this flag.

You can see how to get the appropriate Bitcoin instance in [here](https://github.com/hemilabs/ui-monorepo/blob/main/webapp/hooks/useBitcoin.ts).

The bitcoin chains definitions are [here](https://github.com/hemilabs/ui-monorepo/blob/main/btc-wallet/chains/index.ts) - there's a `testnet` flag if you need to `if/else` somewhere given a bitcoin chain

Considering this wasn't already supported (the btc API was hardcoded to testnet), _this is ok and can be merged_, but you want to create an issue and link it in the comments or something.

gabmontes: @gndelia I think we should add that support in a different PR as `utils/btcApi.ts` never had a way to switch networks nor this PR introduces that.

To allow switching networks, `utils/btcApi.ts` may need to export a function that receives the current network name and returns an object with the utility functions. Other option is to add a `network` parameter to all calls. Either way I guess it is a big change worth doing in a different PR.

gndelia: sounds ok for me, can you just create an issue so we don't forget about that?

gabmontes: https://github.com/hemilabs/ui-monorepo/issues/557

gndelia (approved):

PR #556, opened 2024-10-01
Enable EVM deposits with drawer

gndelia:
Related to #529
Follow-up of #552 

This PR adds the missing pieces to enable the drawer when depositing. Here there was no dialog to replace 😄  This PR saves in memory the approval tx hash and switches between steps. 

Note: It's pending to prevent users from closing the drawer while the operation is in progress. 

https://github.com/user-attachments/assets/3142c6c9-7a8f-4f17-876f-062838a57408

https://github.com/user-attachments/assets/60c0c330-fb38-4066-abc5-2e7e6e2d848d


gabmontes (approved):
uACK

ArturDolzan (approved):

Issue #558, opened 2024-10-02
Find a reliable way to retrieve Prove TX Hash and Claim TX hash given a withdrawal to EVM L1 when resyncing history


gndelia:
Given a withdrawal to an L1 EVM chain, we need two transactions to show in the Drawer related to the withdrawal:

- The prove transaction hash
- The claim transaction hash

I am not sure how to get the Prove transaction hash - Discussion here https://github.com/ethereum-optimism/developers/discussions/609 in OP repo. 
To get the Claim transaction hash, the `crossChainMessenger` offers a method named [getMessageReceipt](https://sdk.optimism.io/classes/crosschainmessenger#getMessageReceipt) which accepts a withdrawal transaction hash. However, unless provided, this will scan the logs of the entire L1 chain (from 0 to latest). This ends up in timeout almost every time. We could narrow the search with at least the _From_ block to be posterior to the withdrawal; however, the withdrawal transaction hash is on the L2, so the block number belongs to that chain. If we could get at least a starting block to which we could search from, we could use that to walk the chain in chunks to get the missing claim hash

gabmontes:
@gndelia did we solve this already?

gndelia:
I asked on the Optimism repo, and I got this answer https://github.com/ethereum-optimism/developers/discussions/609#discussioncomment-10848903 which seems to be helpful, but links to a diagram which needs further analysis to figure out

gndelia:
I did a bit of extra research, and it seems the only option to get this is by reading Ethereum logs. It's up to us if we want to implement this or not. Thoughts, @gabmontes  ?

gndelia:
After discussing it with Gabriel, the best option here seems to update the withdrawals subgraph to also process the Prove and Finalize events. This way, we could add, if found, the hashes for both into the withdrawal

Then, the subgraph-API needs to be updated to return these new hashes in the checksum format.

Lastly, the FE should check these and add them if found

PR #559, opened 2024-10-02
Save prove and claim tx hashes

gndelia:
Adding some changes to `main` as they are not strictly related to the UI, but are required for the redesign branch (and it is easier to review, in smaller PRs 😄 ).

This PR:

- Saves the prove and claim transaction hash, so they can be shown when viewing the whole Withdrawal flow (in the redesign)
- For that, I now need to split the `EvmWithdrawOperation` into two types: `ToEvmWithdrawOperation` (when L1 is EVM compatible) and `ToBtcWithdrawOperation`(when L1 is bitcoin)
- For that, the types are updated in a few places, and the `withdrawals` are read, in most places where the L1 is an EVM-compatible with `useToEvmWithdrawals`  instead of directly reading them from the Tunnel History, as doing so already types them in the proper type
- I took advantage of this change and now Bitcoin withdrawals use `BtcWithdrawStatus`, an enum I've created instead of using the `MessageStatus` from OP which has extra statuses that are not valid for Bitcoin

The issue https://github.com/hemilabs/ui-monorepo/issues/558  was created as well. 

No visible changes for the UI of the user

### DIFF
--- a/webapp/app/[locale]/tunnel/_components/actionableOperations.tsx
+++ b/webapp/app/[locale]/tunnel/_components/actionableOperations.tsx
@@ -1,12 +1,12 @@
 import { MessageStatus } from '@eth-optimism/sdk'
 import { featureFlags } from 'app/featureFlags'
 import { useBtcDeposits } from 'hooks/useBtcDeposits'
-import { useTunnelHistory } from 'hooks/useTunnelHistory'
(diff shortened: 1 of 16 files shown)


ArturDolzan (approved):

gabmontes (approved):
uACK

PR #560, opened 2024-10-03
View evm withdrawals drawer (Part 1)

gndelia:
Related to #529

This is Part 1 to replace the Dialogs with Drawers for withdrawals. After this PR, users will be able to see the withdrawal state in the drawer. Note that there is no call to action: For that, I need to move some code of the Prove/Claim components.

Note that while the file `webapp/app/[locale]/tunnel/_components/reviewOperation/reviewEvmWithdrawal.tsx` shows as a very large diff, actually it is a brand new file, which _happens_ to share the file name with the previous version. However, removing the original file would make it complicated to review this PR (or merge it partially) because it contains the logic of the previous modal, which also had code for proving/claiming.

To make this easier for you folks to review, I'm then renaming the original `reviewEvmWithdrawal.tsx` to `reviewEvmWithdrawalOld.tsx` in the same folder, updating the references, and then it will be removed in the next PR. Not ideal, but it will make it easier for all of you to review. This way, I can send a 2nd part with the Call to Action added to the drawer.

In summary, this PR is a set of changes + `webapp/app/[locale]/tunnel/_components/reviewOperation/reviewEvmWithdrawal.tsx` which should be reviewed as a full new file.

Incoming PR with Part 2, where the drawer has the Call to Action!

https://github.com/user-attachments/assets/af68776e-b427-47b5-a334-ade851bbaf5f

Some screenshots for mobile

image

image

### DIFF
--- a/webapp/app/[locale]/tunnel/_components/claim.tsx
+++ b/webapp/app/[locale]/tunnel/_components/claim.tsx
@@ -30,7 +30,7 @@ import {
 } from '../_hooks/useTunnelState'
 
 import { ReviewBtcDeposit } from './reviewOperation/reviewBtcDeposit'
-import { ReviewEvmWithdrawal } from './reviewOperation/reviewEvmWithdrawal'
+import { ReviewEvmWithdrawal } from './reviewOperation/reviewEvmWithdrawalOld'
 
 const EvmSubmitButton = function ({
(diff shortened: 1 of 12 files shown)


ArturDolzan (approved):

gabmontes (approved):
uACK

PR #562, opened 2024-10-03
View evm withdrawals drawer pt2

gndelia:
Related to #529 

This PR adds the redesigned view for withdrawals. A new feature of this view is that now the prove transaction hash and the claim transaction hash are saved  (doesn't work for previous operations, and for resync see https://github.com/hemilabs/ui-monorepo/issues/558)

This PR adds the success path, in the next PR I will implement the errors (Both for withdrawals and deposits in EVM)

https://github.com/user-attachments/assets/0418220c-ead3-453b-86c8-5f936ad46234

Some screenshots on mobile

image

image

image

image


gabmontes (approved):
uACK

ArturDolzan (approved):

PR #566, opened 2024-10-08
Handle errors deposit withdrawal

gndelia:
Related to #529 

This PR adds error handling and re-tries for Ethereum Deposits and Withdrawals. After this PR, the Sepolia flow should be completed. It is pending to implement it for bitcoin

### Some screenshots

State updated in the transaction history

image

Failed operations are listed if users want to retry them, at any step

image

image

image

### DIFF
--- a/webapp/app/[locale]/tunnel/_components/reviewOperation/step.tsx
+++ b/webapp/app/[locale]/tunnel/_components/reviewOperation/step.tsx
@@ -67,9 +67,7 @@ const Completed = function ({
         top={{
           bgColor: 'bg-neutral-50',
           children: (
-            <>
-              <span className="mr-auto text-neutral-600">{description}</span>
-            </>
+            <span className="mr-auto text-neutral-600">{description}</span>
           ),
         }}
       />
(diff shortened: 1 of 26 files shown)


gabmontes (approved):
uACK

webapp/app/[locale]/tunnel/_components/reviewOperation/step.tsx
```diff
@@ -60,16 +56,23 @@ const Completed = function ({
...
+            <SeeOnExplorer chainId={explorerChainId} txHash={txHash} />
+          </>
+        }
+        top={{
+          bgColor: 'bg-neutral-50',
+          children: (
+            <>
+              <span className="mr-auto text-neutral-600">{description}</span>
+            </>
+          ),
```
Comment (ArturDolzan) [inline webapp/app/[locale]/tunnel/_components/reviewOperation/step.tsx:73]: I think we don't need this fragment

```suggestion
          children: (
             <span className="mr-auto text-neutral-600">{description}</span>
          ),
```

ArturDolzan (commented):
A minor comment otherwise LGTM.

webapp/app/[locale]/tunnel/_components/reviewOperation/step.tsx
Comment (gndelia) [reply webapp/app/[locale]/tunnel/_components/reviewOperation/step.tsx:73]: You're correct, it is not needed indeed!

Comment (gndelia) [reply webapp/app/[locale]/tunnel/_components/reviewOperation/step.tsx]: Removed in 35cd5f2

PR #584, opened 2024-10-11
Improve repository setup

gabmontes:
This PR adds templates and improves the `package.json`, actions and workflows.


.github/ISSUE_TEMPLATE/bug-report-template.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Steps to reproduce
+
+<!-- List the steps needed to reproduce the bug. Include any other relevant
+information if available. -->
+
+### Environment
+
+- **OS**: _e.g., Windows, macOS, Linux_
+- **Runtime/Browser/Version**: _e.g., Node 22, Chrome 120_
```
gndelia, line 22: I would add wallet, maybe?

.github/actions/deploy-portal/action.yml
```diff
@@ -1,75 +1,74 @@
...
+    description: Claim tokens microservice URL
 
 runs:
   using: composite
   steps:
-    - name: Build site
-      run: npx lerna run deploy --scope=app
+    - run: npx lerna run deploy --scope=app
       shell: bash
+      name: Build site
```
gndelia, line 48: I think `name` should go before `shell`?

Both by natural order and our guideline order which recommends following [this order](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsname)

image

gndelia:
As the name of some workflows/checks changed, do we need to update anything on the repo settings (for status checks for example?) I can't check it as I don't have access to settings though.

.github/actions/deploy-portal/action.yml
gabmontes: No. The action metadata syntax docs have a different order than the workflow docs. 🤷🏻‍♂️

https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runs-for-composite-actions

I will update the docs.

gabmontes:
Yes, I need to change the settings.

.github/actions/deploy-portal/action.yml
gabmontes: Done!

https://github.com/bloqpriv/software-development-process/blob/master/guides/GitHub%20Actions.md#composite-action-guidelines

.github/actions/setup-env/action.yml
```diff
@@ -1,27 +0,0 @@
-name: Setup Environment
-description: Does minimal setup of environment for a job
-
-runs:
-  using: composite
-  steps:
-    - name: Add extra required branches
-      if: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.ref != 'main' }}
-      run: git fetch origin main:main ${{ github.event.pull_request.base.ref }}:${{ github.event.pull_request.base.ref }}
-      shell: bash
```
gndelia, line 10: I think this is still needed because otherwise, PRs targeting a branch different than `main` won't run

package.json
```diff
@@ -1,27 +1,17 @@
 {
   "name": "ui-monorepo",
   "version": "1.0.0",
-  "private": true,
   "scripts": {
     "format:check": "prettier --check .",
-    "lint": "eslint --cache .",
-    "postinstall": "patch-package"
+    "postinstall": "patch-package",
+    "lint": "eslint --cache ."
```
gndelia, line 7: Is this order correct?

image


it seems to me `lint` should go before `postinstall`

gndelia (approved):
I left a minor comment (which may be fixed in the `better-sort-package-json` package if it is confirmed as a bug), but otherwise LGTM

package.json
gabmontes: The logic used to sort aims to place all the pre- and post- scripts around the target script. Therefore, `postinstall` would go right after `install`, which goes before `lint`. If all the scripts were present, the order should be:

- preinstall
- install
- postinstall
- prelint
- lint
- postlint

Some are scripts are not defined in this case but the order reflects that logic.

gndelia: I know the logic of pre/post around their scripts, it seems to me that if the "main" script (`install` in this case) is missing, the pre/post should go as a regular script, sorted alphabetically. But I guess for that I should make my own package 😆

jcvernaleo (approved):
OK

.github/actions/setup-env/action.yml
gabmontes: We found that code is never triggered.

Issue #640, opened 2024-10-28
Prepare for mainnet


gndelia:
These changes need to be applied in order to prepare for mainnet to be enabled

#### Portal
 
- [X] https://github.com/hemilabs/hemi-viem/issues/23
- [X] - https://github.com/hemilabs/ui-monorepo/issues/639
- [X] - https://github.com/hemilabs/token-list/issues/7
- [X] - https://github.com/hemilabs/ui-monorepo/issues/647
- [x] - https://github.com/hemilabs/ui-monorepo/issues/557
- [x] - https://github.com/hemilabs/ui-monorepo/issues/586 
- [x] - https://github.com/hemilabs/ui-monorepo/issues/667
- [x] - https://github.com/hemilabs/ui-monorepo/issues/675 (To be confirmed if this is needed for mainnet or not)
- [x] - https://github.com/hemilabs/ui-monorepo/issues/709
- [x] - https://github.com/hemilabs/ui-monorepo/issues/719
- [x] - #658

#### Bitcoin 

- https://github.com/hemilabs/ui-monorepo/issues/738

Lastly, we should make the Network switch visible to users with this
- https://github.com/hemilabs/ui-monorepo/issues/479

gndelia:
Mainnet is live [after this release](https://github.com/hemilabs/ui-monorepo/releases/tag/20250305_1)

PR #653, opened 2024-11-01
Navigation keep testnet mainnet

gndelia:
### Description

After enabling `mainnet` through a feature flag, I notice a bug in which if the user navigates through relative links, the "network type" would be lost, defaulting to mainnet. This could cause a user to accidentally go from testnet to mainnet when going from (for example) `/tunnel` to `/demos`. 

As part of the changes, I updated the typescript `target` config value from `es5` to `es6`, needed for properly using `UrlSearchParams`. `es5` was a very lower version, so `es6` is still very safe. (Either way, this is just for type-checking; next uses their own target).

And as I also had to update `Tabs` component, I moved it out of `ui-common` into the `webapp` folder, as it is the only project that uses it 😅 

### Screenshots

Navigation now works as expected

https://github.com/user-attachments/assets/3cf3317b-9af3-4739-b88f-9bd0e18bc5c6

### Related issue(s)

Related to #640

### Checklist

- [x] Manual testing passed.


webapp/app/[locale]/_components/header.tsx
```diff
@@ -1,8 +1,8 @@
 import { CloseIcon } from 'components/icons/closeIcon'
 import { HamburgerIcon } from 'components/icons/hamburgerIcon'
+import { Link } from 'components/link'
```
Comment (leanazulyoro) [inline webapp/app/[locale]/_components/header.tsx:3]: but doesn't this cause other issues related to i18n? I mean, does navigation work fine while mantaining the selected language?

Comment (gndelia) [reply webapp/app/[locale]/_components/header.tsx:3]: it works because this wrapper imports the Link from `next-intl` and not the link from `navigation/next`

Comment (gndelia) [reply webapp/app/[locale]/_components/header.tsx:3]: so it also preserves the locale (the locale is part of the `pathname`, and not a query string)

leanazulyoro (approved):

gabmontes (approved):

PR #671, opened 2024-11-07
Improve estimations for mainnet

gndelia:
### Description

Estimations are a bit off in mainnet. This is due 2 reasons:

- There was an extra `overestimation` when calculating fees; however, this made no sense because the OP SDK already had the overestimation. As in Sepolia, fees are low, this was barely noticeable, but on `mainnet` it is very visible.
- Deposit of native tokens was throwing an error. This' been fixed before, so I applied the same fix.
We must also note that for ERC20 tokens, the estimation can't be retrieved without approving it first, because the simulation of the deposit reverts. This makes sense (as you can't simulate a deposit if the user doesn't have enough allowance), but in testnet it works. It may be that these tokens were simplified. We reviewed this with Max.

### Screenshots

Estimations on mainnet are now more accurate (similar to what MM asks)

Native token:

image

image

ERC20 (I had already approved the token at this point)

image

image

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/tunnel/_components/deposit.tsx
+++ b/webapp/app/[locale]/tunnel/_components/deposit.tsx
@@ -458,7 +458,7 @@ const EvmDeposit = function ({ state }: EvmDepositProps) {
   const gas = {
     amount: formatNumber(
       formatUnits(
-        depositGasFees + approvalTokenGasFees,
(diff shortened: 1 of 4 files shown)


gabmontes (approved):
uACK

ArturDolzan (approved):

PR #680, opened 2024-11-12
Fix spacing incorrect in var config in workflow

gndelia:
### Description

There's an extra space between `$` and `{{` in the workflow that deploys to staging


gabmontes (dismissed):

gabmontes (approved):

ArturDolzan (approved):
Ok

Issue #702, opened 2024-12-10
Improper input validations


jcvernaleo:
## Code Examples

These examples in `ui-monorepo` are mentioned specifically as places to double-check that necessary validation is occurring.

### Number Input

In the following example, a regular expression is used to verify if the input is a valid number. But the check is very basic and may not cover all cases, such as if the input is a very large number. It could be helpful to use a library like [joi](https://joi.dev/api/?v=17.13.3) to for easy-to-read schema validation.

File reference: `webapp/app/[locale]/tunnel/_hooks/useTunnelState.ts`
Line: 234
```
  const updateFromInput = useCallback(
    function (payload: UpdateFromInput['payload']) {
      // verify if input is a valid number
      const validationRegex = /^\d*\.?\d*$/
      if (!validationRegex.test(payload)) {
        return
      }
      dispatch({
        payload: payload.replace(/^0+/, ''),
        type: 'updateFromInput',
      })
```

In the following example, a zero is added to make the input a valid number if the input ends with a dot. While this approach may work, it may be helpful to do further checks and/or change the way input is gathered from the user.

File reference: `webapp/app/[locale]/tunnel/_hooks/useTunnelState.ts`
Line: 95
```
    case 'updateFromInput': {
      const { payload: value } = action
      // if input ends with a dot, add a zero so it is a valid number.
      const fromInput = value === '' ? '0' : value
      return {
        ...state,
        fromInput: fromInput.startsWith('.') ? `0${fromInput}` : fromInput,
      }
    }
```

### URLs

In the following example, it may be necessary to do more in-depth checks for the URL rather than just ensuring that the string starts with `https`. It is possible extra validation is done elsewhere, such as with a URL whitelist, but please confirm. It may also be helpful to developers to make it clearer within the code what other validation is done elsewhere.

File reference: `webapp/networks/utils.ts`
Line: 5
```
export const overrideRpcUrl = function (chain: Chain, rpcUrl?: string) {
  const isValidCustomSepoliaRpc = !!rpcUrl && rpcUrl.startsWith('https')
```

Similar to above, it may be useful to do more than just checking the URL starts with `https`.

File reference: `webapp/app/[locale]/get-started/_components/addChainManually.tsx`
Line: 28
```
}: {
  clickableLink?: boolean
  label: string
  value: string
}) => (
...
    {value.startsWith('https') ? (
      <div className="flex w-full items-center">
        <ConfigurationUrl clickableLink={clickableLink} href={value} />
[...]

PR #706, opened 2024-12-13
Improve user input validation

gabmontes:
Addresses the first comment in #702 by improving the validation of the number inputs.

Given that `joi` is a large and complex library, I just extracted the important bits out of it and created `not-joi`, keeping the API so we can swap it for the full library in the future if needed. And yes, doing so was not needed at all, I did that for fun.

### Related issue(s)

Related to #702

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/utils/notJoi.ts
+++ b/webapp/utils/notJoi.ts
@@ -1,0 +1,35 @@
+// Super lightweight partial re-implementation of joi.
+
+const notJoi = {
+  number() {
+    const positiveFixedPointNumberRx = /^(?:(?:\d+(?:\.\d*)?)|(?:\.\d+))$/
+    return {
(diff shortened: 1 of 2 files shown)


ArturDolzan (approved):

webapp/utils/notJoi.ts
```diff
@@ -0,0 +1,35 @@
...
+          return { error: 'Must be a finite number', value }
+        }
+        // Validated!
+        return { value: _value }
+      },
+    }
+  },
+}
+
+export default notJoi
```
gndelia, line 35: I would use a named export here, as all the codebases uses, to avoid naming the exported object differently in each context

gndelia (approved):

PR #707, opened 2024-12-13
Detect tunnel op depending on connected wallet

### LINKED ISSUE #704: Remember to keep the selected 'From network' item from the tunnel dropdown when moving through the pages.
As a user, I want the tunnel dropdown's 'From Network' selection to persist when navigating between pages, so I don't have to reselect it each time I return to or interact with the dropdown.

**Problem or use case addressed**
When the user selects the Hemi Sepolia network in the 'From Network' dropdown, navigates to the transaction history, and returns to the tunnel, the selection defaults to 'Sepolia' instead of retaining the previous selection.

gndelia:
### Description

This PR changes the relative links in the app that target the `/tunnel` page. Now, we will check if the wallet is connected to an L1 or L2 URL. If so, it will set the appropriate operation (deposit or withdrawal, depending on which one is connected).

For unsupported chains, it will resolve to `deposit.`

### Screenshots

https://github.com/user-attachments/assets/56b4e79d-51b0-4e0c-b96f-dce3719eb758

### Related issue(s)

Closes #704

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/_components/header.tsx
+++ b/webapp/app/[locale]/_components/header.tsx
@@ -2,6 +2,7 @@ import { CloseIcon } from 'components/icons/closeIcon'
 import { HamburgerIcon } from 'components/icons/hamburgerIcon'
 import { Link } from 'components/link'
 import { TunnelTabs } from 'components/tunnelTabs'
+import { useTunnelOperationByConnectedWallet } from 'hooks/useTunnelOperationByConnectedWallet'
 import dynamic from 'next/dynamic'
 import { HemiSymbol } from 'ui-common/components/hemiLogo'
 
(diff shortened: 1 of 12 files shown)


gabmontes (dismissed):

ArturDolzan (commented):
A single comment otherwise LGTM

webapp/app/[locale]/_components/header.tsx
```diff
@@ -19,27 +20,30 @@ type Props = {
...
+        <Link href={href}>
+          <HemiSymbol />
+        </Link>
+      </div>
+      <div className="hidden pl-6 md:block">
+        <TunnelTabs />
+      </div>
+      <WalletConnection />
+      <button
+        className="flex h-8 w-8 cursor-pointer items-center
```
Comment (ArturDolzan) [inline webapp/app/[locale]/_components/header.tsx:40]: It could be replaced as
```suggestion
        className="flex size-8 cursor-pointer items-center
```

Comment (gndelia) [reply webapp/app/[locale]/_components/header.tsx:40]: cool, I did not know about it. Thanks for the tip!

Comment (gndelia) [reply webapp/app/[locale]/_components/header.tsx]: Sugestion applied here [924d183](https://github.com/hemilabs/ui-monorepo/pull/707/commits/924d183ec85369d3da88209fc6cca7b9c9a75b35)

ArturDolzan (approved):

gabmontes (approved):

PR #724, opened 2024-12-28
fix: update "claimBtcDeposit" naming to "confirmBtcDeposit

### LINKED ISSUE #711: Update "claimBtcDeposit" naming to "confirmBtcDeposit"
### Description

The smart contract code doesn't use `claim` anywhere. We originally used this term when discussing the implementation, but the solidity code uses function related to `confirmDeposit`. We should use that wording in the code (Even though we may keep "Claim" in the UI's text) so it is more consistent with the solidity implementation. 

This is implemented in `webapp/hooks/useBtcTunnel.ts` and in `webapp/utils/hemi.ts` (And I think that's it)

DeVikingMark:
- Renamed function from claimBtcDeposit to confirmBtcDeposit in useBtcTunnel.ts
- Renamed function from claimBtcDeposit to confirmBtcDeposit in hemi.ts
- Updated all related imports and function calls
- Maintains consistency with smart contract implementation

Fixes #711

### DIFF
--- a/webapp/utils/hemi.ts
+++ b/webapp/utils/hemi.ts
@@ -125,43 +125,11 @@ export const initiateBtcDeposit = function ({
   )
 }
 
-export const claimBtcDeposit = ({
-  deposit,
-  from,
-  hemiClient,
-  hemiWalletClient,
-}: {
-  deposit: BtcDepositOperation
-  from: Address
-  hemiClient: HemiPublicClient
-  hemiWalletClient: HemiWalletClient
-}) =>
-  Promise.all([
-    getVaultIndexByBTCAddress(hemiClient, deposit),
-    getVaultAddressByDeposit(hemiClient, deposit).then(vaultAddress =>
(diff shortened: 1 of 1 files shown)


webapp/utils/hemi.ts
```diff
@@ -125,43 +125,11 @@ export const initiateBtcDeposit = function ({
...
-      from,
-      outputIndex: BigInt(outputIndex),
-      txId: deposit.transactionHash,
-      vaultIndex,
-    })
-  })
+export const confirmBtcDeposit = async (
+  params: ConfirmBtcDepositParams
+): Promise<TransactionResponse> => {
+  // существующий код...
```
gabmontes, line 131: I guess this is not the proposed change.

gabmontes (changes requested):

gndelia (changes requested):
@DeVikingMark #711 is just to rename a few functions and variables. No function should be removed

PR #727, opened 2024-12-30
Add support for mainnet in BTC Api

### LINKED ISSUE #557: Add multichain support to the bitcoin API helpers
The code in `utils/btcApi.ts` is hardcoded to point to the testnet. It should instead get the network from the user.

See: https://github.com/hemilabs/ui-monorepo/pull/555#discussion_r1783510066

gndelia:
### Description

This PR bumps `esplora-client` to the latest version (which uses typescript), and then updates `btcApi` to work with both `testnet` and `mainnet`. As `mainnet` is not enabled, there are no visible changes to the user, but now once the user switches from/to `mainnet`, it will hit the appropriate btc api.

### Screenshots

No visible changes to users

### Related issue(s)

Closes #557

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12372,9 +12372,10 @@
       }
     },
     "node_modules/esplora-client": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/esplora-client/-/esplora-client-1.1.0.tgz",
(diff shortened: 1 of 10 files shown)


rafappelt (approved):

gabmontes (approved):
uACK

PR #777, opened 2025-01-21
Fix getting the Status when a withdrawal TX is not completed 

### LINKED ISSUE #776: "Transaction not found" during withdraw in process
### Description

The code that watches for pending withdrawals runs is trying to get the timestamp of a withdrawal that hasn't been confirmed yet. This is because we're sending uncompleted withdrawals to this worker. As there's no way to distinguish a withdrawal that is pending from one confirmed after the user left the app, we need to load the TX's receipt before doing any other check.

### Steps to reproduce

1. Start a withdrawal
2. See errors in the console while unconfirmed.

Image

gndelia:
### Description

This PR fixes a bug in which the watcher workers would try to check the status of an EVM withdrawal (from Hemi to Sepolia, for example) when the TX wasn't confirmed yet. After that, I did a small refactor in the workers code

- ece997a89b58a6662948f908cac1b7d347fe00ca Applies the actual fix (getting the receipt and checking for its existence before querying the withdrawal status)
- 4be48d7172a77485a1059f7fea54dd85e45886b4 Splits the code in `webapp/workers/watchEvmWithdrawals.ts` in 2: This file will now only handle the workers communication, and the actual watching is now run in another file, `webapp/utils/watch/evmWithdrawals.ts`. This commit is mostly moving code around.
- As now the files are split, tests can be added in db4cd92

### Screenshots

https://github.com/user-attachments/assets/7e17dfbd-7c92-4a99-b5b0-1d54b92ffbf1

See - no errors in the console.

### Related issue(s)

Closes #776

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/test/utils/watch/evmWithdrawals.test.ts
+++ b/webapp/test/utils/watch/evmWithdrawals.test.ts
@@ -1,0 +1,105 @@
+import { MessageDirection, MessageStatus } from '@eth-optimism/sdk'
+import { hemiSepolia } from 'hemi-viem'
+import { ToEvmWithdrawOperation } from 'types/tunnel'
+import { createQueuedCrossChainMessenger } from 'utils/crossChainMessenger'
+import { getEvmBlock, getEvmTransactionReceipt } from 'utils/evmApi'
+import { createPublicProvider } from 'utils/providers'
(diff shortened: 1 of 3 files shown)


gabmontes (approved):
uACK

ArturDolzan (approved):

Issue #383, opened 2024-07-11
Multiple calls to create EVM caches


gabmontes:
When opening the tunnel app, the following line is logged multiple times to the browser's console:

<img width="513" alt="image" src="https://github.com/BVM-priv/ui-monorepo/assets/2621975/f67ff526-daa7-433e-b926-0a47ab5a133f">

The hook `useWeb3Provider` uses a `useMemo` to end call `createEthRpcCache` so this should not happen.

https://github.com/BVM-priv/ui-monorepo/blob/5721190528d389f01199881ad741b06ceebcb07a/webapp/hooks/useEthersSigner.ts#L52

The problem should be investigated. We may be leaking memory and/or losing cache advantages if multiple instances are being created.

gndelia:
After #37 is completed, `eth-rpc-cache` will be only used to create viem clients. I think we should review if a memorization per chainId is needed, but I don't think anything other than that is needed

PR #778, opened 2025-01-21
Improve rpc handling

### LINKED ISSUE #658: Add support for fallback when RPC fails

The current RPC URL we're using for Sepolia is failing. Perhaps, instead of using the override function we currently have, we could use [viem's support for fallbacks](https://viem.sh/docs/clients/transports/fallback.html) so multiple RPC URLs are configured and the library resolves which to call.

### LINKED ISSUE #773: Custom RPC should not land in wallet's settings
### Description

We've set a custom RPC URL per chain. The RPC is added to the chain definition.  
However, if users connect to this chain through their wallet, the custom RPC ends up leaking into the user wallet's settings. We should avoid that. The RPC URL should only be used to run read-only requests from the Portal.

gndelia:
### Description

This PR adds support for multiple RPC URLs by leveraging [viem's Fallback transport](https://viem.sh/docs/clients/transports/fallback.html) and [ether's fallback provider](https://docs.ethers.org/v6/api/providers/fallback-provider/).  
Remember that we use viem everywhere in the app, except when using the OP SDK, which requires a ethers provider (hence the double usage redundant until we get rid of the already deprecated OP SDK).

Now multiple URLs can be configured in each env variable, as long as they are split with `+`. For example:

```sh
NEXT_PUBLIC_CUSTOM_RPC_URL_SEPOLIA="https://sepolia.drpc.org+https://eth-sepolia-public.unifra.io"
```

The recommendation to this is to place the URL we want to be added to the wallet first. The rest can be used as fallbacks. Ethers in particular distribute the requests along all the URLs, so it's not a fallback mechanism.

I also simplified a bit the transformation from viem's client to ethers' providers, and the creating of public client as well.

### Screenshots

No changes are visible to the users

### Related issue(s)

Closes #658
Closes #773
Related to #383

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/networks/utils.ts
+++ b/webapp/networks/utils.ts
@@ -1,14 +1,18 @@
+import { isValidUrl } from 'utils/url'
 import { type Chain } from 'viem'
 import { defineChain } from 'viem/utils'
 
-export const overrideRpcUrl = function (chain: Chain, rpcUrl?: string) {
-  const isValidCustomSepoliaRpc = !!rpcUrl && rpcUrl.startsWith('https')
-  if (isValidCustomSepoliaRpc) {
+export const updateRpcUrls = function (chain: Chain, rpcUrlEnv?: string) {
+  if (typeof rpcUrlEnv !== 'string') {
+    return chain
+  }
+  const urls = rpcUrlEnv.split('+').filter(isValidUrl)
+  if (urls.length > 0) {
     return defineChain({
       ...chain,
       rpcUrls: {
         default: {
-          http: [rpcUrl],
+          http: urls,
         },
       },
     })
--- a/webapp/README.md
+++ b/webapp/README.md
@@ -8,10 +8,14 @@ The environment variables are defined in the `.env` file at the root of the proj
 The prefix `NEXT_PUBLIC_` is required for the variables to be available in the browser. A few variables can be set locally (in a `.env.local`), in addition to the ones already defined in the `.env`.
 
 ```sh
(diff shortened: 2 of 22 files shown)


webapp/networks/utils.ts
```diff
@@ -1,14 +1,18 @@
...
 import { type Chain } from 'viem'
 import { defineChain } from 'viem/utils'
 
-export const overrideRpcUrl = function (chain: Chain, rpcUrl?: string) {
-  const isValidCustomSepoliaRpc = !!rpcUrl && rpcUrl.startsWith('https')
-  if (isValidCustomSepoliaRpc) {
+export const updateRpcUrls = function (chain: Chain, rpcUrlEnv?: string) {
+  if (typeof rpcUrlEnv !== 'string') {
+    return chain
+  }
```
gabmontes, line 8: Why is this needed?

gndelia: in case a malformed configuraion. Then, it returns the chains as-is (without changes).

The `chain` is a read-only object (It can't be updated), so in order to update it we need to create a new one, and use that instead. By returning the OG without modifications, we can just import everywhere from this file (instead of the definition in `viem`) and it will work.

gabmontes (dismissed):

ArturDolzan (dismissed):
uACK

ArturDolzan (approved):

gabmontes (approved):

Issue #752, opened 2025-01-14
Add stake tokens


ArturDolzan:
### Feature or change description

Identify and define the list of assets to be used, such as mBTC, pumpBTC, bfBTC, etc., along with their icons. Determine whether these assets and their metadata will be stored in a shared library or a repository. This decision is still pending and needs further discussion.

### Tokens to add

- [x] enzoBTC
- [x] iBTC
- [x] M-BTC
- [x] pumpBTC
- [x] stBTC
- [x] uBTC
- [ ] Solv
- [x] Obelisk https://github.com/hemilabs/token-list/pull/32
- [x] HBTC https://github.com/hemilabs/ui-monorepo/pull/908
- [x] WBTC
- [x] tBTC v2
- [x] USDT
- [x] USDC
- [x] DAI
- [x] VUSD
- [ ] bfBTC (Bitfi)
- [ ] sBTC (Stakestone)
- [x] uniBTC (Bedrock) https://github.com/hemilabs/ui-monorepo/pull/1005
- [x] brBTC (Bedrock)
- [ ] LBTC
- [ ] solvBTC.BBN
- [ ] suBTC (Sumer) https://github.com/hemilabs/ui-monorepo/pull/1020
- [x] rsETH (Kelp) https://github.com/hemilabs/ui-monorepo/pull/960
- [x] WETH
- [x] MagpieBtc
- [x] Eigenpie ETH

HBTC is HemiBtc, which can be taken from the [tokenlist](https://github.com/hemilabs/token-list/blob/9dffe5ee61f19c65b2c58ab883278eb4609f1a55/src/hemi.tokenlist.json#L345)

gndelia:
This is a partial list of the tokens.

```json
[
  {
      "address": "0x9BFA177621119e64CecbEabE184ab9993E2ef727",
      "chainId": 43111,
      "decimals": 18,
      "logoURI": "TBD",
      "name": "Merlin BTC",
      "symbol": "M-BTC",
      "extensions": {
        "birthBlock": 849309
      }
    },
    {
      "address": "0x8154Aaf094c2f03Ad550B6890E1d4264B5DdaD9A",
      "chainId": 43111,
      "decimals": 18,
      "logoURI": "TBD",
      "name": "iBTC",
      "symbol": "iBTC",
      "extensions": {
        "birthBlock": 629450
      }
    },
    {
      "address": "0x6A9A65B84843F5fD4aC9a0471C4fc11AFfFBce4a",
[...]

Issue #35, opened 2024-01-31
Allow changing language


gndelia:
After #32 , we need to allow users changing languages.

qgatssdev:
@gndelia gm, what's the intention for this feature? What is the intended UX and where should the modal or dropdown be located?

gndelia:
@qgatssdev it hasn't yet been defined. If you want to work on the repo, I'd recommend taking a look at bugs or those "good first issue"

gndelia:
Closing in favor of https://github.com/hemilabs/ui-monorepo/issues/1203

PR #781, opened 2025-01-22
Add Staking Assets Table

### LINKED ISSUE #751: Staking Campaign - Implement Staking Assets Table

Create the Staking Assets table with the designed columns. Add a placeholder section below the table for when there are no stakes, displaying a message (“No stakes yet” / check for designs) and a + Stake button. Ensure proper styling and responsiveness of the table and placeholder elements.

ArturDolzan:
### Description

This PR implements the Staking Assets Table component.

### Screenshots

https://github.com/user-attachments/assets/b02581bb-adf0-4437-b043-d062a1a52b01

### Related issue(s)

Closes #751 

### Checklist

- [x] Manual testing passed.


webapp/app/[locale]/stake/_components/rewardTag.tsx
```diff
@@ -0,0 +1,77 @@
+import { DiamondPointsIcon } from '../../stake/_components/icons/diamondPoints'
+import { HemiPointsIcon } from '../../stake/_components/icons/hemiPoints'
+import { PointsIcon } from '../../stake/_components/icons/points'
+import { SolvXpPointsIcon } from '../../stake/_components/icons/solvXpPoints'
+
+type Props = {
+  label: string
+  value?: string | number
+  icon: React.ReactNode
+  backgroundColor?: string
```
Comment (gndelia) [inline webapp/app/[locale]/stake/_components/rewardTag.tsx:10]: Unless you expect to reuse this with many options, narrowing the type to the only available options works better (so you get proper intellisense when typing the value). See [example](https://github.com/hemilabs/ui-monorepo/blob/92ab2da2dc50dfe98b0e8b6630303d6dd6f29815/webapp/app/%5Blocale%5D/stake/dashboard/_components/stakePointsCards/index.tsx#L24)

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+import { WelcomeStake } from './welcomeStake'
+
+declare module '@tanstack/react-table' {
+  // This is needed to add the width property to the ColumnMeta
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  interface ColumnMeta<TData, TValue> {
+    width?: string
+  }
+}
```
Comment (gndelia) [inline webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx:28]: I think this should be done in [here](https://github.com/hemilabs/ui-monorepo/blob/main/webapp/global.d.ts) (it could also ve done in its own file, but make sure it gets included in the `tsconfig.json`

ArturDolzan:
It would be great to have your feedback as well @dvnahuel. If you want me to share the localhost environment, please let me know. Keep in mind there are some mobile things to fix, as I got this PR before your mobile designs, however, the basics were done.

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
```diff
@@ -0,0 +1,122 @@
+import { Card } from 'components/card'
+import { useTranslations } from 'next-intl'
+
+import { StakeAndEarnPointsGraphIcon } from '../../../_components/icons/stakeAndEarnPointsGraph'
+
+interface Props {
+  onClick: VoidFunction
```
Comment (gndelia) [inline webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:7]: I did not now about `VoidFunction` 😮  TIL!

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
```diff
@@ -0,0 +1,122 @@
...
+  rotation = 0,
+  size = 14,
+}: StarIconProps) => (
+  <svg
+    className={className}
+    fill={color}
+    height={size}
+    style={{
+      transform: `rotate(${rotation}deg)`,
+    }}
```
Comment (gndelia) [inline webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:29]:
```suggestion
    {...( rotation !== 0 ? { style: { transform: `rotate(${rotation}deg)` } } : undefined )}
```

I think this works better, so we don't add an unnecessary rotation styling when it isn't needed. You could also remove the default value `0` and check `rotation` against undefined.

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
```diff
@@ -0,0 +1,122 @@
...
+          size={12}
+        />
+      </div>
+    </button>
+  )
+}
+
+function GradientLabel() {
+  const t = useTranslations('stake-page.dashboard')
+  const labelParts = t('stake-earn-points-label').split('|')
```
Comment (gndelia) [inline webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:76]: See if [using t.rich](https://next-intl.dev/docs/usage/messages#rich-text) works for you what you're trying to achieve.

There are also examples of usage for this function in the repo

webapp/app/[locale]/stake/tokens/index.ts
```diff
@@ -0,0 +1,87 @@
+import { StakeToken } from 'types/stake'
+
+import enzoBtc from './images/enzoBtc.png'
+import iBtc from './images/iBtc.png'
+import mBtc from './images/mBtc.png'
+import pumpBtc from './images/pumpBtc.png'
+import stBtc from './images/stBtc.png'
+import uBtc from './images/uBtc.png'
+
+// Not sure if it's the best place to put this list of tokens
```
Comment (gndelia) [inline webapp/app/[locale]/stake/tokens/index.ts:10]: Not a blocker for this PR, but we discussed with @gabmontes  a way to move this into the token list. For the time being, use the list here as you did

dvnahuel:
Hi team! I’ve reviewed the issue and noticed some points for improvement:

1. **Empty State:**
- The "Stake and earn points" text should be justified.
- The "Stake CTA" should have no shadow. On hover, the opacity should increase, and the button should scale to 1.1x with a smooth transition animation.
- The Hemi background image is missing and should be added to provide more depth.
2. **Table:**
- Remove the container styles and shadow from the "three-dots" CTA.
- On hover, the three dots' color should change to `grayscale-950`.
3. **Responsive Design:**
- As Artur mentioned, we should do a walkthrough to ensure the responsive designs match the Figma layouts.

webapp/app/[locale]/stake/tokens/index.ts
Comment (gabmontes) [reply webapp/app/[locale]/stake/tokens/index.ts:10]: Right. Let's avoid having multiple token lists. All these should go to `hemilabs/token-list`. If not done in this PR, open an issue and do so immediately after.

Comment (gndelia) [reply webapp/app/[locale]/stake/tokens/index.ts:10]: The issue was already created https://github.com/hemilabs/ui-monorepo/issues/752

webapp/app/[locale]/stake/_components/rewardTag.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/rewardTag.tsx:10]: Sounds good. Done!

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx:28]: Done. thanks!

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:76]: Unfortunately, t.rich doesn't render very well in this case. That's why I went for a DIY solution.

webapp/app/[locale]/stake/tokens/index.ts
Comment (gndelia) [reply webapp/app/[locale]/stake/tokens/index.ts:10]: Here's the draft PR for the token list https://github.com/hemilabs/token-list/pull/31

Note that

- Some tokens are still TBD (listed in the issue) as they haven't been deployed to hemi
- A new hook may need to be created on top of [useTokenList](https://github.com/hemilabs/ui-monorepo/blob/main/webapp/hooks/useTokenList.ts), to partition the tokens from the whole token list into the tokens that are tunnelable and those that are from staking (which for now can be defined by having a `extensions.protocol` property)
- Token logos are in the repo, but protocol logos are not part of it, so there must be some kind of mapping in the portal from protocol name => logo (And even display name if needed)

webapp/types/stake.ts
```diff
@@ -0,0 +1,26 @@
...
+  address: string
+  chainId: number
+  decimals: number
+  logo: StaticImageData
+  name: string
+  symbol: string
+  extensions: {
+    birthBlock: number
+  }
+}
```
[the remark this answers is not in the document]
ArturDolzan: Done. Thanks!

webapp/app/[locale]/stake/tokens/index.ts
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/tokens/index.ts:10]: @gndelia would you like to put this PR to draft to wait the one you created (https://github.com/hemilabs/token-list/pull/31) to be merged first?

Comment (gndelia) [reply webapp/app/[locale]/stake/tokens/index.ts:10]: I'd move as forward as possible on this PR. IF it can be merged, but it is blocked by the other one, I'd merge it anyway. If it happens to get merged before this one is ready, then I'd update this PR with the new token list.

By the way, the token list PR depends on https://github.com/hemilabs/token-list/pull/30 which you can review 😄

Comment (ArturDolzan) [reply webapp/app/[locale]/stake/tokens/index.ts:10]: @gndelia I'm sorry, I saw that PR yesterday however I forgot to click on approve. Done!

webapp/app/[locale]/stake/_components/rewardTag.tsx
Comment (gndelia) [reply webapp/app/[locale]/stake/_components/rewardTag.tsx:9]: I still see `backgroundColor` as a `string` 🤔

Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/rewardTag.tsx]: Fixed.

webapp/app/[locale]/stake/_components/rewardTag.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const RewardTag = ({
+  label,
+  icon,
+  backgroundColor = 'bg-orange-500',
+  style,
+}: {
+  backgroundColor?: 'bg-orange-500' | 'bg-purple-900/70' | 'bg-neutral-950'
+} & Pick<Props, 'style' | 'label' | 'icon'>) => (
```
Comment (gndelia) [inline webapp/app/[locale]/stake/_components/rewardTag.tsx:19]: It seems to me here it got mixed up? There's only one component using this Props, perhaps you want to merge them?

```suggestion
type Props = {
  backgroundColor?: 'bg-orange-500' | 'bg-purple-900/70' | 'bg-neutral-950'
  label: string
  icon: React.ReactNode
  style?: React.CSSProperties
}

const RewardTag = ({
  label,
  icon,
  backgroundColor = 'bg-orange-500',
  style,
}:  Props) => (
```

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
```diff
@@ -0,0 +1,134 @@
...
+  rotation,
+  size = 14,
+}: StarIconProps) => (
+  <svg
+    className={className}
+    fill={color}
+    height={size}
+    {...(rotation !== 0
+      ? { style: { transform: `rotate(${rotation}deg)` } }
+      : {})}
```
Comment (gndelia) [inline webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:30]: careful here. If you want to check for `!== 0`, if `rotation` is optional, then it must be added a default value of `0`. Otherwise, if the consumer of this component does not pass any value, `rotation` will be equal to `undefined`, which is different than 0, and you will end up with

```tsx
style={{ transform: 'rotate(undefineddef)'}}
```


Either do `rotation !== undefined`, or add a default value of `0` to `rotation` (as it is optional)

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
```diff
@@ -0,0 +1,134 @@
...
+                className="absolute -top-[54px] left-[9px] -translate-x-1/2"
+                color="#FF7849"
+                size={22}
+              />
+            </span>
+          </>
+        ) : (
+          labelParts[1]
+        )}
+        {labelParts[2]}
```
Comment (gndelia) [inline webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:97]: For now, this is ok. But I think this should be solved on the SVG itself (meaning: given a locale, we load different svgs, each one in the appropriate language, with the text and the appropriate start placed).

Otherwise, what if the translation does not have an `i` character? Or, for example, if it was a Chinese alphabet. It would be a mess. Having different svgs, each one with the text and the starts located appropriately would solve this.

As we don't support multilanguage (though we add Spanish, there's no language selector yet - See https://github.com/hemilabs/ui-monorepo/issues/35 ), this is fine for now, but it may be to keep an eye on.

webapp/app/[locale]/stake/dashboard/page.tsx
```diff
@@ -12,11 +22,12 @@ import {
 
 const Page = function () {
   const t = useTranslations('stake-page')
+  const containerRef = useRef<HTMLDivElement>(null)
```
Comment (gndelia) [inline webapp/app/[locale]/stake/dashboard/page.tsx:25]: This ref is created in this file, and forwarded to `StakeAssetsTable` as a prop. It is not used elsewhere. Do we need it here for some reason? Can the ref creation be moved to the `StakeAssetsTable` component?

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:30]: I just set a default value (0) and changed the logic to:

`rotation !== 0 ? { transform: `rotate(${rotation}deg)` } : undefined`

IMO it's clearer now.

webapp/app/[locale]/stake/_components/rewardTag.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/rewardTag.tsx:19]: Yes this also works. I followed your earlier suggestion.

webapp/app/[locale]/stake/dashboard/page.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/dashboard/page.tsx:25]: Oh, good point. Moved to the inside component.

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:97]: @gndelia It will work if does not have the i character. There is a conditional. I know that is not the best solution but it's the faster I got for now. Let's iterate on this later for sure.

gndelia (dismissed):

webapp/types/stake.ts
```diff
@@ -0,0 +1,20 @@
...
+  | 'merlinChain'
+  | 'solv'
+  | 'stakeStone'
+  | 'bedRock'
+  | 'exSat'
+  | 'lorenzo'
+  | 'bitFi'
+  | 'pumpBtc'
+  | 'uniRouter'
+  | 'hemi'
```
gabmontes, line 15: Sort?

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx
Comment (gabmontes) [reply webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/welcomeStake.tsx:75]: @ArturDolzan the way the `labelParts` is parsed and used below is tricky. Please add a comment to the code for the function or the constant explaining how it is used and possibly a few examples for our future selves.

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+    columns,
+    data,
+    getCoreRowModel: getCoreRowModel(),
+    state: {
+      columnOrder:
+        // move "action" to the left in small devices
+        // and keep original order in larger devices
+        width < 1024
+          ? ['action'].concat(
+              columns.filter(c => !['action'].includes(c.id)).map(c => c.id),
```
Comment (gabmontes) [inline webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx:254]: Mapping to `c.id` and then applying the filter seems simpler.

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+        Array.from(Array(4).keys()).map(index => (
+          <tr className="flex items-center" key={index}>
+            {columns.map((c, i) => (
+              <Column key={c.id || i} style={{ width: c.meta?.width }}>
+                {/* @ts-expect-error it works */}
+                {c.cell()}
+              </Column>
+            ))}
+          </tr>
+        ))}
```
Comment (gabmontes) [inline webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx:104]: Is the idea of this block to add 4 empty rows?

Regarding the column attribute `key={c.id || i}`, when can be `c.id` not defined?

webapp/types/stake.ts
ArturDolzan: Done.

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx:254]: Good point. Done.

webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/dashboard/_components/stakeAssetsTable/index.tsx:104]: Yes, it is basically to render the skeleton and yeah, the c.id should not be undefined, so I made a change. Thanks.

gabmontes (approved):

PR #782, opened 2025-01-22
Add watcher for EVM deposits

gndelia:
### Description

This PR adds the last [watcher](https://cdn.marvel.com/content/1x/102_cba0100_comp_005_v901_f89ebdcc.jpeg) for operations so if any information is missing that is already part of the history, it will be updated.

This was implemented previously for withdrawals and bitcoin, but it was pending for EVM deposits.

If you want to review this commit by commit:

- 9704782c97d68dd8225ffa18a31724ab828aeee8 adds `typeWorker` as a small reusable utility for typing workers properly
- 8baee8b10c6ab8953f66cd19e0539a0e74719493 adds the watcher component, worker, and tests
- a2c5b0e5c2a97dd56b9cc6e410d3e4f3891af7eb Removes a commented line from a previous PR which should've been deleted

### Screenshots

Closing the drawer

https://github.com/user-attachments/assets/523cbb96-0168-4bcb-811f-f39ddce5ec7f

Closing the app

https://github.com/user-attachments/assets/9130dda1-4ae3-408e-8eb7-546e2ade3f8d

### Related issue(s)

Closes https://github.com/hemilabs/ui-monorepo/issues/549

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/context/tunnelHistoryContext/evmDepositsStatusUpdater.tsx
+++ b/webapp/context/tunnelHistoryContext/evmDepositsStatusUpdater.tsx
@@ -1,0 +1,117 @@
+import { WithWorker } from 'components/withWorker'
+import { useConnectedToUnsupportedEvmChain } from 'hooks/useConnectedToUnsupportedChain'
+import { useEvmDeposits } from 'hooks/useEvmDeposits'
+import { useTunnelHistory } from 'hooks/useTunnelHistory'
+import { useEffect } from 'react'
+import { EvmDepositOperation } from 'types/tunnel'
+import { isPendingOperation } from 'utils/tunnel'
+import { hasKeys } from 'utils/utilities'
+import { useAccount } from 'wagmi'
+import { type AppToWorker, getDepositKey } from 'workers/watchEvmDeposits'
+
+const WatchEvmDeposit = function ({
+  deposit,
+  worker,
+}: {
+  deposit: EvmDepositOperation
+  worker: AppToWorker
+}) {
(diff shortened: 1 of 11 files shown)


gabmontes (approved):
uACK

webapp/context/tunnelHistoryContext/evmDepositsStatusUpdater.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+      // refetch every 15 seconds
+      const interval = 15 * 1000
+
+      worker.postMessage({
+        deposit,
+        type: 'watch-evm-deposit',
+      })
+
+      const intervalId = setInterval(function () {
```
ArturDolzan, line 56: Have you considered using react-query here?

ArturDolzan (approved):
uACK

webapp/context/tunnelHistoryContext/evmDepositsStatusUpdater.tsx
gndelia: Yes! In fact, the original implementation (Before I migrated others into using these workers [for deposits it was never implemented lol]) originally used useQuery. The problem is that, in order to prevent 429 requests in the workers, the requests are enqueued, so the promises may take long until they are resolved. (Perhaps if the enqueuing is solved on provider level, this could be simplified)

However, I do agree that this parts needs to be improved. The communication between workers and React components is cumbersome to implement (mostly due to how react re-renders)

Issue #754, opened 2025-01-14
Staking Campaign - Implement Stake Drawer


ArturDolzan:
### Feature or change description

Develop the Stake Drawer that allows users to stake assets.

gndelia:
Closing. Completed in related PRs - only thing pending is https://github.com/hemilabs/ui-monorepo/issues/880

PR #789, opened 2025-01-27
Stake drawers pre-changes

gndelia:
### Description

As part of implementing the Stake drawer, I made some small refactors changes and moved files for better re-usage (as they will be re-used in the next PR). I'm splitting those changes into this PR so it is easier to review.

### Description commit by commit

- c87d1c0dbecd238409eda50996384cf3e7f94cf7 Adds a new component for the "gray section" in the drawers (See image in spoilers)

<details>
  <summary>Drawer's "gray" section</summary>
image
</details>

- 84d24ddb255b38e5e05ac39d18116bf3f6ca81b5 Moves the `SubmitWhenConnectedToChain` to the `components` folder 
- d691411715c681b03d80bfe9ae2c62c9c5585246 Removes an old-css style for Recatpcha, that was used when we had the Claim capsule process in the `/get-started` page. This is no longer used (Recaptcha is no longer part of our app), so this can be safely removed
- 35fe205ec52f2082c658b8a950a6f01b510dc0df Moves the `Drawer` component into its own folder, inside `components`. This is because I expect to split it into different files, so they can all go into this new folder
- fae5b251ba59ae045ba56630c1a8cd3104deb44b moves the `TokenInput` from the tunnel into the `components` folder. aebcc8dfe076c6432c7d691422c5e1a413febbb9 does the same for the `SetMaxBalance` file. These two will be used for the stake drawer, and the components are from the tunnel (See accordion image). I also did some renaming of the props, as they now belong to a more generic context, and are not that coupled into the tunnel form alone.   
The biggest change of this CSS class in comparison to the current implementation is that it defines the padding-x for all its children, except those with the class `skip-parent-padding-x`. This is because all drawer elements have padding... except for some particular components, such as the Token Preview (when adding a custom token), or the Gray Drawer section (See above). These components ignore the padding and extend to the whole drawer's width.  This also requires some components to add an extra `div` to absorb the padding, so it is not defined in the component definition (Like the `WarningBox` component)
The token list also required some custom styles, as it works as a drawer on mobile, but as a modal on larger screens

<details>
  <summary>Input section</summary>
image
</details>

- 06e5cdb09cc1f3509407a4fe0872fc5477076a84 and cfcd994efd1504c16f168c5ce7a7d7733a6d5e9ecreates `drawer-content` CSS class, unifying some of the styles from the different drawers across the app. Why not a component, you may ask? The thing is that in some cases I need a `div`, but in others, I need a `form`. So instead of using a `prop` stating which one I need, I think it has better readability if I directly compose the tag I need (`div` or `form`, depending on the case) with the CSS class. This also allows custom styles, as the `height` is not standard across all drawers (This is something we discussed with Nahuel, and we preferred to leave the height dynamic for some scenarios, and static for some others, in particular, in mobile).

### Screenshots

No visible changes for the users

### Related issue(s)

Related to #754 
Related to #754 

### Checklist

- [x] Manual testing passed.


webapp/components/customTokenDrawer/addPairToHemi.tsx
```diff
@@ -17,19 +17,21 @@ export const AddPairToHemi = function () {
...
-          <ExternalLink
-            className="text-orange-500 hover:text-orange-700"
-            href="https://github.com/hemilabs/token-list/issues/new"
-          >
-            {chunk}
-          </ExternalLink>
-        ),
-      })}
-    />
+    <div>
```
gabmontes, line 20: Why are these divs needed?

webapp/components/customTunnelsThroughPartner/index.tsx
```diff
@@ -47,14 +47,18 @@ export const CustomTunnelsThroughPartner = function ({
             {t(`${operation}.subheading`, { symbol: fromToken.symbol })}
           </DrawerParagraph>
         </div>
-        <Stargate fromToken={fromToken} toToken={toToken} />
+        <div>
```
gabmontes, line 50: And these...

webapp/components/customTokenDrawer/addPairToHemi.tsx
gndelia: The component itself doesn't have a padding. If I don't add the div, they will inherit the padding defined in `drawer-content`, and it won't look as expected. This div is placed so padding is added  - it works as a container of the component. I wouldn't like to add the padding to the component definition because this padding comes from the Drawer, and it is context dependant. The same applies to the other `div`

webapp/components/customTunnelsThroughPartner/index.tsx
gndelia: See https://github.com/hemilabs/ui-monorepo/pull/789#discussion_r1930939579

webapp/components/customTokenDrawer/addPairToHemi.tsx
gndelia: Note that now the `padding-x` is defined at drawer level for each children ([see here](https://github.com/hemilabs/ui-monorepo/pull/789/files#diff-80f90bcbe6a827d266de5580a576f0f796840f61160c2a547081c3818d32f4a9R1)) without needing for manually apply it for every child. The only 2 exceptions in all drawers, can also be excluded with `skip-parent-padding-x`).

ArturDolzan (approved):
uACK

gabmontes (approved):

PR #791, opened 2025-01-27
Read stake tokens from @hemilabs/token-list

gndelia:
### Description

This PR bumps `@hemilabs/token-list` and updates the existing stake code to reads from the new list. 

- 45b5d443cec511455a032fdb4a52676d839faf09 Bumps the library to the new version (and then a newer version came out 😆  which I've updated in eb74457da02ece29506bf4c84a76c4cdae37bbc2)
- 707a8ea4c8ce204d6c839c0e06a023c0247fcf46 updates the `useTokenList` to be named as `useTunnelTokens`. As of now, there won't be just "one" token list inside the app, but differents depending on the usage needed. This new hook is for tunneling
- 628af25b4b86492a1b2be435aead00aa6e911e2a updates `TokenInput` so instead of reading tokens from the component itself (using a hook), they are now received as a prop. But this is only if the token selector is not read-only (if users can't select a token, and they are stuck with the selected token, then there's no need to send the token list). I also updated the `isRunningOperation` prop to `disabled`.
- d988387e8fde31a46c7049a06099c0d8fd921086 adds `useStakeTokens`. I also made some minor improvements, such as using the `TokenLogo` component in the table or improving the typing of the protocols. As now token logos are read from `@hemilabs/token-list`,  I removed the logos from this repo.

### Screenshots

No visible changes for the users

### Related issue(s)

Closes #
Fixes #
Related to #

### Checklist

- [ ] Manual testing passed.


ArturDolzan (approved):

webapp/utils/token.ts
```diff
@@ -56,3 +56,6 @@ export const getErc20TokenBalance = ({
     args: [address],
     functionName: 'balanceOf',
   })
+
+export const isStakeToken = (token: Token) =>
+  token.extensions?.protocol !== undefined
```
gabmontes: Why thats is better than this?

```suggestion
export const isStakeToken = (token: Token) =>
  !!token.extensions?.protocol
```

gndelia: It's not, it's just different. Sometimes I find it more readable because I don't have to think about falsy/truthy values, but it's not that "it's better". 🤷🏽

gndelia:
Converting to Draft because I've just found a bug

gndelia:
> Converting to Draft because I've just found a bug

Solved in [03e15b4](https://github.com/hemilabs/ui-monorepo/pull/791/commits/03e15b49f60144d40e6b5ce9aae3d870685f84db)

gabmontes (approved):

ArturDolzan (approved):

Issue #755, opened 2025-01-14
Staking Campaign - Implement Withdraw Drawer


ArturDolzan:
### Feature or change description

Develop the Withdraw Drawer to allow users to manage and withdraw their staked assets.

gndelia:
Closing. Completed in related PRs - only thing pending is https://github.com/hemilabs/ui-monorepo/issues/880

PR #798, opened 2025-01-28
Add stake drawer

gndelia:
### Description

This PR adds the UI for the Stake drawer. Note that the submission code is not done yet, so that's why the issues are left open

#### Description by commit

- 5951639b125a92254fa0d9c203b97814c77acaed Creates `sanitizeAmount ` with the logic used in the tunnel for the input, and moves it into the `utils` folder, so it can be reused here. Tests added
- c04dc3f3c0420a932631730ee14d54f9689b491a Improves even further the `TokenInput`. As many props were part of the `TokenSelector`, I did some component composition and updated it so now `TokenSelector` is a prop. Now the interface of `TokenInput` is more clean and simple
- 5d1ea99ea35a2dd2efee46517cd7d58e602dad22 Moves the fees component that's rendered below the tunnel form, so I Can reuse it in the stake drawer. Now it is part of the `components` folder
- 7c5201d9aab081beb215bcec572212412529e76a The `MaxBalance` button component has a `isRunningOperation` prop to disable it. Its naming was confusing , so I updated it into using a more standard `disabled`.
- 6698d2f0a38a10a4875dcd6410aabc80f88ca885 Moves the `nuqs` parser `parseAsEvmAddress` into a common folder, so I can reuse it for opening the drawer. This parser is used to validate a query string is a valid EVM Address
- ab119a1598de1d1c2ad509ec825ab1c6e64e8bec based in all the previous changes, this is the one that actually creates the Drawer.
- cb5d1e31399b931a541f33df0b25619a44f0f05c Updates the `/stake` layout to allow the drawer to be opened from anywhere under the routes of this layout (which are `/stake` and `/stake/dashboard`). Depending the token selected, the query string `tokenAddress` is set (which uses the parser from the commit above), and the `drawerMode` query string value is used to give some context to the drawer, and show (or not) the "Stake/Unstake" tabs.  
With this commit, users clicking a row in the dashboard will open the Drawer with the "Unstake" tab active (See video)
- 55247e0b258d72eac55b6f652ed8bd0db0675eed updates the "Stake" in the "/stake" page to open the drawer, and the "+Stake more" to redirect to the Stake page

### Screenshots

<details>
  <summary>Mobile</summary>

image

image

image

</details>

<details>
  <summary>Desktop</summary>

image

image

image

</details>

https://github.com/user-attachments/assets/78ebebf9-8247-426d-b0aa-228563fbea15

(Note that the drawer without "tabs" is opened from the "Stake" tab)

### Related issue(s)

Related to #754 and #755

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/stake/_components/manageStake/website.tsx
+++ b/webapp/app/[locale]/stake/_components/manageStake/website.tsx
@@ -1,0 +1,42 @@
+import { ExternalLink } from 'components/externalLink'
+import { ArrowDownLeftIcon } from 'components/icons/arrowDownLeftIcon'
+import { useTranslations } from 'next-intl'
+import { StakeToken } from 'types/stake'
+
+const websitesMap: Partial<
+  Record<StakeToken['extensions']['protocol'], string>
+> = {
+  bedRock: 'https://www.bedrock.technology',
+  bitFi: 'https://www.bitfi.one/#waitlist',
+  exSat: 'https://exsat.network/app/bridge',
+  lorenzo: 'https://app.lorenzo-protocol.xyz/staking',
+  merlinChain: 'https://merlinchain.io/bridge',
+  // TODO add nodeDao website https://github.com/hemilabs/ui-monorepo/issues/794
+  pumpBtc: 'https://mainnet.pumpbtc.xyz',
+  solv: 'https://app.solv.finance/solvbtc?network=bitcoin-mainnet',
+  stakeStone: 'https://app.stakestone.io/u/sbtc/stake',
+  uniRouter: 'https://app.unirouter.io',
+}
+
+type Props = {
+  token: StakeToken
+}
+
+export const Website = function ({ token }: Props) {
+  const t = useTranslations('common')
+  return (
+    <>
+      {websitesMap[token.extensions.protocol] ? (
+        <ExternalLink
+          className="group/link flex cursor-pointer items-center gap-x-1 text-sm font-medium text-neutral-500"
+          href={websitesMap[token.extensions.protocol]}
+        >
+          <span className="group-hover/link:text-neutral-700">{t('open')}</span>
+          <ArrowDownLeftIcon className="[&>path]:fill-neutral-500 [&>path]:group-hover/link:fill-neutral-700" />
+        </ExternalLink>
+      ) : (
+        <span className="text-neutral-500">-</span>
+      )}
+    </>
+  )
+}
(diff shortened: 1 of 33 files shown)


dvnahuel:
LGTM!

webapp/app/[locale]/stake/_components/manageStake/website.tsx
```diff
@@ -0,0 +1,42 @@
...
+  bedRock: 'https://www.bedrock.technology',
+  bitFi: 'https://www.bitfi.one',
+  exSat: 'https://exsat.network',
+  lorenzo: 'https://lorenzo-protocol.xyz',
+  merlinChain: 'https://merlinchain.io',
+  nodeDao: 'https://www.nodedao.com',
+  pumpBtc: 'https://mainnet.pumpbtc.xyz',
+  solv: 'https://solv.finance',
+  stakeStone: 'https://stakestone.io/#/home',
+  uniRouter: 'https://www.unirouter.io/#',
```
Comment (gabmontes) [inline webapp/app/[locale]/stake/_components/manageStake/website.tsx:18]: Perhaps we should move this into the token-list too. Having part of the information there and part here seems an invitation to mismatches, missing data, etc.

webapp/app/[locale]/stake/_hooks/useEstimateStakeFees.ts
Comment (gabmontes) [inline webapp/app/[locale]/stake/_hooks/useEstimateStakeFees.ts:1]: This and `useEstimateUnstakeFees.ts` are so similar that it seems a good opportunity to merge both into one. Why should we keep both?

webapp/app/[locale]/stake/_components/manageStake/website.tsx
Comment (gndelia) [reply webapp/app/[locale]/stake/_components/manageStake/website.tsx:18]: I am unsure if we want to bloat the token list with this informational stuff, which is part of the portal. We also need to add the reward list mapped to each tokens. Would you prefer to move that into the token list? I'd rather not 🤔

webapp/app/[locale]/stake/_hooks/useEstimateStakeFees.ts
Comment (gndelia) [reply webapp/app/[locale]/stake/_hooks/useEstimateStakeFees.ts:1]: I can simplify them a bit, but note that the gasUnits for each hook may differ, so I'd still need 2 hooks

Comment (gabmontes) [reply webapp/app/[locale]/stake/_hooks/useEstimateStakeFees.ts:1]: Or one function and a map.

webapp/app/[locale]/stake/_components/manageStake/website.tsx
Comment (gndelia) [reply webapp/app/[locale]/stake/_components/manageStake/website.tsx:18]: As discussed offline, I convinced him and we're keeping this info in the portal, and keeping the token-list as is

webapp/app/[locale]/stake/_hooks/useEstimateStakeFees.ts
Comment (gndelia) [reply webapp/app/[locale]/stake/_hooks/useEstimateStakeFees.ts:1]: Fair suggestion. As discussed offline, I will do so in another PR as it requires updating the gas limits for other operations (the tunnel ones), so I'd rather do it in another PR. Issue here https://github.com/hemilabs/ui-monorepo/issues/801

gabmontes (dismissed):

gndelia:
Rebased to fix conflicts

PR #800, opened 2025-01-30
Add Strategy Table Component

### LINKED ISSUE #753: Staking Campaign - Implement Staking Strategies Table

Create the Staking Strategies table with the properly columns. The table should display detailed information about available staking options, including the asset, balance, and multiplier incentives (e.g., x3 Points).

### LINKED ISSUE #779: Make "Tunnel" and "Transaction History" Tabs Full Width
The "Tunnel" and "Transaction History" tabs currently do not utilize the full width of the screen. Updating these tabs to be full width will improve the user experience by making better use of available space.

**Expected Behavior:**

1. Both the "Tunnel" and "Transaction History" tabs should extend to the full width of the container/screen.
2. Content within the tabs should adjust proportionally to maintain proper alignment and spacing.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=9516-17619&t=t4rxkGsLM94kKqaD-1)
Expected behavior screenshot:
Image
Current behavior screenshot:
Image

ArturDolzan:
### Description

This PR introduces the Strategy Table implementation while also simplifying the table logic by removing unnecessary complexity, such as rowVirtualizer, and refactoring shared components with the Staking Asset Table.

### Screenshots

https://github.com/user-attachments/assets/3befd753-6796-470a-8174-4ef8435affe5

Captura de Tela 2025-01-30 às 08 59 10

Captura de Tela 2025-01-30 às 08 59 23

### Related issue(s)

Closes #753 
Closes #779 

### Checklist

- [x] Manual testing passed.


alvarengathomas:
@gndelia @gabmontes Do we really need to create the staking app in the same code as `ui-monorepo`?

I think this code is already too big and too complex to add another functionality.

Can we create it in another repository, but applying the same designs of `ui-monorepo`?

gndelia:
@alvarengathomas It's the same app (the portal). 

To me, if we move to another repo, we will add complexity (instead of reducing it) because then later we need merge all pages together into one app to deploy it as one site. It moves the complexity into the build/deploy process.

Besides, moving it into another repo won't make it smaller, as the amount of code would be the same (if not more, again, due to having to add extra processing, or libraries to reduce duplication across the repos, etc, etc).

webapp/app/[locale]/stake/_components/stakeStrategyTable/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+  // TODO - The learn how to stake on hemi link is TBD
+  return (
+    <div className="rounded-2.5xl relative z-20 -translate-y-32 bg-neutral-100 p-1 text-sm font-medium md:-translate-y-24">
+      <div className="flex w-full flex-wrap items-center justify-between gap-x-2 gap-y-2 px-3.5 py-2 md:flex-nowrap md:px-3">
+        <h5 className="flex-shrink-0 md:flex-grow-0 md:basis-auto">
+          {t('strategy')}
+        </h5>
+        <p className="flex text-sm font-normal text-neutral-600">
+          {t('new-here')}
+          <Link href={'https://hemi.xyz/'}>
```
Comment (gndelia) [inline webapp/app/[locale]/stake/_components/stakeStrategyTable/index.tsx:207]: not a blocker, but perhaps this should be moved to [hemi-socials](https://github.com/hemilabs/hemi-socials/blob/main/socials.json)

webapp/app/[locale]/stake/_components/tokenRewards.tsx
```diff
@@ -0,0 +1,26 @@
+import { ReactNode } from 'react'
+
+import { HemiTag } from './rewardTag'
+
+interface Props {
+  rewards: string[]
```
Comment (gndelia) [inline webapp/app/[locale]/stake/_components/tokenRewards.tsx:6]: Here you could keep the types consistent by importing StakeToken and

```suggestion
  rewards: StakeToken['extensions']['rewards']
```

This way, if you change `rewards` in `StakeToken`, these are kept in sync automatically

webapp/types/stake.ts
```diff
@@ -19,5 +19,8 @@ export type StakeProtocols = (typeof stakeProtocols)[number]
 export type StakeToken = EvmToken & {
   // EvmToken has a broad definition of "protocol", but for StakeToken let's make a
   // defined list of protocols that make a token a Stake one.
-  extensions: Omit<Extensions, 'protocol'> & { protocol: StakeProtocols }
+  extensions: Omit<Extensions, 'protocol'> & {
+    protocol: StakeProtocols
+    rewards: string[]
```
gndelia, line 24: Perhaps you can narrow the type of `rewards` for better type consistency. Let's say you define

```ts
Type Reward = 'hemi' | 'points'
```

Then you do 

```suggestion
    rewards: Reward[]
```


Whenever you refer to this type (using `Stake['extensions']['rewards']`, then you ensure that all type strings must be first defined here, and you can't add an incorrect defined type

webapp/app/[locale]/stake/_components/tokenRewards.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+// TODO: It's pending to implement some missing tags
+// Issue #799 https://github.com/hemilabs/ui-monorepo/issues/799
+const rewardComponentMap: Record<string, ReactNode> = {
+  hemi: <HemiTag />,
+}
```
Comment (gndelia) [inline webapp/app/[locale]/stake/_components/tokenRewards.tsx:13]: Same as [this comment](https://github.com/hemilabs/ui-monorepo/pull/800/files#r1935646330)

```suggestion
const rewardComponentMap: Record<StakeToken['extensions']['rewards'], ReactNode> = {
  hemi: <HemiTag />,
}
```

Plus, here it would prevent you from adding incorrect reward names if `rewards` is not typed as a string (See [this comment first](https://github.com/hemilabs/ui-monorepo/pull/800/files#r1935652293))

webapp/app/[locale]/stake/_components/tokenRewards.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export function TokenRewards({ rewards }: Props) {
+  if (!rewards || rewards.length === 0) {
+    return null
+  }
+  return (
+    <>
+      {rewards.map(reward => (
+        <div key={reward}>{rewardComponentMap[reward] || null}</div>
```
Comment (gndelia) [inline webapp/app/[locale]/stake/_components/tokenRewards.tsx:22]: If we do the updates mentioned of typings `rewards` (so it is not `string[]`), you can ensure this will always be defined, so we can remove `|| null`

```suggestion
        <div key={reward}>{rewardComponentMap[reward]}</div>
```

Nevertheless, if it wasn't defined, `rewardComponentMap[reward]` returns `undefined` which renders nothing, so it could be skipped as well.

Going even beyond that, I think if you add the key for each component where there are defined in line 12, like 

```tsx
const rewardComponentMap: Record<StakeToken['extensions']['rewards'], ReactNode> = {
  hemi: <HemiTag key="hemi" />,
}
```

you can remove the extra div, like

```tsx
<>
  {rewards.map(reward => rewardComponentMap[reward])}
</>
```

if you don't feel like defining the key above, you can use fragments to also skip the extra `div`

```tsx
<>
      {rewards.map(reward => <React.Fragment key={reward}>{rewardComponentMap[reward]}</React.Fragment>)
</>
```

gabmontes:
@ArturDolzan cc/ @gndelia over time we learned the monorepos are a much more efficient way to host related projects. We should default to a monorepo and go with a separate repo only when there is a good enough reason to do so (i.e. it is a library that can be forked or can evolve independently).

webapp/hooks/useStakeTokens.ts
```diff
@@ -4,11 +4,32 @@ import { isStakeToken } from 'utils/token'
 
 import { useHemi } from './useHemi'
 
+// TODO: It's pending to implement some missing badges
+// Issue #799 https://github.com/hemilabs/ui-monorepo/issues/799
+const tokenRewardMap: Record<string, string[]> = {
```
gndelia, line 9: here you can even do 
```suggestion
const tokenRewardMap: Record<StakeProtocols, StakeToken['extensions']['rewards']> = {
```

being `StakeProtocols` imported from `types/stake`

webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+      secondText: 'from-[#dd3818] via-[#e46830] to-[#ea8d61]',
+      starColor: '#e46830',
+    },
+  }
+
+  const selectedGradients = gradients[gradientMode]
+
+  function getStarIconClassName() {
+    if (textSize === 'text-6xl') {
+      return 'absolute -top-[54px] left-[9px] -translate-x-1/2'
```
Comment (gndelia) [inline webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx:39]: I think `-top-13` is [enabled in tailwind config](https://github.com/hemilabs/ui-monorepo/blob/22c1ce401fc1ef755afd4ad58a8fe58e4aae8202/webapp/tailwind.config.ts#L215)

```suggestion
      return 'absolute -top-13 left-2 -translate-x-1/2'
```
In addition to that, could we use `left-2` instead of `-[9px]` ? Does that px make a difference visually?

webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+    }
+    if (textSize === 'text-4xl') {
+      return 'absolute -top-[35px] left-[5px] -translate-x-1/2'
```
Comment (gndelia) [inline webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx:42]: Same as [this comment](https://github.com/hemilabs/ui-monorepo/pull/800/files#r1935676607), does this extra pixel in each direction make a difference visually?

```suggestion
      return 'absolute -top-9 left-2 -translate-x-1/2'
```

This will create extra CSS classes, it's ok if needed, but I am wondering if it is truly the case 😄

webapp/app/[locale]/stake/_components/stakeStrategyTable/index.tsx
Comment (gndelia) [reply webapp/app/[locale]/stake/_components/stakeStrategyTable/index.tsx:207]: (I know this is a temporal link until the proper one is defined, but by seeing this I wondered if we should have `hemi.xyz` in the `socials` package 😅 )

alvarengathomas:
@gabmontes Can we at least create the staking app inside the `ui-monorepo`, but outside the `webapp` folder?

We don't need to create another repository for that, I just want to create a separation between the tunnel code and staking code.

@gndelia For me nothing change reviewing this PR the way it is right now, but I'm thinking about the complexity that keeps growing in the `webapp` code and everybody on the team will have to maintain it in the future.

webapp/app/[locale]/stake/page.tsx
```diff
@@ -1,18 +1,45 @@
...
-    <div className="h-fit-rest-screen">
-      <PageTitle subtitle={t('stake.subtitle')} title={t('stake.title')} />
-    </div>
+    <>
+      <div
+        className="absolute inset-0 z-10"
+        style={{
+          background: `url(${welcomeStakeBackgroundImg.src}) lightgray 50% / cover no-repeat`,
+          mixBlendMode: 'overlay',
+          opacity: 0.8,
```
Comment (gndelia) [inline webapp/app/[locale]/stake/page.tsx:23]: There's tailwind classes for these 2:

`opacity-80` and `mix-blend-overlay`

webapp/messages/es.json
```diff
@@ -179,24 +179,29 @@
...
       "stake-more": "Apuesta más",
       "stake": "Apuesta",
       "stake-earn-points-label": "Apuesta y|ganar p|untos"
     },
+    "protocol": "Protocolo",
+    "rewards": "Recompensas",
     "stake": {
+      "action": "Acción",
+      "learn-how-to-stake-on-hemi": "Aprenda como apostar en Hemi",
+      "new-here": "Nuevo aquí?",
```
gndelia, line 199:
```suggestion
      "new-here": "¿Nuevo aquí?",
```

webapp/components/stakeTabs.tsx
```diff
@@ -22,15 +22,16 @@ const StakeTabsImpl = function () {
   const isInStake = pathname === `/${locale}/stake/`
 
   return (
-    <div className="flex items-center justify-center gap-x-4">
-      <Tabs>
+    <div className="relative z-20 flex items-center justify-center gap-x-4 sm:flex-col sm:gap-y-4">
+      <Tabs className="flex w-full [&>li>a]:w-full [&>li]:flex-1">
```
gndelia, line 26: See https://github.com/hemilabs/ui-monorepo/issues/779 it seems that this must be solved directly in the `Tabs` component

webapp/components/tabs.tsx
```diff
@@ -30,7 +30,7 @@ export const Tab = function ({
   return (
     <li
       className={`
-      box-border flex h-7 items-center rounded-md px-2 py-1 text-sm font-medium
+      box-border flex h-7 items-center rounded-md px-2 py-1 text-sm font-medium
```
gndelia, line 33: Extra space added. Let's remove it so we don't add extra changes in the history 😅  (making a bit more annoying when using `git blame`)

```suggestion
      box-border flex h-7 items-center rounded-md px-2 py-1 text-sm font-medium
```

webapp/app/[locale]/stake/_components/stakeStrategyTable/index.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/stakeStrategyTable/index.tsx:207]: I just checked with Ryan and he asked to hidden this paragraph until we have the link defined

webapp/app/[locale]/stake/_components/tokenRewards.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/tokenRewards.tsx:22]: Done. Thanks

webapp/app/[locale]/stake/_components/tokenRewards.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/tokenRewards.tsx:13]: Done. thanks

webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx:39]: Yes, we can go with 8px. done.

webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+  gradientMode = 'primary',
+  textSize = 'text-6xl',
+}: Props) {
+  const t = useTranslations('stake-page.dashboard')
+  const labelParts = t('stake-earn-points-label').split('|')
+
+  // This function is designed to render a gradient label with a special behavior for the letter 'i'.
+  // If the second part of the label (labelParts[1]) contains the letter 'i', a star icon is rendered
+  // above it. However, in cases where the translation does not include the letter 'i' (e.g., in Spanish),
+  // a conditional logic ensures that the label is rendered correctly without the star icon.
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx:20]: Done. thanks!

webapp/hooks/useStakeTokens.ts
gndelia: you're correct, Protocols was incorrect 😅  Thanks!

webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+    }
+    return ''
+  }
+
+  function getStarIconSize() {
+    if (textSize === 'text-4xl') {
+      return 16
+    }
+    return 22
+  }
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx:52]: Yes agreed. I've done some refactor to use group variant so that way we can get closer to the tailwind philosophy. Thanks

webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/stakeAndEarnPointsSpecialLabel.tsx:42]: I've done some refactor to use group variant so that way we can get closer to the tailwind philosophy.

webapp/app/[locale]/stake/page.tsx
Comment (ArturDolzan) [reply webapp/app/[locale]/stake/page.tsx:23]: Done. thanks

webapp/messages/es.json
ArturDolzan: Fixed. Thanks.

webapp/components/tabs.tsx
ArturDolzan: Ops. fixed. thanks

dvnahuel:
LGTM!

webapp/components/stakeTabs.tsx
ArturDolzan: Oh ok, fixed in the Tabs component. I also set #779 to be closed when merging this PR.

webapp/components/tabs.tsx
```diff
@@ -63,5 +63,9 @@ type TabsProps = {
 }
 
 export const Tabs = ({ children }: TabsProps) => (
-  <ul className="flex flex-wrap items-center gap-x-2 gap-y-1">{children}</ul>
+  <ul
+    className={`flex w-full flex-wrap items-center gap-x-2 gap-y-1 [&>li>a>div]:justify-center [&>li>a>span]:flex [&>li>a>span]:justify-center [&>li>a]:w-full [&>li]:flex-1 md:[&>li]:flex-auto`}
```
gndelia, line 67: can all the `[&>li]` and similar be moved to the appropiate HTML tag in `Tab` ? It seems to me that yes

ArturDolzan: Reduced a bit these selectors

webapp/components/tabs.tsx
```diff
@@ -63,5 +63,7 @@ type TabsProps = {
 }
 
 export const Tabs = ({ children }: TabsProps) => (
-  <ul className="flex flex-wrap items-center gap-x-2 gap-y-1">{children}</ul>
+  <ul className="flex w-full flex-wrap items-center gap-x-2 gap-y-1 [&>li>a]:w-full [&>li]:flex-1 md:[&>li]:flex-auto">
```
gndelia, line 66: I must insist: I think this can and should be moved into the tab component.

The `li` tag can add `flex-1` and `md:flex-auto`, and add `[&>a]:w-full`

```suggestion
  <ul className="flex w-full flex-wrap items-center gap-x-2 gap-y-1">
```

ArturDolzan: Done.

gndelia (dismissed):

ArturDolzan:
> @ArturDolzan conflicts!

@gndelia rebased. Thanks!

gndelia (approved):

gabmontes (approved):

Issue #743, opened 2025-01-13
Improve transaction history sync process


alvarengathomas:
### Description

As a user, I want to see my transaction history as soon as possible, so I don't have to wait a long time until the application load all my transactions.

### Problem addressed

![Image](https://github.com/user-attachments/assets/552cd3fc-d033-485a-8dab-0cf034fc038c)

The sync process takes a long time to load all the transactions because it is reading all blocks in the chain by chunks.

### Proposed implementation

An alternative solution to fix this problem is to use Blockscout Explorer API to return all transactions related to the user address.

Reference: https://testnet.explorer.hemi.xyz/api-docs 
Endpoint: `/addresses/{address_hash}/transactions`

gndelia:
Note that we will need to get transactions for the user from the L1 chains as well

gabmontes:
Searching through all the transactions to the tunnel contract initiated by the user may not cover all the scenarios. These two have to be considered too:

- The user may have called a 3rd contract that makes the call to the tunnel contract.
- Another user may have called that 3rd contract and the tunnel operation done affects the user's balances.

While the second option is unlikely and may not be even possible with the current tunnel implementation, the first has to be considered for sure.

gndelia:
We're trying to implement [Subgraphs](https://thegraph.com/docs/en/indexing/tooling/graph-node/) to handle this problem. After all, it fits their use case: indexing information from the blockchain. This exposes a GraphQL API that we could consume from the transaction history. We would still save the TXs in local storage because the subgraph won't return the status of the operations (for example: it won't return if a withdrawal is ready to prove, or already finalized). After all, this information doesn't depend on the indexed data, but on the current view of the chain.  
For this reason, the web workers used to sync history are still needed, though they definitely will be simpler.

The burden of indexing is moved to indexers of the subgraph (TBD if we need to [signal our subraph](https://thegraph.com/docs/en/subgraphs/developing/publishing/publishing-a-subgraph/#adding-signal-to-your-subgraph) or if we should index it ourselves), and then the GraphQL query requires an API key, which is paid. See [subgraph query billing]/https://thegraph.com/docs/en/subgraphs/billing/). However, it should require significantly fewer queries than the current approach, as the current implementation indexes the entire blockchain per user, while this new approach queries information that is already indexed once, and stored.

The implementation of this should go  like this


- [ ] Create a subgraph for Deposits (Sepolia and Mainnet).
[...]

PR #805, opened 2025-02-03
Get Sepolia/Mainnet deposits through a subgraph

gndelia:
### Description

This PR creates a subgraph (which is deployed in [The Graph Studio](https://thegraph.com/studio/) and needs to be published before production use) to retrieve ETH and ERC20 deposits from users.  

The first commit d1569cc3c74de1f75f93590a260b96e2239067eb takes care of the creation of the subgraph, its different commands (See the Readme file), and local usage. I recommend checking [The Graph Docs](https://thegraph.com/docs/en/) to understand better its usage.  
The summary is that it is an indexing service that lets you run commands on custom Log events from transactions, and store that in a blockchain, so queries can be run against it. As the info is in the blockchain, this protocol is decentralized. Billing needs to be set to query it in production. However, by using The Graph Studio (which is a centralized version that is thought to be used for testing/staging), we can get a general idea of how it works. **Note that the subgraph code is written in AssemblyScript, not in Typescript**, which is a subset of the latter (And more limited). See [AssemblyScript docs](https://www.assemblyscript.org/introduction.html).

The second commit eef77e9f0969f2073ddb3c6632de069ea07116b8 updates the portal to consume this subgraph. The subgraph exposes the info through a GraphQL API. I added a feature flag to test this functionality, in case we need to quickly revert it for whatever reason. I think it is still reasonable to keep saving in local storage the info already synced in the past for users, so the workers will continue to exist, though some fields can be removed if we move forward with this path.

Although there are no changes to users, I hope the video shows you how fewer requests are now needed to sync the whole blockchain. Bear in mind that the "status" isn't a historical value that needs to be indexed, as it only exists in the "present", so for withdrawals, we will still need to run many requests per transaction to check its status. Nevertheless, with the subgraph, we can forgive all the `eth_getLogs` calls.

If this approach works on staging, the next steps are

- Publishing the subgraph (This means moving it out of The Graph Studio into a decentralized network, that anyone can query)
- Set billing for the published subgraph, and update the portal to use the published subgraph in Staging
- If it works ok, try it in production
- If it works ok, update the other calls (Bitcoin Withdrawals, Hemi withdrawals) into subgraphs as well

The last commit bc6310e updates the GitHub workflows to include the required Env Variables

### Screenshots

https://github.com/user-attachments/assets/1a0ffdab-1609-455f-843d-d8b06fc7b42a

Note that fewer requests are needed when resyncing the whole history for deposits. And if the users refresh the page, requests are only made for searching deposits since the last block that was checked the last time.

### Related issue(s)

Related to https://github.com/hemilabs/ui-monorepo/issues/743

### Checklist

- [x] Manual testing passed.


subgraphs/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+services:
+  graph-node:
+    image: graphprotocol/graph-node
+    ports:
+      - '8000:8000'
+      - '8001:8001'
+      - '8020:8020'
+      - '8030:8030'
+      - '8040:8040'
```
gabmontes, line 9: Take a look at the changes proposed in the following PR for improvement opportunities here:

https://github.com/graphprotocol/graph-node/pull/5763/files

subgraphs/tunnel-deposits-subgraph/subgraph.yaml
```diff
@@ -0,0 +1,29 @@
...
+      entities:
+        - Deposit
+      abis:
+        - name: L1StandardBridge
+          file: ./abis/L1StandardBridge.json
+      eventHandlers:
+        - event: ETHDepositInitiated(indexed address,indexed address,uint256,bytes)
+          handler: handleETHDeposit
+        - event: ERC20DepositInitiated(indexed address,indexed address,indexed
+            address,address,uint256,bytes)
```
gabmontes, line 27: The Blockscout explorer uses `TransactionDeposited` to identify all the deposits. Let's analyze using that event instead.

https://github.com/hemilabs/blockscout/blob/5b1e9b1c8883cfa926674a8993f5f7fc304c1a4d/apps/indexer/lib/indexer/fetcher/optimism/deposit.ex#L37

gndelia: I went for Deposit Initiated as [this is the Event used by the OP SDK](https://github.com/ethereum-optimism/optimism-legacy/blob/8205f678b7b4ac4625c2afe351b9c82ffaa2e795/packages/sdk/src/adapters/eth-bridge.ts#L36), which is the package we based our code from

subgraphs/docker-compose.yml
gndelia: I will apply them

gndelia: Applied here [14a86b8](https://github.com/hemilabs/ui-monorepo/pull/805/commits/14a86b84dec64532489d9f0fa0c628e664afd452)

subgraphs/tunnel-deposits-subgraph/subgraph.yaml
gndelia: Note that the `TransactionDeposited` event is missing some information that the other event does have, particularly the L1, L2 token addresses and the amount.

See

```

ETHDepositInitiated (index_topic_1 address _l1Token, index_topic_2 address _l2Token, index_topic_3 address _from, address _to, uint256 _amount, bytes _data)

TransactionDeposited (index_topic_1 address from, index_topic_2 address to, index_topic_3 uint256 version, bytes opaqueData)
```

Perhaps we can parse them from the TransactionInput, but maybe that means extra work as we need to parse the input

subgraphs/README.md
```diff
@@ -0,0 +1,65 @@
...
+npm run deploy-local:sepolia
+# deploys mainnet, if the container started is a mainnet one.
+npm run deploy-local:mainnet
+```
+
+After deploying, the terminal will output the URL for querying the subgraph through HTTP requests.
+
+## Deploying to Subgraph studio
+
+The following commands are to deploy a subgraph in [the graph studio](https://thegraph.com/studio/). This environment is intended for testing and staging. Note that subgraphs deployed here are archived after 45 days if a newer version isn't published, so they are considered short term. [Further info](https://thegraph.com/docs/en/subgraphs/developing/deploying/multiple-networks/#subgraph-studio-subgraph-archive-policy).
```
gabmontes, line 38:
```suggestion
The following commands deploy a subgraph in [Subgraph Studio](https://thegraph.com/studio/). This environment is intended for testing and staging. Note that subgraphs deployed there are archived after 45 days unless a newer version is published, so they are considered short term. Read their [archive policy].(https://thegraph.com/docs/en/subgraphs/developing/deploying/multiple-networks/#subgraph-studio-subgraph-archive-policy).
```

subgraphs/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+[See Graph studio docs](https://thegraph.com/docs/en/subgraphs/developing/deploying/using-subgraph-studio/) for further info. Here's a quick summary:
```
gabmontes, line 40:
```suggestion
[See Subgraph Studio docs](https://thegraph.com/docs/en/subgraphs/developing/deploying/using-subgraph-studio/) for further info. Here's a quick summary:
```

subgraphs/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+First, an account in the graph studio needs to be created. Once there, a subgraph needs to be manually created. A subgraph in The Graph Studio needs to be created per subgraph/chain defined here in the code.
```
gabmontes, line 42:
```suggestion
First, an account in the Subgraph Studio needs to be created. Once there, a subgraph needs to be manually created. A subgraph in The Subgraph Studio needs to be created per subgraph/chain defined here in the code.
```

subgraphs/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+The subgraph naming convention should be
+
+`${package.json}.name-${chain}`. For example, in order to deploy the subgraph `tunnel-deposits-subgraph` in Sepolia, the subgraph `tunnel-deposits-subgraph-sepolia` must be created in the Graph Studio.
```
gabmontes, line 46:
```suggestion
The subgraph naming convention should be `${package.json}.name-${chain}`. For example, in order to deploy the subgraph `tunnel-deposits-subgraph` in Sepolia, the subgraph `tunnel-deposits-subgraph-sepolia` must be created in the Subgraph Studio.
```

gabmontes (changes requested):

gndelia:
@gabmontes  All readme suggestions applied here [72d94a8](https://github.com/hemilabs/ui-monorepo/pull/805/commits/72d94a875d2cea1ec1c342e4bfbafb48de7a2ce1)

gabmontes (approved):

ArturDolzan (approved):

Issue #806, opened 2025-02-03
Implement Custom Staking Page for Testnet


dvnahuel:
When users access the staking page on Testnet, it should display a customized version of the page that informs them they can only stake on Mainnet.

Design [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=9959-11373&t=qZfJ8zaDNWT7RRHL-1)

Issue #774, opened 2025-01-17
Define functions to stake/unstake


gndelia:
The functions that call the Smart contracts to stake/unstake need to be implemented. Currently, there's no implementation defined. I defined an interface inspired by a protocol _that should remain nameless_ but the real implementation needs to be added. TBD if it needs to be moved to `hemi-viem` or somewhere else

gndelia:
Include estimating the appropriate amount of units for Fees for both operations as part of this ticket

PR #808, opened 2025-02-04
Add testnet tokens for Stake

gndelia:
### Description

This PR adds testnet tokens for the Stake campaign. AFAIK these tokens are only for us to test the contracts, and we don't really expect them to be used by users in prod. Because of that, I added an Env Variable so we can enable this only locally. I did so in the first commit e6165de34bb99cf3f9bfc7e21b3fae04a5a475ea

The second commit simplifies the check for the feature flag for the whole stake campaign a bit, moving it into the layout page for the `/stake` routes.

### Screenshots

Now DAI and USDT are enabled if the feature flag is enabled on testnet

image

image

image

### Related issue(s)

Related to https://github.com/hemilabs/ui-monorepo/issues/806
Related to https://github.com/hemilabs/ui-monorepo/issues/774

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/stake/dashboard/page.tsx
+++ b/webapp/app/[locale]/stake/dashboard/page.tsx
@@ -1,6 +1,5 @@
 'use client'
 
-import { featureFlags } from 'app/featureFlags'
 import { PageTitle } from 'components/pageTitle'
 import { useStakeTokens } from 'hooks/useStakeTokens'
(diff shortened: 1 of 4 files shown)


ArturDolzan (approved):

gabmontes (approved):
uACK

PR #825, opened 2025-02-06
Implement stakeToken and unstakeToken Functions

ArturDolzan:
### Description

This PR implements three functions for staking and unstaking tokens in the staking contract.

- **stakeERC20Token**, which allows users to stake ERC-20 tokens by calling depositFor.
- **stakeETHToken**, which enables staking of native ETH via depositETHFor.
- **unstakeToken**, which handles unstaking for both ERC-20 and native ETH

### Related issue(s)

Related to #774 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/hemi-viem-stake-actions/actions/wallet/stakeManager.ts
+++ b/hemi-viem-stake-actions/actions/wallet/stakeManager.ts
@@ -1,0 +1,102 @@
+import { type Address, type Client } from 'viem'
+import { writeContract } from 'viem/actions'
+
+import {
+  stakeManagerAbi,
+  stakeManagerAddresses,
+} from '../../contracts/stakeManager'
+
+/**
+ * Performs staking of an ERC-20 token by calling the `depositFor` function on the contract.
+ *
+ * @param {Client} client - The Viem client configured to interact with the blockchain.
+ * @param {Object} parameters - The transaction parameters for staking.
+ * @param {bigint} parameters.amount - The amount of tokens to be staked.
+ * @param {Address} parameters.forAccount - The user's address initiating the staking.
+ * @param {Address} parameters.tokenAddress - The ERC-20 token address being staked.
+ *
+ * @returns {Promise<Address>} - The transaction sent to the blockchain.
+ */
+export function stakeERC20Token(
+  client: Client,
+  parameters: {
+    amount: bigint
+    forAccount: Address
+    tokenAddress: Address
+  },
+) {
+  const { amount, forAccount, tokenAddress } = parameters
+  return writeContract(client, {
+    abi: stakeManagerAbi,
+    account: forAccount,
+    // @ts-expect-error: TS is complaining about client.chain!.id definition, but this works
+    address: stakeManagerAddresses[client.chain!.id],
+    args: [tokenAddress, forAccount, amount],
+    // @ts-expect-error: TS is complaining about client.chain definition, but this works
+    chain: client.chain,
+    functionName: 'depositFor',
+  })
+}
+
+/**
+ * Performs staking of native ETH by calling the `depositETHFor` function on the contract.
+ *
+ * @param {Client} client - The Viem client configured to interact with the blockchain.
+ * @param {Object} parameters - The transaction parameters for staking.
+ * @param {bigint} parameters.amount - The amount of tokens to be staked.
+ * @param {Address} parameters.forAccount - The user's address initiating the staking.
+ *
+ * @returns {Promise<Address>} - The transaction sent to the blockchain.
+ */
+export function stakeETHToken(
+  client: Client,
+  parameters: {
+    amount: bigint
+    forAccount: Address
+  },
+) {
+  const { amount, forAccount } = parameters
+  return writeContract(client, {
+    abi: stakeManagerAbi,
+    account: forAccount,
+    // @ts-expect-error: TS is complaining about client.chain!.id definition, but this works
+    address: stakeManagerAddresses[client.chain!.id],
+    args: [forAccount],
+    // @ts-expect-error: TS is complaining about client.chain definition, but this works
+    chain: client.chain,
+    functionName: 'depositETHFor',
+    value: amount,
+  })
+}
+
+/**
+ * Unstakes a previously staked token by calling the `withdraw` function on the contract.
+ *
+ * @param {Client} client - The Viem client configured to interact with the blockchain.
+ * @param {Object} parameters - The transaction parameters for unstaking.
+ * @param {bigint} parameters.amount - The amount of tokens to be unstaked.
+ * @param {Address} parameters.forAccount - The user's address initiating the unstaking.
+ * @param {Address} parameters.tokenAddress - The token address to be unstaked.
+ *
+ * @returns {Promise<Address>} - The transaction sent to the blockchain.
+ */
+export function unstakeToken(
+  client: Client,
+  parameters: {
+    amount: bigint
+    forAccount: Address
+    tokenAddress: Address
+  },
+) {
+  const { amount, forAccount, tokenAddress } = parameters
+  return writeContract(client, {
+    abi: stakeManagerAbi,
+    account: forAccount,
+    // @ts-expect-error: TS is complaining about client.chain!.id definition, but this works
+    address: stakeManagerAddresses[client.chain!.id],
+    args: [tokenAddress, amount],
+    // @ts-expect-error: TS is complaining about client.chain definition, but this works
+    chain: client.chain,
+    functionName: 'withdraw',
+  })
+}
--- a/webapp/utils/stake.ts
+++ b/webapp/utils/stake.ts
@@ -1,10 +1,11 @@
+import hemilabsTokenList from '@hemilabs/token-list'
 import {
   type HemiPublicClient,
   type HemiWalletClient,
 } from 'hooks/useHemiClient'
 import { NetworkType } from 'hooks/useNetworkType'
 import { EvmToken } from 'types/token'
-import { getErc20TokenBalance } from 'utils/token'
+import { getErc20TokenBalance, isNativeToken } from 'utils/token'
 import { type Address, Chain, parseUnits } from 'viem'
 
 export const isStakeEnabledOnTestnet = (networkType: NetworkType) =>
@@ -50,17 +51,17 @@ export const canSubmit = function ({
 
 const validateOperation = async function ({
   amount,
-  from,
+  forAccount,
   hemiPublicClient,
   token,
 }: {
   amount: bigint
-  from: Address
+  forAccount: Address
   hemiPublicClient: HemiPublicClient
   token: EvmToken
 }) {
   const balance = await getErc20TokenBalance({
-    address: from,
+    address: forAccount,
     // It works in runtime, but Typescript fails to interpret HemiPublicClient as a Client.
     // I've seen that the typings change in future viem's version, so this may be soon fixed
     // @ts-expect-error hemiPublicClient is Client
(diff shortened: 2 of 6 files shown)


hemi-viem-stake-actions/actions/wallet/stakeManager.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @param {Address} parameters.token - The ERC-20 token address being staked.
+ *
+ * @returns {Promise<void>} - The transaction sent to the blockchain.
+ */
+export function stakeERC20Token(
+  client: Client,
+  parameters: {
+    amount: bigint
+    from: Address
+    token: Address
```
gndelia, line 25:
```suggestion
    tokenAddress: Address
```

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Performs staking of an ERC-20 token by calling the `depositFor` function on the contract.
+ *
+ * @param {Client} client - The Viem client configured to interact with the blockchain.
+ * @param {Object} parameters - The transaction parameters for staking.
+ * @param {bigint} parameters.amount - The amount of tokens to be staked.
+ * @param {Address} parameters.from - The user's address initiating the staking.
+ * @param {Address} parameters.token - The ERC-20 token address being staked.
+ *
+ * @returns {Promise<void>} - The transaction sent to the blockchain.
```
gndelia, line 18: I think the TX hash is returned

```suggestion
 * @returns {Promise<Address>} - The transaction sent to the blockchain.
```

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Performs staking of native ETH by calling the `depositETHFor` function on the contract.
+ *
+ * @param {Client} client - The Viem client configured to interact with the blockchain.
+ * @param {Object} parameters - The transaction parameters for staking.
+ * @param {bigint} parameters.amount - The amount of tokens to be staked.
+ * @param {Address} parameters.from - The user's address initiating the staking.
+ *
+ * @returns {Promise<void>} - The transaction sent to the blockchain.
```
gndelia, line 49: Same here as in `stakeERC20Token`

```suggestion
 * @returns {Promise<Address>} - The transaction sent to the blockchain.
```

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Unstakes a previously staked token by calling the `withdraw` function on the contract.
+ *
+ * @param {Client} client - The Viem client configured to interact with the blockchain.
+ * @param {Object} parameters - The transaction parameters for unstaking.
+ * @param {bigint} parameters.amount - The amount of tokens to be unstaked.
+ * @param {Address} parameters.from - The user's address initiating the unstaking.
+ * @param {Address} parameters.token - The token address to be unstaked.
+ *
+ * @returns {Promise<void>} - The transaction sent to the blockchain.
```
gndelia, line 81: Same here

```suggestion
 * @returns {Promise<Address>} - The transaction sent to the blockchain.
```

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+export function unstakeToken(
+  client: Client,
+  parameters: {
+    amount: bigint
+    from: Address
+    token: Address
```
gndelia, line 88:
```suggestion
    tokenAddress: Address
```

webapp/utils/stake.ts
```diff
@@ -140,15 +147,19 @@ export const unstake = async function ({
...
     hemiPublicClient,
     token,
   })
+
   return hemiWalletClient.unstakeToken({
     amount: amountUnits,
     from,
-    token,
+    token: isNativeToken(token)
+      ? zeroAddress
```
gndelia, line 162: Per the contracts code (See [here](https://github.com/hemilabs/hemi-staking-contracts/blob/0d62d21dd8b2969d6d6da83316e69fe390e819cf/contracts/ZtakingPool.sol#L84)), I think when staking, the user sends ETH and the balance is updated on Wrapped Ether. For unstaking, then you need to send the address of WETH - See the [withdraw function](https://github.com/hemilabs/hemi-staking-contracts/blob/0d62d21dd8b2969d6d6da83316e69fe390e819cf/contracts/ZtakingPool.sol#L96)

WETH is available in our token list - See [here](https://github.com/hemilabs/token-list/blob/8c6b194085c8bf8d79bad9d2107200e7b0499998/src/hemi.tokenlist.json#L27).

My only doubt is if unstaking ETH would return WETH to the user (instead of ETH) - am I interpreting the code correctly @gabmontes  ?

gndelia (changes requested):

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
ArturDolzan: Done.

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
ArturDolzan: Done.

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
ArturDolzan: Done.

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
ArturDolzan: Done.

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
ArturDolzan: Done.

webapp/utils/stake.ts
ArturDolzan: I just did a quick test after changing to Wrapped Ether and I got the Sep ETH back to my wallet.

Captura de Tela 2025-02-07 às 09 07 32

gabmontes: There is no "withdraw ETH" option in the contract's code. If the user deposits ETH or WETH, he can only withdraw WETH.

webapp/utils/stake.ts
```diff
@@ -140,15 +148,30 @@ export const unstake = async function ({
...
+  // ERC-20 token interactions, and WETH is the wrapped version of ETH that follows
+  // the ERC-20 standard.
+  const tokenAddress: `0x${string}` = isNativeToken(token)
+    ? (hemilabsTokenList.tokens.find(
+        item => item.chainId === token.chainId && item.symbol === 'WETH',
+      )?.address as `0x${string}`) ??
+      (function () {
+        throw new Error('WETH token not found')
+      })()
+    : (token.address as `0x${string}`)
```
gndelia, line 170: minor, but perhaps let's better use viem's `Address` instead of `0x${string}`.

Then, this ternary is a bit complex to read, perhaps you can split it into a function with if/else?

webapp/utils/stake.ts
```diff
@@ -108,10 +109,17 @@ export const stake = async function ({
...
+      amount: amountUnits,
+      from,
+    })
+  }
+
+  return hemiWalletClient.stakeERC20Token({
     amount: amountUnits,
     from,
-    token,
+    tokenAddress: token.address as `0x${string}`,
```
gndelia, line 122: same here re: using `Address`

gndelia (commented):
Left one minor comment, but overall looks good

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
```diff
@@ -0,0 +1,102 @@
...
+  parameters: {
+    amount: bigint
+    from: Address
+    tokenAddress: Address
+  },
+) {
+  const { amount, from, tokenAddress } = parameters
+  return writeContract(client, {
+    abi: stakeManagerAbi,
+    account: from,
```
gabmontes, line 31: If the viem client does not have a wallet initialized for "from", this call will fail. I think we should not set this property at all and let the wallet client do its job.

webapp/utils/stake.ts
gndelia: > There is no "withdraw ETH" option in the contract's code. If the user deposits ETH or WETH, he can only withdraw WETH.

I feel that it adds burden on the users 😆  Either way

- I'd solve that on a contract level
- As discussed, we may not even solve it because ETH may not be supported 😆

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
```diff
@@ -0,0 +1,102 @@
...
+    // @ts-expect-error: TS is complaining about client.chain!.id definition, but this works
+    address: stakeManagerAddresses[client.chain!.id],
+    args: [tokenAddress, from, amount],
```
gabmontes, line 34: Let's try to keep the variable names matching the names used in the contracts. So "from" -> "for".

Having said that, even when the contracts support depositing tokens and adding those to other's balance, we may want to simplify our client-side code and always use the user's account.

webapp/utils/stake.ts
ArturDolzan: Created the function `getTokenAddress` to do this logic.

webapp/utils/stake.ts
ArturDolzan: Done.

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
gndelia: _I think_ in the latest version of `viem` this field is required. See [account](https://viem.sh/docs/contract/writeContract.html#account) (although null is acceptable)

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
ArturDolzan: Renamed to forAccount as for is a reserved word.

hemi-viem-stake-actions/actions/wallet/stakeManager.ts
ArturDolzan: Regarding setting account to null (check the [doc here](https://viem.sh/docs/contract/writeContract)), it's causing an side effect that I was not able to figure out yet. Somehow, the hook that controls the wallet connection is not handling this well (or perhaps a viem issue?), even setting account manually. All tests were done connected to hemi sepolia properly, but still without success.

This is a log when the transaction is properly sent:
Captura de Tela 2025-02-07 às 18 19 42

and this is a log when the transaction is sent wrongly:
Captura de Tela 2025-02-07 às 18 20 01

Note that is missing the parameter `from`. I'll try to figure out this issue by tomorrow, anyway I've pushed all other fixes so you guys can tackle this issue or even split in another PR and merge this one, which works better for you.

gndelia: Checking the source code for [writeContract](https://github.com/wevm/viem/blob/5689b084ee2dfe0ba75eca64e412da8ce06595fa/src/actions/wallet/writeContract.ts#L176), it seems that if `account` is not provided, it defaults to `client.account`, where `client` here looks to be `walletClient`. Perhaps you need to see if `walletClient` has `walletClient.account` set. 

If not, perhaps you can set it on creation time - [createWalletClient](https://viem.sh/docs/clients/wallet.html#account-optional) accepts an `account` parameter, and we can set it from the `address` coming from [wagmi's useAccount()](https://wagmi.sh/react/api/hooks/useAccount#address). I think  that we're calling `createWalletClient` through wagmi's [useWalletClient here](https://github.com/hemilabs/ui-monorepo/blob/25c7cf9431f851011f819273de1896a712b23183/webapp/hooks/useHemiClient.ts#L52).

ArturDolzan: Yep I've tried passing the account parameter as you mentioned during the wallet creation but I got the same error. Will check it out again carefully though.

gndelia: I just tested this and it worked. [Here](https://testnet.explorer.hemi.xyz/tx/0x820fd3bd8badcfb1b04d42efeb2910003c2d4dc4b3bec6265b5dea3d6d6cfe89) is a stake transaction

gndelia (approved):

Issue #836, opened 2025-02-11
Selecting USDT or DAI in the tunnel crashes the app


gndelia:
### Description

On testnet, selecting DAI or USDT crashes

<img width="1585" alt="Image" src="https://github.com/user-attachments/assets/255b97a7-3142-4609-8074-4152eb8c8d82" />

### Steps to reproduce

On the testnet tunnel, select either DAI or USDT to tunnel from Sepolia to Hemi Sepolia. The app crashes

### Environment

- **OS**: _e.g., Windows, macOS, Linux_
- **Runtime/Browser/Version**: _e.g., Node 22, Chrome 120_
- **Wallet**: _e.g., MetaMask, Trust, Coinbase Wallet_

PR #845, opened 2025-02-13
Whitelist tokens per operation (Stake or Tunnel)

gndelia:
### Description

This PR adds whitelists for "Stake" and "Tunnel". Before this, tokens could either work on Stake, or Tunnel, but not both. This caused the bug mentioned in the issue, where USDT and DAI were enabled for Stake testnet, and broke the tunnel. With this change, they can now be selected for tunneling. Given the whitelisting does not belong to the token list (which will be an "official" list of listed tokens in Hemi), the whitelisting is done here. In addition to that, some metadata is added.

#### Description by commit

This may be easier to review by commit:

- cbba3ab4c0f8f59e56e4982cfe2afd2e10b60b90 This commit moves the `tokenList.ts` into its own folder.
- 5462bc7cbee753f1796b73c18d341738b80bea02 splits the `nativeTokens` list into its own file. This was needed because in the next commit, some changes were applied that caused a circular reference when importing functions. Same way, the `utils/token.ts` was split into 2 files, and the functions related to native tokens were moved into their own file: `utils/nativeToken.ts`
- f70f0620d756e6f20808e56b277c50a7438177ce Adds the whitelist for Stake, and for Tunnel. As Stake requires some extra metadata (Website, protocol, reward), given this list, I decided to better centralize everything here, instead of having the info scattered through the app.   
For the tunnel, we need to both whitelist the Hemi token and the L1 remote one. I did not want to manually list both addresses, as it could be cumbersome and cause wrong states (what if the L1 address was skipped?). So I did some extra work to ensure that we only whitelist the Hemi addresses, and the remote versions are retrieved. This way, no "half-states" are possible.
- 47f1573175ab7d3ee2f87a8e00d38837cd5a81f6 Adds a few tests

This change also means that publishing a token into the official token list is not enough for it to become enabled in the tunnel or the stake page. It also means that we manually need to add it to the appropriate white list here. As publishing a token also requires us to come here and update the version, it's not that it will be easy for us to forget about it!.

### Screenshots

Note that now all tokens are visible in all chains for both the Tunnel and Stake

https://github.com/user-attachments/assets/af832753-ebed-4b93-9ea5-290513a17275

### Related issue(s)

Closes https://github.com/hemilabs/ui-monorepo/issues/836

### Checklist

- [x] Manual testing passed.


gabmontes (dismissed):

gndelia:
@gabmontes  requesting re-review as I added this new commit [a94dbee](https://github.com/hemilabs/ui-monorepo/pull/845/commits/a94dbeea7b6026f4649bdcc4e436de9b4d81e592) with another whitelist for tunneling through custom partners (we had this in a component + the token list, but now that the whitelist has been added, I felt it made sense to move the functionality here)

gabmontes (approved):

alvarengathomas (approved):

Issue #796, opened 2025-01-28
Fetch token prices in USD


gndelia:
### Feature or change description

Using either Coingecko or CoinMarketCap APIs, we should bring the stake tokens prices in USD

gndelia:
### Coingecko

> CoinGecko API offers both free and paid plans. The Demo API plan is accessible to all CoinGecko users at zero cost, with a stable rate limit of 30 calls/min and a monthly cap of 10,000 calls. Paid plans start at $129/mo, offering a higher rate limit of 500 calls/min and a monthly cap of 500,000 calls.

See [Pricing](https://www.coingecko.com/en/api/pricing) for paid versions

<img width="1362" alt="Image" src="https://github.com/user-attachments/assets/99ad010e-6d43-4fee-b5da-aab02d0fcc36" />

Note that  the `*` states

> * For 1 product only, up to 100k users

and that the Free version is for _personal use_

### Coinmarketcap

See [Pricing](https://pro.coinmarketcap.com/api/pricing)

<img width="1289" alt="Image" src="https://github.com/user-attachments/assets/f1344d9b-035d-426f-ab60-963a08bfb188" />

Note that free tiers may not have the Commercial License and that

> Your product shall prominently display the message 'Data provided by CoinGecko' and include a hyperlink to https://www.coingecko.com/en/api. Learn more about our branding guide here [here](https://brand.coingecko.com/)

rmilrad:
@gndelia @gabmontes How many calls do you estimate we run every month? Im assuming we have to make a call for the asset every time a user goes to stake / unstake, as well as every X minutes to display current conversion rates / TVL data on the dashboard?

Is a baseline estimate as simple as (lets say for 10 assets):

10 assets called every 5 min x 30 days = 86,500 calls? (ie the number of times we call all assets to get updated pricing to display TVL and estimated asset conversions on the dashboard). 

Right now, we have 380k EOAs, 50k addresses making daily txns on Hemi, and about 3k daily via Portal. assuming about 1/3 of active addresses click to stake (which is a generous number), can we estimate another 200k calls (one call for both stake and unstake for ONE asset per user average). 

I think the 300k-1M call / month tier should be appropriate?

gndelia:
@rmilrad if I understand correctly, these calls would be on Ethereum Mainnet (as there is where the tokens exists).

[...]

PR #857, opened 2025-02-14
Add the token prices service

gabmontes:
### Description

This PR adds the service to fetch and retrieve the token prices from the pricing APIs. In a next PR, the deployment scripts will be added. See the README for all the information related to this service.

The goal was to make it as simple as possible while keeping it robust enough for production usage.

### Related issue(s)

Related to #796

### Checklist

- [x] Manual testing passed.


stake-token-prices/README.md
```diff
@@ -0,0 +1,57 @@
...
+This architecture allows making very few calls to the price APIs, scales well with the user's base and even allows the prices to be fed even if the pricing APIs are temporarily down.
+
+### Configuration
+
+These environment variables control how the `cron` job behaves:
+
+| Variable                                     | Description                                                                         | Default               |
+| -------------------------------------------- | ----------------------------------------------------------------------------------- | --------------------- |
+| CACHE_EXPIRATION_SEC                         | How long the prices will be kept in the cache.                                      | 3600                  |
+| REFRESH_INTERVAL_SEC                         | How frequently the cache will be refreshed. If set to 0, it will run once and exit. | 300                   |
```
gndelia, line 22: `3600` and `300` - What units are these? I assume `ms`, but it also could be seconds 🤔

stake-token-prices/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Local development and testing
+
+To start all the services, run:
+
+```sh
+COIN_GECKO_API_KEY=MY_API_KEY docker compose up -d
+```
+
+Alternatively, if the environment variable are stored in a `.env.local` file, run:
```
gndelia, line 46:
```suggestion
Alternatively, if the environment variable is stored in a `.env.local` file, run:
```

stake-token-prices/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```sh
+docker compose --env-file=.env.local up
+```
+
+Then the prices can be obtained as follows:
+
+```console
+$ curl http://localhost:3000
+{"prices":{"bitcoin":"97324","lorenzo-stbtc":"97334","merlin-s-seal-btc":"94216","pumpbtc":"97175","wrapped-bitcoin":"96443"},"time":"2025-02-14T21:05:36.599Z"}
```
gndelia, line 56: It would be ideal if we could return this mapping to either the token addresses, or at least, the token symbols defined in the token list. This way, the mapping on the FE is simplified.

stake-token-prices/README.md
gndelia: so stupid, I just noticed the `_SEC` suffix 🤦🏽

.eslintrc.json
```diff
@@ -3,6 +3,7 @@
   "overrides": [
     {
       "excludedFiles": [
+        "stake-token-prices/**/*",
```
gndelia, line 6: _I think_ if you add the glob in a overrides, you don't need to add them here.

```suggestion
```


Note that the other overridden values are not listed in `excludedFiles`

stake-token-prices/README.md
gabmontes: I'll change it to the token symbol as the address will not match the one in Hemi but in their home network, i.e. Ethereum mainnet.

alvarengathomas (dismissed):

gndelia:
I think we should remove the `stake-` prefix, as we are now planning to use this to return prices for tunnel tokens as well.

token-prices/cron/package.json
```diff
@@ -0,0 +1,12 @@
...
+    "refresh-prices": "node src/refresh-prices.js"
+  },
+  "dependencies": {
+    "redis": "4.7.0",
+    "startinterval2": "1.0.1"
+  },
+  "engines": {
+    "node": ">=20"
+  }
+}
```
gndelia, line 12: same in this file

token-prices/api/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "dependencies": {
+    "promise-limit-one": "1.1.0",
+    "redis": "4.7.0"
+  },
+  "engines": {
+    "node": ">=20"
+  }
+}
```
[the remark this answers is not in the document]
gabmontes: No, it is not actually needed as this is not a package.

token-prices/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+These environment variables control how the `cron` job behaves:
+
+| Variable                | Description                                                                         | Default               |
+| ----------------------- | ----------------------------------------------------------------------------------- | --------------------- |
+| CACHE_EXPIRATION_SEC    | How long the prices will be kept in the cache.                                      | 3600                  |
+| REFRESH_INTERVAL_SEC    | How frequently the cache will be refreshed. If set to 0, it will run once and exit. | 300                   |
+| COIN_MARKET_CAP_API_KEY | The CoinMarketCap API key.                                                          |                       |
+| COIN_MARKET_CAP_SLUGS   | String of comma separated token slugs. I.e. "bitcoin,ethereum"                      | bitcoin               |
+| REDIS_URL               | The URL of the Redis database.                                                      | http://localhost:6379 |
```
gndelia, line 25: The [rule](https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md) mentions that we should use backticks to prevent it from being flagged. 

```suggestion
| REDIS_URL               | The URL of the Redis database.                                                      | `http://localhost:6379` |
```

The same below. Otherwise, we should remove this rule

package-lock.json
```diff
@@ -27193,6 +27354,36 @@
         "vitest": "2.0.5"
       }
     },
+    "stake-token-prices/api": {
```
gndelia, line 27357: I still see the old name being referenced here

gndelia (changes requested):

package-lock.json
gabmontes: Now de diffs are huge. Review that! 😆

gabmontes: Solved!

gndelia: > Now de diffs are huge. Review that! 😆

cmd + f and type `stake`, it was never so easy! 😆

gndelia (approved):

alvarengathomas (approved):

PR #898, opened 2025-02-21
Standardize decimals accross the portal

### LINKED ISSUE #860: Standardize decimal display for wallet balance across the portal
Currently, the wallet balance across the portal is displayed with 3 decimals. We need to update this to show 4 decimals and introduce a hover interaction to display the full balance when applicable.

Expected Behavior

1. Display 6 decimals instead of 3 across all wallet balance instances, but only when needed.
2. If the balance has more than 6 decimals, truncate with `...`
3. On hover, show the full balance in a tooltip for better accuracy.

Design [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=5376-11659&t=UOohedfG2DPMjwCq-1) and [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=9248-35588&t=UOohedfG2DPMjwCq-1)

Screenshots:

Image

Image

gndelia:
### Description

This PR attempts to standardize how we should the decimals across the whole app. For this

- Tooltips are shown for every amount. If it's a token, the logo is shown. The full amount is shown here. If it's not a token (for example, blockchain fees), no logos are shown
- The `...` are shown for small numbers less than the minimum that can be shown. In this case, we can show up to 6 decimals (depending on the precision of the larger number), so if the number is smaller than `0.000001`, the `...` become visible, so it is more obvious there's a tooltip
- We use the [smart-round](https://github.com/bloq/smart-round/) package everywhere to be consistent in how we get the formatted amount.

### Screenshots

https://github.com/user-attachments/assets/476e25cf-914c-418e-8556-275fb438c1e3

### Related issue(s)

Closes #860 

### Checklist

- [x] Manual testing passed.


webapp/components/displayAmount.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const showDots = bigAmount.lt(0.000001) && notZero
+  return (
+    <Tooltip
+      id={`amount-tooltip-${symbol}`}
+      overlay={
+        notZero ? (
+          <div className="flex items-center gap-x-1 px-2 py-1 text-sm font-medium text-white">
+            {'token' in props && <TokenLogo size="small" token={props.token} />}
+            <span>{`${new Intl.NumberFormat('en-US', {
+              maximumFractionDigits: 18,
```
gabmontes, line 48: We should use here the `digits` of the displayed token.

webapp/components/displayAmount.tsx
```diff
@@ -0,0 +1,66 @@
...
+  ...props
+}: Props) {
+  const formattedAmount = formatNumber(amount)
+
+  const symbol = 'symbol' in props ? props.symbol : props.token.symbol
+  const bigAmount = Big(amount)
+  const notZero = !bigAmount.eq(0)
+  // Only show dots for small numbers, less than the max 6 digits we're showing
+  // for formatted numbers.
+  const showDots = bigAmount.lt(0.000001) && notZero
```
gabmontes, line 39: This is the same as the old "<0.00001" label we said we do not want to use because some degens may not understand math. We either re-evaluate using that for small amounts or use the elipsis when numbers less than 1 are removed its less-significant (rightmost) digits as we discussed previously. cc/ @dvnahuel

gabmontes (changes requested):

webapp/components/displayAmount.tsx
gndelia: Done in [e902045](https://github.com/hemilabs/ui-monorepo/pull/898/commits/e902045a7be1d7c018b1d08589486f77d7d0e25b)

webapp/components/displayAmount.tsx
gndelia: As discussed with you and Nahuel, we will now only show decimals if the amount is less than 1, and we're truncating decimals.
Applied here [4765112](https://github.com/hemilabs/ui-monorepo/pull/898/commits/4765112f910695c0c69aaa17365deb05d973e741)

webapp/components/balance.tsx
```diff
@@ -29,8 +29,13 @@ export const RenderBalance = ({
     )}
     {(status === 'error' || (status === 'pending' && fetchStatus === 'idle')) &&
       '-'}
-    {status === 'success' &&
-      formatNumber(formatUnits(balance, token.decimals), 2)}
+    {status === 'success' && (
```
ArturDolzan, line 32: It's not related to this PR itself, but shouldn't we use enums for this kind of status? this would avoid typos, like comparing 'sucess' for example.

gabmontes (approved):

webapp/components/balance.tsx
ArturDolzan: I knew I should have checked on VS first haha. thanks!

dvnahuel (commented):
LGTM!

ArturDolzan (approved):

Issue #794, opened 2025-01-28
Define "Strategy Details" values in StakeDrawer


gndelia:
### Feature or change description

<img width="536" alt="Image" src="https://github.com/user-attachments/assets/c57002cf-37e5-412d-b5fa-96245f7c5ac0" />

We need to figure out how to retrieve these values

- TVL _may be_ `totalSupply()` of the token, TBC

0xJonHoldsCrypto:
Please refer to https://docs.google.com/document/d/1XBGllRK6Tt60ASk2kuWICgsBhjv1QOH9dyhxVhyMfRw/edit?tab=t.0#heading=h.g1xvvuiws3qt

PR #918, opened 2025-02-26
Add TVL in "Strategy Details"

gndelia:
### Description

This PR adds the TVL in the "Strategy Details" section. As discussed with Nahuel, we use the "compact" notation for TVL larger than 1 million. TVL is calculated as the price of the token times the supply of emitted tokens.
It also updates the `extensions` for some tokens to add the symbol they should consider for the pricing - for example, some of the BTC tokens use the BTC price.

If the token prices API is not enabled, a `-` is shown

### Screenshots

https://github.com/user-attachments/assets/dfdd2bd8-836f-46b9-a248-338d8924dccf

### Related issue(s)

Closes https://github.com/hemilabs/ui-monorepo/issues/794

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/components/fiatBalance.tsx
+++ b/webapp/components/fiatBalance.tsx
@@ -5,8 +5,8 @@ import { useTokenBalance, useNativeTokenBalance } from 'hooks/useBalance'
 import { useTokenPrices } from 'hooks/useTokenPrices'
 import { ComponentProps } from 'react'
 import Skeleton from 'react-loading-skeleton'
-import { smartRound } from 'smart-round'
 import { type BtcToken, type EvmToken, type Token } from 'types/token'
+import { formatFiatNumber } from 'utils/format'
 import { isNativeToken } from 'utils/nativeToken'
 import { isEvmToken } from 'utils/token'
 import { formatUnits } from 'viem'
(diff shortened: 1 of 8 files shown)


dvnahuel:
LGTM!

webapp/components/fiatBalance.tsx
```diff
@@ -37,7 +37,8 @@ const RenderFiatBalanceUnsafe = function ({
 
   const stringBalance = formatUnits(balance, token.decimals)
 
-  const price = data?.[token.symbol.toUpperCase()] ?? '0'
+  const price =
+    data?.[(token.extensions?.priceSymbol ?? token.symbol).toUpperCase()] ?? '0'
```
ArturDolzan, line 41: This line of code is a bit complex to read at first glance. I would suggest breaking it down into at least two steps, something like this below. Thoughts?
```suggestion
const priceKey = (token.extensions?.priceSymbol ?? token.symbol).toUpperCase()
const price = data?.[priceKey] ?? '0'
```

gndelia: Simplified here [276dc21](https://github.com/hemilabs/ui-monorepo/pull/918/commits/276dc21904586e1b381d3768ba351af747ba0e9f)

ArturDolzan (approved):

gabmontes (approved):

PR #927, opened 2025-02-27
Disable button when allowance is loading

### LINKED ISSUE #878: Submit button in the Tunnel should be disabled if allowance hasn't been loaded
For ERC20 tokens, allowance must be checked before tunneling and staking. Before that happens, the submit button can't be enabled, because the code doesn't know if it should request approval or not.

When the user selects a token, the tunnel form loads the allowance. However, while that happens, the submit button is enabled, allowing the user to submit. Generally, this is not a problem, because the allowance data loads fairly quickly. However, if the RPCs are slow, as it's been happening lately, the user could potentially submit the form without the allowance info being loaded, which could cause an error (as it will skip the approve step).

We should check the allowance and disable the submit button until this information is loaded, (perhaps showing a "..." indicating that the info is being loaded).

This is already fixed on the Stake page, but it needs to be applied on the Tunnel.

qgatssdev:
### Description
Submit button should be disabled while allowance is loading for ERC20 tokens.

Closes #878 
Fixes #
Related to #

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
+++ b/webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
@@ -6,15 +6,18 @@ import { CustomTunnelsThroughPartner } from 'components/customTunnelsThroughPart
 import { EvmFeesSummary } from 'components/evmFeesSummary'
 import { useNativeTokenBalance, useTokenBalance } from 'hooks/useBalance'
 import { useChain } from 'hooks/useChain'
+import { useHemi } from 'hooks/useHemi'
 import { useNetworkType } from 'hooks/useNetworkType'
 import dynamic from 'next/dynamic'
 import { useTranslations } from 'next-intl'
 import { useEffect, useState } from 'react'
+import { getTunnelContracts } from 'utils/crossChainMessenger'
 import { getNativeToken, isNativeToken } from 'utils/nativeToken'
 import { tunnelsThroughPartner } from 'utils/token'
 import { walletIsConnected } from 'utils/wallet'
 import { formatUnits } from 'viem'
 import { useAccount as useEvmAccount } from 'wagmi'
+import { useAllowance } from 'wagmi-erc20-hooks/useAllowance'
 
 import { useDeposit } from '../_hooks/useDeposit'
 import { EvmTunneling, TypedTunnelState } from '../_hooks/useTunnelState'
@@ -41,11 +44,13 @@ const SubmitEvmDeposit = function ({
   isRunningOperation,
   needsApproval,
   operationRunning,
+  isAllowanceLoading,
 }: {
   canDeposit: boolean
   isRunningOperation: boolean
   needsApproval: boolean
   operationRunning: OperationRunning
+  isAllowanceLoading: boolean
 }) {
   const t = useTranslations()
 
@@ -63,6 +68,9 @@ const SubmitEvmDeposit = function ({
     if (!isRunningOperation) {
       return texts[needsApproval ? 'approve' : 'deposit'].idle
     }
+    if (isAllowanceLoading) {
+      return '...'
+    }
     if (operationRunning === 'approving') {
       return texts.approve.loading
     }
@@ -73,7 +81,10 @@ const SubmitEvmDeposit = function ({
   }
 
   return (
-    <Button disabled={!canDeposit || isRunningOperation} type="submit">
+    <Button
+      disabled={!canDeposit || isRunningOperation || isAllowanceLoading}
+      type="submit"
+    >
       {getOperationButtonText()}
     </Button>
   )
@@ -91,6 +102,7 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
 
   const t = useTranslations()
   const { track } = useUmami()
+  const hemi = useHemi()
 
   const {
     fromInput,
@@ -101,7 +113,7 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
     updateFromInput,
   } = state
 
-  const { chain, status } = useEvmAccount()
+  const { chain, status, address } = useEvmAccount()
   const operatesNativeToken = isNativeToken(fromToken)
 
   const { balance: walletNativeTokenBalance } = useNativeTokenBalance(
@@ -112,6 +124,23 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
     fromToken,
     !operatesNativeToken,
   )
+
+  const l1StandardBridgeAddress = getTunnelContracts(
+    hemi,
+    fromToken.chainId,
+  ).L1StandardBridge
+
+  const { data: allowance, isPending } = useAllowance(
+    fromToken.address as `0x${string}`,
+    {
+      args: {
+        owner: address,
+        spender: l1StandardBridgeAddress,
+      },
+    },
+  )
+
+  const isAllowanceLoading = isPending || !allowance
 
   const canDeposit = canSubmit({
     balance: operatesNativeToken
@@ -254,6 +283,7 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
       return (
         <SubmitEvmDeposit
           canDeposit={canDeposit}
+          isAllowanceLoading={isAllowanceLoading}
           isRunningOperation={isRunningOperation}
           needsApproval={needsApproval}
           operationRunning={operationRunning}
--- a/webapp/app/[locale]/tunnel/_components/withdraw.tsx
+++ b/webapp/app/[locale]/tunnel/_components/withdraw.tsx
@@ -10,6 +10,7 @@ import { useNativeTokenBalance, useTokenBalance } from 'hooks/useBalance'
 import { useWithdrawBitcoin } from 'hooks/useBtcTunnel'
 import { useChain } from 'hooks/useChain'
 import { useEstimateFees } from 'hooks/useEstimateFees'
+import { useHemi } from 'hooks/useHemi'
 import { useNetworks } from 'hooks/useNetworks'
 import { useNetworkType } from 'hooks/useNetworkType'
 import dynamic from 'next/dynamic'
@@ -19,12 +20,14 @@ import Skeleton from 'react-loading-skeleton'
 import { type RemoteChain } from 'types/chain'
 import { Token } from 'types/token'
 import { isEvmNetwork } from 'utils/chain'
+import { getTunnelContracts } from 'utils/crossChainMessenger'
 import { formatBtcAddress } from 'utils/format'
 import { getNativeToken, isNativeToken } from 'utils/nativeToken'
 import { tunnelsThroughPartner } from 'utils/token'
 import { walletIsConnected } from 'utils/wallet'
 import { formatUnits, parseUnits } from 'viem'
 import { useAccount } from 'wagmi'
+import { useAllowance } from 'wagmi-erc20-hooks/useAllowance'
 
 import { useMinWithdrawalSats } from '../_hooks/useMinWithdrawalSats'
 import {
@@ -234,6 +237,7 @@ const EvmWithdraw = function ({ state }: EvmWithdrawProps) {
   const [networkType] = useNetworkType()
   const [isWithdrawing, setIsWithdrawing] = useState(false)
   const [isPartnersDrawerOpen, setIsPartnersDrawerOpen] = useState(false)
+  const hemi = useHemi()
 
   const t = useTranslations()
   const { track } = useUmami()
@@ -248,7 +252,7 @@ const EvmWithdraw = function ({ state }: EvmWithdrawProps) {
     updateFromInput,
   } = state
 
-  const { chainId, status } = useAccount()
+  const { chainId, status, address } = useAccount()
 
   const operatesNativeToken = isNativeToken(fromToken)
 
@@ -263,7 +267,25 @@ const EvmWithdraw = function ({ state }: EvmWithdrawProps) {
     !operatesNativeToken,
   )
 
+  const l1StandardBridgeAddress = getTunnelContracts(
+    hemi,
+    fromToken.chainId,
+  ).L1StandardBridge
+
+  const { data: allowance, isPending } = useAllowance(
+    fromToken.address as `0x${string}`,
+    {
+      args: {
+        owner: address,
+        spender: l1StandardBridgeAddress,
+      },
+    },
+  )
+
+  const isAllowanceLoading = isPending || !allowance
+
   const canWithdraw =
+    !isAllowanceLoading &&
     canSubmit({
       balance: operatesNativeToken
         ? walletNativeTokenBalance
@@ -272,7 +294,8 @@ const EvmWithdraw = function ({ state }: EvmWithdrawProps) {
       fromInput,
       fromNetworkId,
       fromToken,
-    }) && hasBridgeConfiguration(fromToken, toNetworkId)
+    }) &&
+    hasBridgeConfiguration(fromToken, toNetworkId)
 
   const {
     clearWithdrawState,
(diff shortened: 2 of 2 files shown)


qgatssdev:
Would have loved to attach a screenshot or video but currently having problems recreating the issue, the RPC's are loading fast enough. If there are any suggestions on how I can test, I would love to hear them.

webapp/app/[locale]/tunnel/_components/withdraw.tsx
```diff
@@ -362,12 +385,17 @@ const EvmWithdraw = function ({ state }: EvmWithdrawProps) {
...
     if (walletIsConnected(status)) {
       return (
-        <Button disabled={!canWithdraw || isWithdrawing} type="submit">
-          {t(
-            `tunnel-page.submit-button.${
-              isWithdrawing ? 'withdrawing' : 'initiate-withdrawal'
-            }`,
-          )}
+        <Button
+          disabled={!canWithdraw || isWithdrawing || isAllowanceLoading}
```
Comment (gabmontes) [inline webapp/app/[locale]/tunnel/_components/withdraw.tsx:389]: How is the allowance affecting the ability of the user to withdraw?

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
```diff
@@ -6,15 +6,18 @@ import { CustomTunnelsThroughPartner } from 'components/customTunnelsThroughPart
...
 import dynamic from 'next/dynamic'
 import { useTranslations } from 'next-intl'
 import { useEffect, useState } from 'react'
+import { getTunnelContracts } from 'utils/crossChainMessenger'
 import { getNativeToken, isNativeToken } from 'utils/nativeToken'
 import { tunnelsThroughPartner } from 'utils/token'
 import { walletIsConnected } from 'utils/wallet'
 import { formatUnits } from 'viem'
 import { useAccount as useEvmAccount } from 'wagmi'
+import { useAllowance } from 'wagmi-erc20-hooks/useAllowance'
```
Comment (gndelia) [inline webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:20]:
```suggestion
import { useAllowance } from 'wagmi-erc20-hooks'
```

This ☝🏽  seems to be causing the build to fail

Comment (gabmontes) [inline webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:20]: This is breaking the build.

gabmontes (changes requested):

webapp/app/[locale]/tunnel/_components/withdraw.tsx
```diff
@@ -263,7 +267,25 @@ const EvmWithdraw = function ({ state }: EvmWithdrawProps) {
...
+    {
+      args: {
+        owner: address,
+        spender: l1StandardBridgeAddress,
+      },
+    },
+  )
+
+  const isAllowanceLoading = isPending || !allowance
+
```
Comment (gndelia) [inline webapp/app/[locale]/tunnel/_components/withdraw.tsx:286]: Allowance should only be checked when going from L1 -> L2. Approvals are not needed the other way around. So all changes in `withdraw.tsx` should be rollbacked

gndelia (changes requested):

gndelia:
> Would have loved to attach a screenshot or video but currently having problems recreating the issue, the RPC's are loading fast enough. If there are any suggestions on how I can test, I would love to hear them.

The DevTools have tools to throttle requests, add lag, etc, perhaps something like that may help you. Otherwise, you can locally replace the call, using a timeout, and returning a hardcoded amount just to simulate it

webapp/app/[locale]/tunnel/_components/withdraw.tsx
Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/withdraw.tsx:389]: My bad, I misunderstood the issue. Resolved!

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:20]: Thank you, resolved.

Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:20]: Thank you! Fixed it

webapp/app/[locale]/tunnel/_components/withdraw.tsx
Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/withdraw.tsx:286]: I misunderstood the issue 😅, thank you! I've rolled it back.

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
```diff
@@ -113,6 +125,23 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
...
+    fromToken.address as `0x${string}`,
+    {
+      args: {
+        owner: address,
+        spender: l1StandardBridgeAddress,
+      },
+    },
+  )
+
+  const isAllowanceLoading = isPending || !allowance
```
Comment (gndelia) [inline webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:143]: `allowance` can be zero. I think here you should just rename `isPending` as `isAllowanceLoading`.

gndelia (changes requested):

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx]: Hmm interesting 🤔, Done!

Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx]: Hmm interesting 🤔, Done!

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
```diff
@@ -63,6 +68,9 @@ const SubmitEvmDeposit = function ({
     if (!isRunningOperation) {
       return texts[needsApproval ? 'approve' : 'deposit'].idle
     }
+    if (isAllowanceLoading) {
+      return '...'
```
Comment (gabmontes) [inline webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:72]: @dvnahuel Do we want to show `...` in the caption of the Deposit button while we load additional information to validate if the deposit conditions are met?

Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:72]: > @dvnahuel Do we want to show `...` in the caption of the Deposit button while we load additional information to validate if the deposit conditions are met?

@dvnahuel any update on this?

Comment (dvnahuel) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:72]: I suggest that the '...' be connected to the action taking place. For example, if the CTA says 'Connect to Hemi network,' the loading state should be something like 'Connecting to Hemi network...' instead of just using the three dots. That way, it's clearer and more informative.

Comment (gndelia) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:72]: @dvnahuel in this case the user performed no action - here information is being loaded from the wallet user to validate they can/cannot tunnel. In this particular case, information whether an approval is required or not is being loaded

Comment (dvnahuel) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:72]: Oh, gotcha! If that's the case, I would suggest using the loader component.

qgatssdev:
@dvnahuel @gndelia 
Screenshot 2025-03-06 at 12 55 00 PM
does this work?

gndelia:
> @dvnahuel @gndelia Screenshot 2025-03-06 at 12 55 00 PM does this work?

I think I prefer the whole button be replaced by the loading skeleton (with the same dimensions)

dvnahuel:
> @dvnahuel @gndelia Screenshot 2025-03-06 at 12 55 00 PM does this work?

I like this approach, but istead of showing the skeleton in this case, I would prefer to use the 'loading' component, something like this.

Screenshot 2025-03-07 at 2.03.07 PM
Screenshot 2025-03-07 at 2.01.48 PM

dvnahuel:
> @dvnahuel @gndelia I created a reusable spinner component, there was an existing instance of a spinner but it was tailored to that particular component so i removed that too. Screenshot 2025-03-10 at 9 32 06 AM Screenshot 2025-03-10 at 9 04 52 AM

This is great. Thanks!

webapp/components/Spinner.tsx
```diff
@@ -0,0 +1,100 @@
...
+            to={`360 ${center} ${center}`}
+            type="rotate"
+          />
+        </g>
+      </svg>
+    </div>
+  )
+}
+
+export default Spinner
```
gndelia, line 100: Can we use named exports here? The whole app uses named exports for all files (There may be an exception, but it is very rare)

In addition to that, the filename should be a lowercase `spinner.tsx`

gndelia (changes requested):

webapp/components/Spinner.tsx
```diff
@@ -0,0 +1,100 @@
+/* eslint-disable sort-keys */
```
gndelia, line 1: Let's not disable this.

```suggestion
```

If you disable it for the Props, let's sort them

gndelia (changes requested):

webapp/components/Spinner.tsx
gndelia: if you disable it for the `sizeMap`, I prefer to disable only for the lines that show an error, and not for the entire file. See how we do something similar in [tailwind config](https://github.com/hemilabs/ui-monorepo/blob/3e3f63e379f03a21c5f35ab0e62e60099b259e36/webapp/tailwind.config.ts#L86) (And in line [192](https://github.com/hemilabs/ui-monorepo/blob/3e3f63e379f03a21c5f35ab0e62e60099b259e36/webapp/tailwind.config.ts#L192))

webapp/components/Spinner.tsx
```diff
@@ -0,0 +1,100 @@
...
+  size = 'medium',
+  color = '#FFF7F0',
+  className = '',
+}: SpinnerProps) {
+  // IMPORTANT: We generate a unique ID for each spinner instance using React's useId() hook.
+  // This ensures that when multiple spinners are used on the same page, each has its own
+  // gradient definitions. Without unique IDs, all spinners would reference the same gradient,
+  // which could cause rendering issues if one spinner is removed from the DOM or when spinners
+  // have different colors. The unique ID prevents these conflicts by giving each spinner
+  // its own isolated gradient definitions in the SVG.
```
gndelia, line 20: very nice comment here 👏🏽

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
```diff
@@ -60,6 +66,9 @@ const SubmitEvmDeposit = function ({
         loading: t('tunnel-page.submit-button.depositing'),
       },
     }
+    if (isAllowanceLoading) {
+      return <Spinner size={'small'} />
+    }
```
Comment (gndelia) [inline webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:71]: I think, unfortunately 😓 , it is a bit more complex than this.

With the current implementation, this gets stuck on loading for ETH... which is not an ERC20 token, and therefore never needs to load anything

image

We may need to update `useAllowance` to consider this scenario (updating the `enabled` property inside `query`, as it must be trying to load for the native token. There's an utility `isNativeAddress` for that. I'd also recommend to move the `useAllowance`  out of `wagmi-hooks` (which is something I've been doing, but haven't quite finished), and use `useReadContract` from `wagmi` directly. Similar to what I did for `useTotalSupply` in `webapp/hooks/useTotalSupply.ts`

I'm recommending doing so, so you can use the `isNativeAddress` from `utils/nativeToken`, and with that, being able to differentiate that case in the Submit button

webapp/components/Spinner.tsx
gndelia: @qgatssdev  Don't worry! This is why revisions are for 😄 

We prefer named exports because this forces to use the same name everywhere. With default exports, the consumer of the package could do

```tsx
// they could do any of these options
// different name
import FakeSpiner from 'components/spinner'
// mispelling
import Spiner from 'components/spinner'
```

and nothing prevents devs from doing so. This makes text searches more complex, and the code more inconsistent.

If they still wish for forever reason rename the component on the consumer, they can still do so

```tsx
import { Spinner as MySpinner } from 'components/spinner'
```

So the option is still available to them if needed. IMHO the named export is a better default.

In addition to that, you can use default export only once, but there may be cases where you want to export multiple components from a file (We don't believe in a rigid one export per file hehe, although it is the most common scenario though), so I feel it would be very inconsistent to use default exports in some places, and named exports in others. Better to use named exports everywhere!

qgatssdev: I totally get it now! Thanks so much for sharing

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:72]: @gndelia Resolved, Let me know if I handled it properly. And you can assign me the task to remove the hooks out of `wagmi-hooks `. I can get to that in my spare time.

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
```diff
@@ -113,6 +127,24 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
...
+
+  const { isLoading: isAllowanceLoading } = useAllowance(
+    fromToken.address as `0x${string}`,
+    {
+      args: {
+        owner: address,
+        spender: l1StandardBridgeAddress,
+      },
+      query: {
+        enabled: !isNativeAddress(address),
```
Comment (gndelia) [inline webapp/app/[locale]/tunnel/_components/evmDeposit.tsx:143]:
```suggestion
        enabled: !isNativeAddress(fromToken.address),
```

`address` is the user's address from the Wallet

gndelia (changes requested):

webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
Comment (qgatssdev) [reply webapp/app/[locale]/tunnel/_components/evmDeposit.tsx]: Resolved

webapp/hooks/useAllowance.ts
```diff
@@ -1,24 +1,22 @@
...
-  query?: UseReadContractParameters<Erc20Abi, 'allowance'>['query']
+  query?: UseReadContractParameters<typeof erc20Abi, 'allowance'>['query']
 }
 
 export const useAllowance = (
   erc20Address: Address,
   { args: { owner, spender }, query }: Options,
 ) =>
-  useReadErc20({
+  useReadContract({
```
gndelia, line 19: I tested this and the allowance was never returned, thus the code was not working. 

After reviewing it, the issue is that, because the previous implementation used `useReadErc20`, the `abi` was automatically added to the call, which you're not doing in `useReadContract` here. See [here](https://github.com/hemilabs/ui-monorepo/blob/76dddde387188fcb220471ca94b2fe2b2b828821/wagmi-erc20-hooks/baseErc20.ts#L13).

You need to

```tsx
useReadContract({
    address: erc20Address,
    //  Note the abi added here
    abi: erc20Abi,
    args: useMemo(() => [owner, spender], [owner, spender]),
    functionName: 'allowance',
    query,
  })
```

Given we need to change this, I'd also add one extra change:

Given allowance is only for ERC20 tokens, I'd move the checks of `enabled` inside this implementation, as they are all the same for all usages. The function, then would look like this

```tsx
useReadContract({
    address: erc20Address,
    abi: erc20Abi,
    args: useMemo(() => [owner, spender], [owner, spender]),
    functionName: 'allowance',
    query: {
       enabled: !isNativeAddress(erc20Address) && !!owner && !!spender
    },
  })
```

This will simplify the code a bit for all the consumers of this hook.

gndelia (changes requested):

webapp/hooks/useAllowance.ts
qgatssdev: So sorry I couldn't test before pushing, I don't have the token. Thanks for the awesome review, I took it a step further to future proof it Incase we have any other needs for fields in the `query` object. Please test and let me know If I need to make any revision.

```suggestion
  useReadContract({
    abi: erc20Abi,
    address: erc20Address,
    args: useMemo(() => [owner, spender], [owner, spender]),
    functionName: 'allowance',
    query: {
      ...query,
      enabled:
        !isNativeAddress(erc20Address) &&
        !!owner &&
        !!spender &&
        query?.enabled !== false,
    },
  })
```

gndelia: > , I don't have the token.

Did you mean you don't have tokens?

You could get some Sepolia ETH [here](https://cloud.google.com/application/web3/faucet/ethereum/sepolia) and some erc20 tokens [here](https://gho.aave.com/faucet/)

From there, you can tunnel them to Hemi if you need L2 tokens for testing

qgatssdev: Thank you, I just got some. I'll be able to test before end of day.

qgatssdev: @gndelia Sorry I'm just getting back to you, had a hectic day. So i tested and it actually didn't work, are there any other factors I'm to consider? It works for non native tokens.

qgatssdev: @gndelia I misunderstood the criteria, it actually works perfectly 😅. Please test and review.

gndelia (approved):
LGTM

gabmontes (approved):

PR #936, opened 2025-03-01
Handle Error queries on subgraphs

### LINKED ISSUE #931: Handle Error queries on subgraphs
When querying subgraphs, if the query fails, a 200 ok is returned instead

Image

Image

qgatssdev:
### Description
This PR addresses an issue where the application fails to properly handle errors from The Graph API. Even though the API returns a 200 OK HTTP status code, it includes error information in the response body that wasn't being checked.

- Updated the `GraphResponse` type to include error fields
- Added a `checkGraphQLErrors` helper function to properly detect and handle GraphQL errors
- Added an `ensureJsonParsed` helper function to handle both string and JSON responses consistently
- Modified the subgraph query functions to check for errors before processing data
- Improved error messages to make debugging easier

### Related issue(s)

Closes #931 
Fixes #931 

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/webapp/utils/subgraph.ts
+++ b/webapp/utils/subgraph.ts
@@ -13,7 +13,9 @@ type Schema = {
   variables?: Record<string, string | number>
 }
 
-type GraphResponse<T> = { data: T } | { errors: { message: string }[] }
+type SuccessResponse<T> = { data: T }
+type ErrorResponse = { errors: { message: string }[] }
+type GraphResponse<T> = SuccessResponse<T> | ErrorResponse
 
 const getSubgraphUrl = function ({
   chainId,
@@ -86,7 +88,9 @@ const getTunnelSubgraphUrl = function (chainId: Chain['id']) {
  * @param response The GraphQL response to check
  * @throws Error if the response contains errors
  */
-function checkGraphQLErrors<T>(response: GraphResponse<T>) {
+function checkGraphQLErrors<T>(
+  response: GraphResponse<T>,
+): asserts response is SuccessResponse<T> {
   // Check if response has errors
   if ('errors' in response && response.errors.length > 0) {
     // Extract error messages and join them
@@ -111,16 +115,12 @@ export const getLastIndexedBlock = function (chainId: Chain['id']) {
       }
     }`,
   }
-  return fetch(url, {
-    body: JSON.stringify(schema),
-    headers: {
-      'Content-Type': 'application/json',
-    },
-    method: 'POST',
-  }).then(
-    (response: GraphResponse<{ _meta: { block: { number: number } } }>) => (
-      checkGraphQLErrors(response),
-      'data' in response && response.data._meta.block.number
+  return request<GraphResponse<{ _meta: { block: { number: number } } }>>(
+    url,
+    schema,
+  ).then(
+    response => (
+      checkGraphQLErrors(response), response.data._meta.block.number
     ),
   )
 }
@@ -193,21 +193,20 @@ export const getBtcWithdrawals = function ({
   return request<GetBtcWithdrawalsQueryResponse>(url, schema).then(
     response => (
       checkGraphQLErrors(response),
-      'data' in response &&
-        response.data.btcWithdrawals.map(d => ({
-          // The Subgraph lowercases all the addresses when saving, so better convert them
-          // into checksum format to avoid errors when trying to get balances or other operations.
-          // GraphQL also converts BigInt as strings
-          ...d,
-          blockNumber: Number(d.blockNumber),
-          // @ts-expect-error OP-SDK does not properly type addresses as Address
-          from: toChecksum(d.from),
-          // @ts-expect-error OP-SDK does not properly type addresses as Address
-          l1Token: toChecksum(d.l1Token),
-          // @ts-expect-error OP-SDK does not properly type addresses as Address
-          l2Token: toChecksum(d.l2Token),
-          timestamp: Number(d.timestamp),
-        }))
+      response.data.btcWithdrawals.map(d => ({
+        // The Subgraph lowercases all the addresses when saving, so better convert them
+        // into checksum format to avoid errors when trying to get balances or other operations.
+        // GraphQL also converts BigInt as strings
+        ...d,
+        blockNumber: Number(d.blockNumber),
+        // @ts-expect-error OP-SDK does not properly type addresses as Address
+        from: toChecksum(d.from),
+        // @ts-expect-error OP-SDK does not properly type addresses as Address
+        l1Token: toChecksum(d.l1Token),
+        // @ts-expect-error OP-SDK does not properly type addresses as Address
+        l2Token: toChecksum(d.l2Token),
+        timestamp: Number(d.timestamp),
+      }))
     ),
   ) satisfies Promise<ToBtcWithdrawOperation[]>
 }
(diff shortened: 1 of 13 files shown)


qgatssdev:
@gndelia Please how can this be tested? I resolved this issue based on the screenshot you shared, I would love to know how to recreate and test.

webapp/utils/subgraph.ts
```diff
@@ -20,7 +20,7 @@ const subgraphIds = {
   [sepolia.id]: process.env.NEXT_PUBLIC_SUBGRAPH_SEPOLIA_ID,
 }
 
-type GraphResponse<T> = { data: T }
+type GraphResponse<T> = { data: T; errors?: Array<{ message: string }> }
```
gndelia, line 23: I think this should be more like this

```suggestion
type GraphResponse<T> = { data: T } | { errors: { message: string }[] }
```

webapp/utils/subgraph.ts
```diff
@@ -61,10 +92,12 @@ export const getLastIndexedBlock = function (chainId: Chain['id']) {
...
     method: 'POST',
-  }).then(
-    (r: GraphResponse<{ _meta: { block: { number: number } } }>) =>
-      r.data._meta.block.number,
-  )
+  })
+    .then(response => ensureJsonParsed(response))
+    .then(
+      (response: GraphResponse<{ _meta: { block: { number: number } } }>) =>
+        checkGraphQLErrors(response)._meta.block.number,
```
gndelia, line 99: this function is confusing from the outside. If the function checks for errors, then it should not return anything there are no errors. Perhaps it should be something like 

```suggestion
        checkGraphQLErrors(response)
        return response.data._meta.block.number,
```

webapp/utils/subgraph.ts
```diff
@@ -131,22 +164,24 @@ export const getEvmDeposits = function ({
...
-      l2Token: toChecksum(d.l2Token),
-      timestamp: Number(d.timestamp),
-      // @ts-expect-error OP-SDK does not properly type addresses as Address
-      to: toChecksum(d.to),
-    })),
-  ) satisfies Promise<EvmDepositOperation[]>
+  })
+    .then(response => ensureJsonParsed(response))
+    .then((response: GetEvmDepositsQueryResponse) =>
+      checkGraphQLErrors(response).deposits.map(d => ({
```
gndelia, line 170: Same here

gndelia:
> @gndelia Please how can this be tested? I resolved this issue based on the screenshot you shared, I would love to know how to recreate and test.

Thanks for your contribution!

These scenarios are always hard to test. You can always replace the `fetch` call with a setTimeout + returning a hardcoded response. If you want to get more sophisticated, you can use libraries such as [msw](https://mswjs.io/) to mock responses at the browser level (If you want to do that, that's ok, but don't try to send it as part of your PR)

webapp/utils/subgraph.ts
```diff
@@ -39,6 +39,37 @@ const getGraphUrl = function (chainId: Chain['id']) {
...
+ * Helper function to ensure the response is properly parsed as JSON
+ * @param response The response that might be a string or already parsed
+ * @returns The parsed JSON response
+ */
+function ensureJsonParsed<T>(response: string | T): T {
+  if (typeof response === 'string') {
+    return JSON.parse(response)
+  }
+  return response as T
+}
```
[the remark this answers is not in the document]
qgatssdev: Okay, noted

gabmontes (dismissed):

gndelia:
Conflicts!

qgatssdev:
> Conflicts!
@gndelia I've resolved the conflicts locally and trying to push. Currently getting this lint error while trying to push 
Screenshot 2025-03-04 at 9 26 33 PM

qgatssdev:
> > Conflicts!
> 
> @gndelia I've resolved the conflicts locally and trying to push. Currently getting this lint error while trying to push Screenshot 2025-03-04 at 9 26 33 PM

@gndelia

qgatssdev:
@gndelia Fixed! Sorry about the messy history, I ran into a problem 😅

webapp/utils/subgraph.ts
```diff
@@ -104,8 +118,10 @@ export const getLastIndexedBlock = function (chainId: Chain['id']) {
     },
     method: 'POST',
   }).then(
-    (r: GraphResponse<{ _meta: { block: { number: number } } }>) =>
-      r.data._meta.block.number,
+    (response: GraphResponse<{ _meta: { block: { number: number } } }>) => (
+      checkGraphQLErrors(response),
+      'data' in response && response.data._meta.block.number
+    ),
```
gndelia, line 124: This is a bit hard to read. Ok, perhaps this one exactly not, but the other `.then` for `GetBtcWithdrawalsQueryResponse`  becomes very large. I think here, it could be better to add a regular function.

In addition to that, there's a bit of a redundant check for `'data' in response`. We know that if `checkGraphQLErrors` did not throw, then `response` must be a successful type. I think we can improve that like this:

Update `GraphResponse<T>` to something like this:

```ts
type SuccessResponse<T> = { data: T }
type ErrorResponse = { errors: { message: string }[] }
type GraphResponse<T> = SuccessResponse<T> | ErrorResponse
```

Then, you can update `checkGraphQLErrors` to something like this 

```ts
/**
 * Helper function to check for errors in GraphQL responses
 * @param response The GraphQL response to check
 * @throws Error if the response contains errors
 */
function checkGraphQLErrors<T>(
  response: GraphResponse<T>,
): asserts response is SuccessResponse<T> {
  // Check if response has errors
  if ('errors' in response && response.errors.length > 0) {
    // Extract error messages and join them
    const errorMessages = response.errors.map(e => e.message).join(', ')
    throw new Error(`GraphQL Error: ${errorMessages}`)
  }
}
```

You can check the [docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) for that `asserts response is SuccessResponse<T>`. It works very similar to a user-defined Type Guard.

If we do that, then we can finally replace the function that I am commenting with

```suggestion
    .then(
    (response: GraphResponse<{ _meta: { block: { number: number } } }>) => {
      checkGraphQLErrors(response)
      // At this point, the compiler knows that `response` is of type `SuccessResponse` and that `response.data` is valid. No need to check again!
      return response.data._meta.block.number
    },
  )
```

Let's do that, and do something similar for the other returned functions here.

We're getting closer to an optimal solution!

gndelia (changes requested):

webapp/utils/subgraph.ts
qgatssdev: @gndelia Thanks for the review! I learnt something new, I've made the changes.

webapp/utils/subgraph.ts
```diff
@@ -97,15 +115,13 @@ export const getLastIndexedBlock = function (chainId: Chain['id']) {
...
-  }).then(
-    (r: GraphResponse<{ _meta: { block: { number: number } } }>) =>
-      r.data._meta.block.number,
+  return request<GraphResponse<{ _meta: { block: { number: number } } }>>(
+    url,
+    schema,
+  ).then(
+    response => (
+      checkGraphQLErrors(response), response.data._meta.block.number
+    ),
```
gndelia, line 124: As suggested in the previous comments, I think this format makes it harder to read. Let's use classic functions with `return`

```suggestion
  ).then(
    function (response) {
      checkGraphQLErrors(response)
      return response.data._meta.block.number
    },
```

Apply this to all the changes, other than that, LGTM

gndelia (changes requested):

webapp/utils/subgraph.ts
qgatssdev: @gndelia all done! Sorry i misunderstood the expected change request. Fixed now

gndelia (approved):

webapp/utils/subgraph.ts
gndelia: Looks great! Thanks!

ArturDolzan (approved):
LGTM

gabmontes (approved):
Good work @qgatssdev and @gndelia!

Issue #895, opened 2025-02-21
Prevent users to tunnel less than the minimum amount allowed by the token decimals


gndelia:
Let's say a token, like USDC, has 6 decimals in the contract definition. That means the smaller amount of USDC possible is `0.000001`. Currently, there's no restriction on the minimum amount to tunnel. So users could try to tunnel `0.00000000000001 USDC`, which the UI parses as `0`.  
The contracts do allow trying to tunnel 0, see [this tunnel deposit](https://sepolia.etherscan.io/tx/0x0655fca75cc7c46a8fcc9dbf28f01dfa49773b8877f56d3edc1393352d53bd0c).
However, allowing this kind of transaction is useless, as no tokens are minted on Hemi, and they are only a waste of fees. We should prevent that

We should restrict the amount to tunnel to/from Hemi, preventing it from operating less than the minimum amount. This needs to be validated in the tunnel form as well.

qgatssdev:
@gndelia How would you like this carried out? If the user tries to pass 0, do we disable the button or prompt a toast. What's the intended solution? 
Also what should be the minimum amount that can be tunnel to/from Hemi?

gndelia:
@qgatssdev The minimum is dictated by the amount of decimals defined in the Token contract.

`USDC` has defined decimals as 6. The smallest amount possible then is `0.000001`. It's not possible to tunnel less than that, because the value can't be represented. Less than that is zero, which should be prevented. If the amount written in the input by the user is invalid, the submit button should be disabled.

TBD by @dvnahuel how we would show to users that the amount is invalid

qgatssdev:
> [@qgatssdev](https://github.com/qgatssdev) The minimum is dictated by the amount of decimals defined in the Token contract.
> 
> `USDC` has defined decimals as 6. Then , the smallest amount possible then is `0.000001`. It's impossible to tunnel less than that because smaller values can't be represented. Less than that is zero, which should be prevented. If the amount written in the input by the user is invalid, the submit button should be disabled, similar to how we disable the submit button when the user tries to tunnel more than the balance amount.
> 
> TBD by [@dvnahuel](https://github.com/dvnahuel) how we would show to users that the amount is invalid

Okay I get what you mean, please assign this to me. I'll love feedback from @dvnahuel so I know how we want to present the error.

qgatssdev:
@gndelia Any update on this? I have a branch fixing this but currently don't know how the data is to be presented.

[...]

PR #937, opened 2025-03-01
Remove "networkType" query string on default network

### LINKED ISSUE #920: Remove "networkType" query string when it matches the default value
We use `?networkType=testnet` or `?networkType=mainnet` in the query string to indicate which is the chain the user is seeing.

The current default value is `testnet`; it eventually will be `mainnet`. 

Whenever we're on the chain that matches the default value, links are still adding the `?networkType=<value>`, which causes that as soon as the user navigates to another page (for example, from the tunnel to the TX history), the URL gets that query string added. If sharing the URL, this query string value becomes part of it, when it wasn't actually needed. 

It would be ideal to prevent adding the query string parameter to links (and to the URL) if the user is in the default chain. This should keep the URL cleaner whenever possible

qgatssdev:
### Description
Improve URL handling for network type query parameter by only adding it when not on the default network. 

This change prevents unnecessary query parameters from cluttering URLs when navigating between pages on the default network (testnet or mainnet). The `Link` component now conditionally adds the `networkType` parameter, keeping URLs clean while preserving network type selection when explicitly changed.

### Screenshots
https://www.loom.com/share/23cdda7652654eb99f9f0488be4414cc?sid=1e744592-66b1-4efb-a7cc-295679738a9f

### Related issue(s)

Closes #920
Fixes #920 

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/webapp/components/link.tsx
+++ b/webapp/components/link.tsx
@@ -1,5 +1,4 @@
-import { featureFlags } from 'app/featureFlags'
-import { useNetworkType } from 'hooks/useNetworkType'
+import { defaultNetworkType, useNetworkType } from 'hooks/useNetworkType'
 import BaseLink from 'next-intl/link'
 import { ComponentProps } from 'react'
 import { type UrlObject } from 'url'
@@ -29,7 +28,6 @@ const getQuery = function (q: UrlObject['query']): Record<string, string> {
 // between mainnet and testnet when explicitly selected.
 export const Link = function (props: ComponentProps<typeof BaseLink>) {
   const [networkType] = useNetworkType()
-  const defaultNetworkType = featureFlags.mainnetEnabled ? 'mainnet' : 'testnet'
 
   let href = props.href
 
--- a/webapp/components/link.tsx
+++ b/webapp/components/link.tsx
@@ -1,3 +1,4 @@
+import { featureFlags } from 'app/featureFlags'
 import { useNetworkType } from 'hooks/useNetworkType'
 import BaseLink from 'next-intl/link'
 import { ComponentProps } from 'react'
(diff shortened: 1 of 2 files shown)


webapp/components/link.tsx
```diff
@@ -22,14 +23,31 @@ const getQuery = function (q: UrlObject['query']): Record<string, string> {
...
   const [networkType] = useNetworkType()
-  const href =
-    typeof props.href === 'string'
-      ? { pathname: props.href, query: { networkType } }
-      : { ...props.href, query: { ...getQuery(props.href.query), networkType } }
+  const defaultNetworkType = featureFlags.mainnetEnabled ? 'mainnet' : 'testnet'
+
+  let href = props.href
+
+  if (networkType !== defaultNetworkType) {
```
qgatssdev, line 36: I opted for If statements over ternary operators for readability sake, there are more than 3 conditionals going on and It would better for future maintainers to understand.

webapp/components/link.tsx
```diff
@@ -22,14 +23,31 @@ const getQuery = function (q: UrlObject['query']): Record<string, string> {
...
+// query parameter when it differs from the default network type. This keeps
+// the URL clean for the default network while still allowing navigation
+// between mainnet and testnet when explicitly selected.
 export const Link = function (props: ComponentProps<typeof BaseLink>) {
   const [networkType] = useNetworkType()
-  const href =
-    typeof props.href === 'string'
-      ? { pathname: props.href, query: { networkType } }
-      : { ...props.href, query: { ...getQuery(props.href.query), networkType } }
+  const defaultNetworkType = featureFlags.mainnetEnabled ? 'mainnet' : 'testnet'
```
gndelia, line 32: This is also defined in [useNetworkType](https://github.com/hemilabs/ui-monorepo/blob/852e37a6910fb9d16915e371843b8bdabec13703/webapp/hooks/useNetworkType.ts#L14).

In that file, I'd move that ternary outside of the hook (but inside the same file), and export it as `defaultNetwork`. Then, I'd import it here. 
This way, when we remove the featureFlag, we won't have to rewrite this again to import the default network

gndelia (commented):
I left a minor improvement, but overall LGTM. Let's change that and then we can merge it.

Thanks for your contribution!

gndelia (approved):

qgatssdev:
> I left a minor improvement, but overall LGTM. Let's change that and then we can merge it.
> 
> Thanks for your contribution!

At your service 🫡 ,  any update on this issue? #895 I'll like to resolve that next.

PR #944, opened 2025-03-03
Build token prices service Docker images

gabmontes:
### Description

Build and push the Docker images for the token prices API on every push to the default branch.

### Related issue(s)

Related to #796


gndelia (approved):
uACK

PR #945, opened 2025-03-03
Token price docker images

gabmontes:
### Description

After analyzing the [build errors](https://github.com/hemilabs/ui-monorepo/actions/runs/13641220100/job/38131332391) in #944, it looks like the solution is to let the Docker action find the Dockerfile in the default path ({context}/Dockerfile). Prefixing it with "./" seems to cause problems. Let's see.

### Related issue(s)

Related to #796


gndelia (approved):
uack

Issue #953, opened 2025-03-04
Add ETH TVL in "Strategy Details"


gndelia:
For most tokens, we've calculated TVL in the "Strategy Details" section as the total supply of the token. However, for ETH, as it is not a token, we will use WETH total supply instead, as defined by @0xJonHoldsCrypto

PR #942, opened 2025-03-03
Enable Staking with Native Token (ETH)

ArturDolzan:
### Description

This PR enables staking with the native token (ETH). When a user stakes ETH, the smart contract automatically deposits it as WETH (Wrapped ETH). Similarly, upon unstaking, the contract returns WETH instead of native ETH.

### Screenshots

Captura de Tela 2025-03-03 às 17 48 21

Captura de Tela 2025-03-03 às 17 48 29

### Related issue(s)

Related to #921 

### Checklist

- [x] Manual testing passed.
- [ ] Automated tests added, or N/A.
- [ ] Documentation updated, or N/A.
- [ ] Environment variables set in CI, or N/A.

PR #954, opened 2025-03-04
Add ETH TVL

gndelia:
### Description

FOLLOW UP OF #952

This PR does 2 things:

- After merging #942, it was no longer valid that the `address` of a `StakeToken` was a `0x${string}` type (`Address`) because we enabled Native token staking (which has no real Address). So for that, I updated the type in 6a3543dec937f8be3269dc74e4c3432e139657ae. This caused different places to need to be fixed. In some cases, just a suppression error (from using `string` where Address was expected) was enough. In others, I had to consider if the token was a Native one and handle that accordingly.
- 5a01ed091ee922e7b371c10c0b80c8c4beaa9efd Implements the issue to be closed, in which we show the WETH supply when staking ETH

### Screenshots

image

### Related issue(s)

Closes https://github.com/hemilabs/ui-monorepo/issues/953

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/stake/_components/manageStake/stakeOperation.tsx
+++ b/webapp/app/[locale]/stake/_components/manageStake/stakeOperation.tsx
@@ -55,11 +55,14 @@ export const StakeOperation = function ({
 }: Props) {
   const { address } = useAccount()
   const [amount, setAmount] = useAmount()
+  const operatesNativeToken = isNativeToken(token)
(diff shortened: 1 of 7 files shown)


gabmontes (approved):

alvarengathomas (dismissed):

gndelia:
I had to rebase 😬  it requires re-review!

PR #955, opened 2025-03-04
Added 404 and 500 error pages

### LINKED ISSUE #620: Design and Implement 500 and 404 error pages

As part of adding Error tracking (See #553), we need a 500 error page (on this page, unhandled errors will be captured). Let's take an advantage and add the 404 as well

[404](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=6648-14273&t=EiZJkLZ2Nfv7neBP-1)

[500](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=6648-13971&t=EiZJkLZ2Nfv7neBP-1)

alvarengathomas:
### Description

Added error pages with some caveats:
- The 500 error page cannot have a menu on the side as the figma designs required because it is a global error page and don't have the necessary react context to show the side menu.
- There is also no translation because the error pages doesn't have the intl react context.
- I could not use the `<Button>/<ButtonLink>` components because it depends on the router/navigation and the error pages don't have access to that as well.

### Screenshots

Error 500 (Mobile):
Captura de Tela 2025-03-04 às 15 57 11

Error 500 (Desktop):
Captura de Tela 2025-03-04 às 16 00 28

Error 404 (Mobile):
Captura de Tela 2025-03-04 às 16 10 01

Error 404 (Desktop):
Captura de Tela 2025-03-04 às 16 10 29

### Related issue(s)

Closes #620

### Checklist

- [x] Manual testing passed.


dvnahuel:
LGTM!

webapp/app/global-error.tsx
```diff
@@ -1,8 +1,15 @@
 'use client'
 
 import * as Sentry from '@sentry/nextjs'
+import { ExclamationMark } from 'components/icons/exclamationMark'
+import Image from 'next/image'
 import { useEffect } from 'react'
 
+import Error500 from './_images/500.svg'
```
gabmontes, line 8: Check and solve these warnings: 👇🏼

webapp/app/not-found.tsx
```diff
@@ -1,3 +1,33 @@
...
+      <Image
+        alt="404"
+        className="absolute inset-0 -top-64 m-auto w-72 md:w-fit"
+        src={Error404}
+      />
+      <div className="z-10 m-auto flex flex-col items-center gap-4">
+        <ExclamationMark />
+        <div className="text-center">
+          <h1 className="text-lg font-medium text-neutral-950">
+            Oops! Page not found
```
gabmontes, line 21: Can we apply translations to these error messages?

gndelia: For the 404, hitting this page may mean that the user requested the page `https://app.hemi.xyz/sarasa`

As locales are part of the URL, what locale would you apply? I think for this particular case, English is just fine.

Now, this is a global `not-found`. I think you can add this file on different route segments. So if a `not-found.tsx` was created below `app/[locale]/` folder, then we would get access to translations. Note that the global not-found will still be used when requesting `https://app.hemi.xyz/sarasa`, but if requesting `https://app.hemi.xyz/en/sarasa`, then we would get the 404 page with the locale defined.

alvarengathomas: @gabmontes I explained about the translations in the PR description.

@gndelia Do you want me to add a `not-found.tsx` file on every route segments?

gndelia: > Do you want me to add a not-found.tsx file on every route segments?

I think the global one (without translation) + one under `/[locale]`  (with translations) should work well for all URLs.

webapp/app/global-error.tsx
alvarengathomas: fixed

webapp/app/not-found.tsx
gabmontes: ☝🏼 that's what should be done here.

alvarengathomas: done

webapp/app/[locale]/not-found.tsx
```diff
@@ -0,0 +1,36 @@
+'use client'
+
+import 'styles/globals.css'
+import { ExclamationMark } from 'components/icons/exclamationMark'
+import Image from 'next/image'
+
+import Error404 from '../_images/404.svg'
```
Comment (gabmontes) [inline webapp/app/[locale]/not-found.tsx:7]: More linter warnings to solve here.

ArturDolzan (approved):
LGTM

gndelia (approved):

PR #962, opened 2025-03-05
Check Docker images for vulnerabilities and improve image generation

gabmontes:
### Description

The Docker images are now generated with proper provenance and SBoM in addition to both "ref" and short commit SHA tags.

Vulnerabilities are checked too, using `aquasecurity/trivy-action`, which is the same tool DockerHub uses internally. Some NPM cache files were skipped to speed up the scanning process by not checking unneeded files.

One vulnerability related to `cross-spawn` was detected in the NPM version that comes globally installed with the Node base image and it was resolved by updating it to a newer version.

### DIFF
--- a/.github/actions/docker-build-push/action.yml
+++ b/.github/actions/docker-build-push/action.yml
@@ -25,6 +25,11 @@ runs:
       uses: docker/metadata-action@v5
       with:
         images: ${{ inputs.images }}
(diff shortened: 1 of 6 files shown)


gndelia (approved):
uACK

alvarengathomas (approved):

Issue #949, opened 2025-03-04
Add Disclaimer when Staking ETH


gndelia:
After enabling ETH/WETH in #942 , we need a Disclaimer for Stake, as users will Stake ETH but only be able to unstake WETH

dvnahuel:
This is my [proposal](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=7941-78126&t=dTGQl326drrfWfQD-1)

Context: This will appear when users click 'Unstake' on their ETH staking position in the dashboard.

Screenshot: 

<img width="348" alt="Image" src="https://github.com/user-attachments/assets/45131fe0-7d4f-4b71-a5dc-89355f646de3" />

@rmilrad can you help us with the text? I tried to make one similar to Zircuit.

<img width="377" alt="Image" src="https://github.com/user-attachments/assets/1af7fdc4-9752-4bd9-8fb6-8126e437d6c6" />

PR #968, opened 2025-03-05
Add disclaimer when unstaking eth

gndelia:
### Description

This PR adds a disclaimer when unstaking WETH.

### Screenshots

image

### Related issue(s)

Related #949

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/stake/_components/manageStake/preview.tsx
+++ b/webapp/app/[locale]/stake/_components/manageStake/preview.tsx
@@ -5,6 +5,7 @@ import { useTranslations } from 'next-intl'
 import { ComponentProps, ReactNode } from 'react'
 import { StakeOperations, StakeToken } from 'types/stake'
 
+import { DisclaimerEth } from './disclaimerEth'
 import { Tabs } from './tabs'
 
 type Props = {
@@ -37,6 +38,8 @@ export const Preview = function ({
 }: Props) {
   const t = useTranslations('common')
 
+  const showEthDisclaimer = ['ETH', 'WETH'].includes(token.symbol)
+
   return (
     <>
       {showTabs && (
@@ -62,6 +65,11 @@ export const Preview = function ({
         </div>
       </DrawerSection>
       <div className="px-4 md:px-6">{strategyDetails}</div>
+      {operation === 'unstake' && showEthDisclaimer && (
+        <div className="px-4 md:px-6">
+          <DisclaimerEth />
+        </div>
+      )}
       <div className="mt-auto flex flex-col gap-y-3 text-center">
         {submitButton}
       </div>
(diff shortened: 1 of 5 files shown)


webapp/messages/en.json
```diff
@@ -224,8 +224,11 @@
         "tvl": "TVL (on Hemi)",
         "website": "Website"
       },
+      "unstake-eth-as-weth-heading": "Unstaked ETH will be received as WETH",
+      "unstake-eth-as-weth-subheading": "When you stake ETH, Hemi automatically wraps it as WETH. Upon unstaking, you’ll receive WETH, which you can unwrap back to ETH if desired.",
```
gabmontes, line 228: Hemi does not do that. Can we just say something like...

```suggestion
      "unstake-eth-as-weth-subheading": "When you stake ETH, it is automatically converted to WETH. Upon unstaking, you’ll receive WETH, which you can unwrap back to ETH if desired.",
```

gndelia: Updated in [b5ba553](https://github.com/hemilabs/ui-monorepo/pull/968/commits/b5ba553cb45e4d4ff53761c57cc63825c11dd572) - See if it now looks better

webapp/app/[locale]/stake/_components/manageStake/preview.tsx
```diff
@@ -62,6 +65,11 @@ export const Preview = function ({
         </div>
       </DrawerSection>
       <div className="px-4 md:px-6">{strategyDetails}</div>
+      {operation === 'unstake' && showEthDisclaimer && (
```
Comment (ArturDolzan) [inline webapp/app/[locale]/stake/_components/manageStake/preview.tsx:68]: Shouldn't we also show when `staking`? I mean, we stake eth but in the dashboard it already shows as wrapped eth...

Comment (gndelia) [reply webapp/app/[locale]/stake/_components/manageStake/preview.tsx:68]: We thought of it, we decided with Nahuel not to do so because when the user stakes, the strategy details section is visible, and there is not enough space. But perhaps we need to rethink this decision.

Comment (ArturDolzan) [reply webapp/app/[locale]/stake/_components/manageStake/preview.tsx:68]: Ah ok, well, IMHO, we should do.. for non-experienced users, it gets a bit confusing when going to the dashboard.

Captura de Tela 2025-03-05 às 18 59 24

Comment (gndelia) [reply webapp/app/[locale]/stake/_components/manageStake/preview.tsx:68]: Ok, I gave some thought to this. In order to show the disclaimer in "Stake", I need to perform other changes, particularly, fixing the CTA at the bottom regardless of the scroll. This is a change that needs to be implemented in all drawers, and not this particular one (For example, it's something that should be done on the Tunnel as well). It's been a long time since Nahuel requested this change and I haven't had the time to implement it.

@ArturDolzan I think we should go like this. Let's merge this, and I already updated the PR description so #949 _is not closed_ after merging this PR, as we still need to implement it in the "Stake" drawer. Let's consider this a partial progress, that is fully functional on its own, but not enough to close the PR.

Do you agree?

PR #995, opened 2025-03-10
Add Mainnet modal + badges

### LINKED ISSUE #991: Implement 'Mainnet now live!' modal.
The default chain before March 5th was Testnet. After that, Mainnet became the default, so users entering after we released Mainnet would automatically be on it. That’s why adding this modal will help notify users of the network change.

The primary CTA will simply close the modal and allow the user to continue navigating the portal, while the secondary CTA will take them to the Get Started page.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=10958-18081&t=MinS86MvY42Sp743-1)

Screenshot

Image

cc: @gndelia @rmilrad

gndelia:
### Description

This PR adds the "Mainnet is now live" modal, and the "Testnet" and "Mainnet" badges.

The modal is shown only once, and not again once closed. For that, there's a boolean flag saved in local storage.

The badges are shown in the nav bar (desktop) and in the header (mobile)

### Screenshots

image

image

image

image

image

image

### Related issue(s)

Closes #991

### Checklist

- [x] Manual testing passed.


webapp/components/mainnetLiveModal/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+    useHideMainnetLiveModal()
+  const { width } = useWindowSize()
+
+  if (hideMainnetLiveModal) {
+    return null
+  }
+
+  const onClose = () => setHideMainnetLiveModal(true)
+
+  return width >= 768 ? (
```
ArturDolzan, line 82: Will not this code force the component to be rendered a lot of times? shouldn't we use any debounce strategy?

gndelia: Not really. `useWindowSize` uses [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore) so this is stable, that is not causing re-renderings when not needed. When resizing, this will cause re-renders, but besides being uncommon, it's not something that affects performance as tested

ArturDolzan: Thanks for clarifying.

ArturDolzan (approved):

PR #1005, opened 2025-03-10
Add and whitelist uniBTC to stake on mainnet

### LINKED ISSUE #997: Add uniBTC to staking
Need to add uniBTC to staking page. 

Website: https://app.bedrock.technology/unibtc?tab=swap_deposit

contract address: 0xF9775085d726E782E83585033B58606f7731AB18

ArturDolzan:
### Description

As the title says, this PR adds and whitelists uniBTC token to be staked on mainnet.

### Screenshots

Captura de Tela 2025-03-10 às 15 32 41

### Related issue(s)

Closes #997 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/patches/@hemilabs+token-list+1.12.0.patch
+++ b/patches/@hemilabs+token-list+1.12.0.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@hemilabs/token-list/src/hemi.tokenlist.json b/node_modules/@hemilabs/token-list/src/hemi.tokenlist.json
-index ba2991f..d354e9b 100644
+index e112ffd..11e9457 100644
 --- a/node_modules/@hemilabs/token-list/src/hemi.tokenlist.json
 +++ b/node_modules/@hemilabs/token-list/src/hemi.tokenlist.json
 @@ -173,8 +173,7 @@
@@ -86,7 +86,7 @@ index ba2991f..d354e9b 100644
        },
        "logoURI": "https://raw.githubusercontent.com/hemilabs/token-list/master/src/logos/usdt.svg",
        "name": "USDT",
-@@ -400,8 +409,7 @@
+@@ -402,8 +411,7 @@
        "chainId": 43111,
        "decimals": 8,
        "extensions": {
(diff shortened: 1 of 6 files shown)


patches/@hemilabs+token-list+1.12.0.patch
```diff
@@ -1,5 +1,9 @@
+diff --git a/node_modules/@hemilabs/token-list/.DS_Store b/node_modules/@hemilabs/token-list/.DS_Store
+new file mode 100644
+index 0000000..5172429
+Binary files /dev/null and b/node_modules/@hemilabs/token-list/.DS_Store differ
```
gndelia, line 4: It seems that when you patched the file, a `.DS_Store` file was created, and included as part of the patch 🤔  Could you try again?

gabmontes (changes requested):

patches/@hemilabs+token-list+1.12.0.patch
ArturDolzan: Fixed!

webapp/tokenList/stakeTokens.ts
```diff
@@ -139,6 +140,13 @@ export const stakeWhiteList: Partial<
       rewards: ['hemi', 'unirouter', 'bsquared'],
       website: websitesMap.uniRouter,
     },
+    // uniBTC
+    '0xF9775085d726E782E83585033B58606f7731AB18': {
+      protocol: 'bedRock',
+      priceSymbol: 'btc',
+      rewards: ['hemi', 'bedrock'],
+      website: websitesMap.uniBtc,
```
gndelia, line 148: You've created the protocol `uniBTC`. I do understand that this is because you need a website that is different from bedRock. 

So we can either:

- Use the protocol `uniBTC` here as well, instead of Bedrock (I think I prefer this one)
- Drop the uniBTC protocol altogether, and inline the website URL here.

Use the option you prefer

ArturDolzan: Yes I've created because we had to set a different website. I know that uniBTC is not a protocol though. Not sure why websites were related to protocols instead tokens.

gndelia: I guess you're correct. I think initially I assumed some websites would be reused (I think that is the case for some) but not to others. Feel free to improve it (either in this PR or in a new one)

ArturDolzan: Set uniBTC as protocol and regarding the improvement, let's work in a different issue/PR.

gndelia (approved):

gabmontes (approved):

PR #1038, opened 2025-03-14
Fix mainnet modal

### LINKED ISSUE #1035: Minor improvements in the "Mainnet is now live!" modal
- [ ] On mobile, the modal is actually a drawer, and it has a horizontal scrollbar

Image

I tested it on mobile, and I also see it.

- [ ] If the user is on Testnet, the modal appears (ok), but upon clicking "Start using Mainnet"... the user is still in Testnet. Nothing happens - the user remains on Testnet. It should switch automatically to Mainnet.

qgatssdev:
### Description
Fix oveflow issues in mainnet modal and switch network  if on testnet

### Screenshots

IMG_6369

Closes #1035
Fixes #1035 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/components/mainnetLiveModal/index.tsx
+++ b/webapp/components/mainnetLiveModal/index.tsx
@@ -40,8 +40,10 @@ const Content = function ({ img: Img }: { img: ComponentType }) {
   const [, setNetworkType] = useNetworkType()
 
   return (
-    <div className="flex h-[70dvh] flex-col items-center gap-y-6 overflow-x-hidden bg-white px-4 md:h-[470px] md:w-[530px]">
-      <Img />
+    <div className="flex h-[70dvh] flex-col items-center gap-y-6 bg-white px-4 md:h-[470px] md:w-[530px]">
+      <div className="w-full max-md:max-w-[360px] max-md:overflow-hidden">
+        <Img />
+      </div>
       <div className="mt-4 flex flex-col gap-y-3">
         <h2 className="md:text-3.25xl text-center text-2xl font-semibold text-neutral-950">
           {t('mainnet-is-now-live')}
@@ -50,7 +52,7 @@ const Content = function ({ img: Img }: { img: ComponentType }) {
           {t('mainnet-is-now-live-description')}
         </p>
       </div>
-      <div className="flex items-center justify-center gap-x-3 [&>*]:w-36">
+      <div className="flex items-center justify-center gap-3 max-[368px]:flex-col [&>*]:w-36">
         <ButtonLink
           href="/get-started"
           onClick={() => setHideMainnetLiveModal(true)}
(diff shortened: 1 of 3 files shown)


webapp/components/mainnetLiveModal/index.tsx
```diff
@@ -57,7 +60,10 @@ const Content = function ({ img: Img }: { img: ComponentType }) {
         </ButtonLink>
         <Button
           height="h-5"
-          onClick={() => setHideMainnetLiveModal(true)}
+          onClick={function () {
+            setHideMainnetLiveModal(true)
+            setNetworkType('mainnet')
```
gabmontes, line 65: Forcing the network here makes sense IMO. @gndelia do you think this is ok or may be redundant/not needed?

webapp/components/mainnetLiveModal/index.tsx
```diff
@@ -36,8 +37,10 @@ const DrawerOverlay = () => (
 const Content = function ({ img: Img }: { img: ComponentType }) {
   const t = useTranslations('common')
   const [, setHideMainnetLiveModal] = useHideMainnetLiveModal()
+  const [, setNetworkType] = useNetworkType()
+
   return (
-    <div className="flex h-[70dvh] flex-col items-center gap-y-6 bg-white px-4 md:h-[470px] md:w-[530px]">
+    <div className="flex h-[70dvh] flex-col items-center gap-y-6 overflow-x-hidden bg-white px-4 md:h-[470px] md:w-[530px]">
```
gndelia, line 43: This is just trimming the image, I think we can do better. I will discuss with @dvnahuel  on Monday what's the best option for mobile (I guess resizing should work ok, but there may be some white spaces left for specific device sizes)

webapp/components/mainnetLiveModal/index.tsx
gndelia: I think this is correct. The modal appears in the app, regardless of which chain the user is. If the user was in Testnet, and we did not do this, the user would click "Start using mainnet" and would continue on Testnet...

webapp/components/mainnetLiveModal/index.tsx
qgatssdev: Okay, please tag me when there's a conclusion so I can push a fix.
Totally unrelated from this PR, I was using the https://hemi.xyz/ site and also found overflow on the mobile page. I'll attach a screenshot below, we should really fix that.

IMG_6374

gndelia: @qgatssdev I discussed with Nahuel and yes, we agree, given the image is an SVG, I think the best should be to resize the image to the container, so the modal doesn't overflow.

qgatssdev: @gndelia Okay, I'll push the fix now

gabmontes (approved):

gndelia (approved):

PR #1041, opened 2025-03-17
Removed all BTC tunnel and mainnet feature flags

### LINKED ISSUE #972: Remove feature toggles for Mainnet and Bitcoin
**Note**

*Wait until next week (After 3/12) to implement this just in case we need for whatever reason disable temporarily mainnet and or/bitcoin*

We need to remove all the code related to the 2 feature flags:

- [ ] enabling Bitcoin (`featureFlags.btcTunnelEnabled`, or `NEXT_PUBLIC_FEATURE_FLAG_ENABLE_BTC_TUNNEL` env variable)
- [ ] Enabling mainnet (`featureFlags.mainnetEnabled` it `NEXT_PUBLIC_FEATURE_FLAG_ENABLE_MAINNET` env variable)

In both cases, the code should work as if these values were `true` for ever. Remember to remove env variables from workflows and coordinate with Gabriel to remove them from the environment secrets/vars

alvarengathomas:
### Description

Removed all environment variables and feature flags checks in the code related to:
- Bitcoin Tunnel (`featureFlags.btcTunnelEnabled`, `NEXT_PUBLIC_FEATURE_FLAG_ENABLE_BTC_TUNNEL`)
- Mainnet (`featureFlags.mainnetEnabled`, `NEXT_PUBLIC_FEATURE_FLAG_ENABLE_MAINNET`)

### Related issue(s)

Closes #972

### Checklist

- [X] Manual testing passed.

### DIFF
--- a/webapp/hooks/useSyncHistory/index.ts
+++ b/webapp/hooks/useSyncHistory/index.ts
@@ -1,5 +1,4 @@
 import { featureFlags } from 'app/featureFlags'
-import { useBitcoin } from 'hooks/useBitcoin'
 import { useConnectedToSupportedEvmChain } from 'hooks/useConnectedToSupportedChain'
 import { useNetworks } from 'hooks/useNetworks'
 import { useNetworkType } from 'hooks/useNetworkType'
@@ -131,7 +130,6 @@ const debouncedSaveToStorage = debounce(
 
 export const useSyncHistory = function (l2ChainId: Chain['id']) {
   const { address } = useAccount()
-  const bitcoin = useBitcoin()
   const { remoteNetworks } = useNetworks()
   const [networkType] = useNetworkType()
 
@@ -298,9 +296,7 @@ export const useSyncHistory = function (l2ChainId: Chain['id']) {
         dispatch({ payload: deposit, type: 'add-deposit' }),
       addWithdrawalToTunnelHistory: (withdrawal: WithdrawTunnelOperation) =>
         dispatch({ payload: withdrawal, type: 'add-withdraw' }),
-      deposits: history.deposits
-        .filter(d => d.chainId !== bitcoin.id)
-        .flatMap(d => d.content),
+      deposits: history.deposits.flatMap(d => d.content),
       resyncHistory: () => setForceResync(true),
       syncStatus: history.status,
       updateDeposit: (
@@ -319,11 +315,9 @@ export const useSyncHistory = function (l2ChainId: Chain['id']) {
           payload: { updates, withdraw },
           type: 'update-withdraw',
         }),
-      withdrawals: history.withdrawals
-        .filter(w => w.chainId !== bitcoin.id)
-        .flatMap(w => w.content),
+      withdrawals: history.withdrawals.flatMap(w => w.content),
     }),
-    [bitcoin.id, dispatch, history],
+    [dispatch, history],
   )
 
   return [...reducer, historyContext] as const
--- a/webapp/tokenList/index.ts
+++ b/webapp/tokenList/index.ts
@@ -55,10 +55,9 @@ export const getRemoteTokens = function (token: EvmToken) {
   })) satisfies EvmToken[]
 }
 
-const hemiTokens: Token[] = (hemilabsTokenList.tokens as EvmToken[])
-  .filter(t => t.chainId === hemiMainnet.id || t.chainId === hemiTestnet.id)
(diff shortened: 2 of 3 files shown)


gabmontes (dismissed):

webapp/hooks/useSyncHistory/index.ts
```diff
@@ -299,7 +299,7 @@ export const useSyncHistory = function (l2ChainId: Chain['id']) {
       addWithdrawalToTunnelHistory: (withdrawal: WithdrawTunnelOperation) =>
         dispatch({ payload: withdrawal, type: 'add-withdraw' }),
       deposits: history.deposits
-        .filter(d => featureFlags.btcTunnelEnabled || d.chainId !== bitcoin.id)
+        .filter(d => d.chainId !== bitcoin.id)
```
gndelia, line 302: I think actually here, the correct thing should be removing the filter altogether

```suggestion
        .filter(d => d.chainId !== bitcoin.id)
```

if `featureFlags.btcTunnelEnabled` was enabled, the filter would always be true => the same as the filter not being there. If the feature was disabled, then we would filter out `bitcoin` deposits so they were not visible to users.

webapp/hooks/useSyncHistory/index.ts
```diff
@@ -320,7 +320,7 @@ export const useSyncHistory = function (l2ChainId: Chain['id']) {
           type: 'update-withdraw',
         }),
       withdrawals: history.withdrawals
-        .filter(w => featureFlags.btcTunnelEnabled || w.chainId !== bitcoin.id)
+        .filter(w => w.chainId !== bitcoin.id)
```
gndelia, line 323: same here

gndelia (changes requested):

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "scripts": {
     "format:check": "prettier --check .",
+    "format:write": "prettier --write .",
```
gndelia, line 6: good addition 👍🏽 , I've typed this manually on the terminal too many times 😅

webapp/tokenList/index.ts
```diff
@@ -59,11 +58,7 @@ export const getRemoteTokens = function (token: EvmToken) {
 const hemiTokens: Token[] = (hemilabsTokenList.tokens as EvmToken[])
   .filter(t => t.chainId === hemiMainnet.id || t.chainId === hemiTestnet.id)
   // TODO the following line once bitcoin is enabled https://github.com/hemilabs/ui-monorepo/issues/738
```
gndelia, line 60: it seems this todo is no longer valid
```suggestion
```

webapp/tokenList/index.ts
```diff
@@ -59,11 +58,7 @@ export const getRemoteTokens = function (token: EvmToken) {
...
-  .filter(
-    t =>
-      (t.symbol !== 'hemiBTC' && t.symbol !== 'tBTC') ||
-      featureFlags.btcTunnelEnabled,
-  )
+  .filter(t => t.symbol !== 'hemiBTC' && t.symbol !== 'tBTC')
```
gndelia, line 61: This should also be removed. 
```suggestion
```

If `featureFlags.btcTunnelEnabled` was true, the `.filter()` callback would always return true, and would only be considered if the feature was disabled. But as now it will be always enabled, the filter needs to be removed.

gndelia (changes requested):

webapp/tokenList/index.ts
gndelia: This ☝🏽  causes the app to crash as soon as the user selects Bitcoin, so let's make sure to run a smoke test at least


https://github.com/user-attachments/assets/591082bf-10fa-4f14-ad18-ad84500a75bd

webapp/hooks/useSyncHistory/index.ts
alvarengathomas: removed

webapp/hooks/useSyncHistory/index.ts
alvarengathomas: removed

webapp/tokenList/index.ts
alvarengathomas: removed

webapp/tokenList/index.ts
alvarengathomas: removed

gabmontes (approved):

gndelia (approved):

Issue #913, opened 2025-02-25
Add USD prices accross the app


gndelia:
We're currently showing the USD prices in 2 places:

- Stake (both dashboard and table)
- The token Selector

There are a few more places where we could show the prices in USD

- [x] The connection wallet drawer
- [ ] The transaction history (perhaps difficult to get the price point in the past)
- [x] In the tunnel form input
- [x] In the stake form input
- [x] The review operation drawer?

This issue is to tackle the design and implementation of this

gndelia:
The only remaining item to implement is to add USD prices to the TX history. However, this would imply getting the prices in the time the transaction was executed. We don't have such a capability, and after discussing with @gabmontes  it is not worth the effort. So I'm closing this issue as completed

PR #1048, opened 2025-03-19
Added USD price to wallet and review drawers

alvarengathomas:
### Description

Added USD prices texts as following:

✅  1. Connect Wallet drawer [Approved by @dvnahuel]:

Captura de Tela 2025-03-19 às 13 22 09

---

✅ 2. Review Operation drawer [Approved by @dvnahuel]:

Captura de Tela 2025-03-19 às 13 48 45

---

❌ 3. Tunnel Form input [Rejected by @dvnahuel as it will change for v3]:

Captura de Tela 2025-03-19 às 13 39 21

---

❌ 4. Stake Form input [Rejected by @dvnahuel as it will change for v3]:

Captura de Tela 2025-03-19 às 15 59 18

---

❌ 5. The transaction history (Could not get the USD price point in the past)

---

### Related issue(s)

Related to #913

### Checklist

- [X] Manual testing passed.

### DIFF
--- a/webapp/components/reviewOperation/amount.tsx
+++ b/webapp/components/reviewOperation/amount.tsx
@@ -1,6 +1,7 @@
 'use client'
 
 import { DisplayAmount } from 'components/displayAmount'
+import { FiatBalance } from 'components/fiatBalance'
 import { useTranslations } from 'next-intl'
 import { Token } from 'types/token'
 import { formatUnits } from 'viem'
@@ -15,11 +16,17 @@ export const Amount = function ({ token, value }: Props) {
   return (
     <div className="flex items-center justify-between text-sm font-medium">
       <span className="text-neutral-500">{t('total-amount')}</span>
-      <div className="text-neutral-950">
-        <DisplayAmount
-          amount={formatUnits(BigInt(value), token?.decimals ?? 18)}
-          token={token}
-        />
+      <div className="flex gap-x-1">
+        <div className="text-neutral-950">
+          <DisplayAmount
+            amount={formatUnits(BigInt(value), token?.decimals ?? 18)}
+            token={token}
+          />
+        </div>
+        <div className="flex items-center gap-x-1 font-normal text-neutral-500">
+          <span>($</span>
+          <FiatBalance token={token} />)
+        </div>
       </div>
     </div>
   )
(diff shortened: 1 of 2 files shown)


webapp/components/reviewOperation/amount.tsx
```diff
@@ -15,11 +16,17 @@ export const Amount = function ({ token, value }: Props) {
...
+      <div className="flex gap-x-1">
+        <div className="text-neutral-950">
+          <DisplayAmount
+            amount={formatUnits(BigInt(value), token?.decimals ?? 18)}
+            token={token}
+          />
+        </div>
+        <div className="flex items-center gap-x-1 font-normal text-neutral-500">
+          <span>($</span>
+          <FiatBalance token={token} />)
```
ArturDolzan, line 28: Shouldn't encapsulate this logic? Wouldn’t something like this be better?
```suggestion
          <span>(${<FiatBalance token={token} />})</span>
```

ArturDolzan (commented):
A minor comment otherwise LGTM.

ArturDolzan (approved):

webapp/components/reviewOperation/amount.tsx
```diff
@@ -15,11 +16,16 @@ export const Amount = function ({ token, value }: Props) {
...
+      <div className="flex gap-x-1">
+        <div className="text-neutral-950">
+          <DisplayAmount
+            amount={formatUnits(BigInt(value), token?.decimals ?? 18)}
+            token={token}
+          />
+        </div>
+        <div className="font-normal text-neutral-500">
+          <span>(${<FiatBalance token={token} />})</span>
+        </div>
```
gndelia, line 28: Adding this here, will add the USD value to all instances of the "Review operation drawer". This means that seeing a completed operation will show the USD price, for example, a past deposit. This seems to be the same issue as the Transaction History. So maybe we can't show it here either.

Thoughts @gabmontes  ?

dvnahuel: Hey Gonza, good catch. I think it would be clearer if we remove that USD value.

dvnahuel: For this instance I would suggest now showing prices at all.
cc: @alvarengathomas

alvarengathomas: removed price from the Review Operation drawer

PR #1049, opened 2025-03-20
Improve sorting on stake tables

### LINKED ISSUE #981: Improve default sorting on Dashboard and Stake tables
We should sort both tables like this

- First, at the top, hemiBTC, USDT, USDC, regardless of wallet amount or staked position
- Then, list all the rest of the assets the user has staked (in the dashboard), or has wallet balance (in the Stake table) sorted by USD value, in descending order
- The rest of the assets in no particular order

qgatssdev:
### Description
This PR adds token sorting functionality to the stake and dashboard tables, following these rules:

1. hemiBTC, USDT, USDC tokens are displayed at the top in that specific order
2. Tokens with wallet balance are displayed next, sorted by descending balance value
3. Remaining tokens are displayed in alphabetical order

Added a `sortTokens` utility function that implements the requested sorting rules
Ensured proper handling of token variants with '.e' suffix

### Screenshots
Screenshot 2025-03-20 at 2 43 27 PM

### Related issue(s)

Closes #981 
Fixes #981 

### Checklist

- [x] Manual testing passed.


webapp/utils/sortTokens.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (aHasBalance && bHasBalance) {
+      const aValue = getComparableBalance(a)
+      const bValue = getComparableBalance(b)
+
+      return bValue - aValue
+    }
+
+    return a.symbol.localeCompare(b.symbol)
+  })
+}
```
gndelia, line 71: Can we add a bunch of tests for this function? It will be easier to understand by reading how it works, and prevent failures if changes are added later. Check the `test` folder

qgatssdev: Okay I'll get to it today

gndelia:
@qgatssdev conflicts!

webapp/test/utils/sortTokens.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  decimals,
+  extensions: {
+    protocol: 'hemi',
+    rewards: [],
+    website: 'https://example.com',
+  },
+  logoURI: '',
+  name: symbol,
+  symbol,
+})
```
gndelia, line 25:
```suggestion
const createMockToken = function (params: {
  symbol: string,
  balance?: bigint,
  address,
  chainId
  decimals
): StakeToken {
  const { address = zeroAddress, balance,  chainId = 43111,   decimals = 18, symbol } = params ?? {}
  return {
    address,
    balance,
    chainId,
    decimals,
    extensions: {
      protocol: 'hemi',
      rewards: [],
      website: 'https://example.com',
    },
    logoURI: '',
    name: symbol,
    symbol,
  }
}
```


with this you can just call

```ts
createMockToken({ symbol: 'DAI', balance: BigInt(1000000) })
// this is also valid, it's easier to infer what each parameter does , and no longer depends on the order of the parameters
createMockToken({ balance: BigInt(1000000), symbol: 'DAI',  })
```


Note that I imported `zeroAddress` from `viem`

qgatssdev: @gndelia Noted, Resolved!

webapp/utils/sortTokens.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    if (aHasBalance !== bHasBalance) {
+      return aHasBalance ? -1 : 1
+    }
+
+    if (aHasBalance && bHasBalance) {
+      const aValue = getComparableBalance(a)
+      const bValue = getComparableBalance(b)
+
+      return bValue - aValue
```
gndelia, line 66: just noticed something. Here we're comparing balances of different tokens. This can't be done, they are not comparable.

We need to get the USD value for each and then compare those USD balances. We can't compare native balances of different tokens. Doing so will make it look as if `1 DAI` is more than `0.1 BTC`, and that's not the case.

I'm sorry I did not notice that before! (Although it is mentioned in the issue 😅 )


There's a component ([fiatBalance](https://github.com/hemilabs/ui-monorepo/blob/3e6b065e1164e694945458df3aeda43a4b8b6032/webapp/components/fiatBalance.tsx#L20)) that has some logic to convert a token's balance into a USD value.

Perhaps we need to extract some of it (Although we don't need the whole formatting part)

gndelia (changes requested):

webapp/utils/sortTokens.ts
qgatssdev: @gndelia Done! Sorry it took a while.

webapp/app/[locale]/stake/page.tsx
```diff
@@ -31,7 +33,10 @@ export default function Page() {
   // Removing WETH from the list of tokens to stake
   // here instead of in the tokenList file
   // It has to be rendered on dashboard page though
-  const stakeTokens = useStakeTokens().filter(t => t.symbol !== 'WETH')
+  const unsortedStakeTokens = useStakeTokens().filter(t => t.symbol !== 'WETH')
+
+  const { data: prices } = useTokenPrices()
+  const stakeTokens = sortTokens(unsortedStakeTokens, prices)
```
Comment (gndelia) [inline webapp/app/[locale]/stake/page.tsx:39]: Here, we're missing to consider prices against the amount of tokens the user is holding onto its wallet.

See the 2nd criterion on the issue

> Then, list all the rest of the assets the user has staked (in the dashboard), or has wallet balance (in the Stake table) sorted by USD value in descending order

Here, each wallet balance is missing.

Another thing to consider (both for this page and the dashboard) is that we should wait for all the data to be loaded before applying the sorting (or show a loading indicator or something per cell). Otherwise, users will see their rows reordered after the initial load.

gndelia (changes requested):

webapp/app/[locale]/stake/page.tsx
Comment (qgatssdev) [reply webapp/app/[locale]/stake/page.tsx:39]: Hey @gndelia, Just getting to this I've been under the weather. I understand what you mean, I'll raise a PR addressing this soon.

gabmontes:
Closing in favor of #1168.

Issue #1050, opened 2025-03-20
Update the 'Stake and earn points' to 'Earn points on Hemi' sticky card


dvnahuel:
**NOTE** The New card is implemented, this is Blocked until Nahuel creates the Spanish version of  the card

We need to update the designs on the sticky card from "Stake and earn points" to "Earn points on Hemi".

**New link**: https://boost.absinthe.network/hemi-mainnet/dashboard
Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=11201-11785&t=ZVZGKiwoqPlq5D1t-1)
Sreenshot:
![Image](https://github.com/user-attachments/assets/d6ea37d1-1128-4e67-a657-f7fc27acac54)

gndelia:
We need the Spanish version of the card from @dvnahuel

PR #1057, opened 2025-03-21
Replace 'Stake and Earn' card

gndelia:
### Description

This PR updates the "Stake and Earn" card with a new "Earn Hemi points" card. We now point to an Absinthe link

As this is a new campaign, the `key` we're using to store in local storage is updated (as well as the analytics event) so all users, even those who already hide the old card, see  the new one.

As this is now linking to an external page, the card is also shown in the `/stake/*` pages

Note that we are not showing the Spanish cards (we show the english ones instead), so that's why I'm leaving the issue open. That'd normally be a problem... but it turns out we weren't showing the current card in Spanish either 😓 

As soon as the new card is created by @dvnahuel I will add it in another PR.

I also replaced next's `Image` component with using SVG directly, as we were using the `png` files but these were being resized and causing some blurring on iPhone devices, due to Next's implementation. By using the SVG directly, this issues no longer appear.

### Screenshots

https://github.com/user-attachments/assets/b07ba5d8-3120-4619-bc57-607293b3e8ab

### Related issue(s)

Related to #1050

### Checklist

- [x] Manual testing passed.


ArturDolzan (approved):
LGTM

Issue #1071, opened 2025-03-25
Avoid enormous network payloads


alvarengathomas:
### Description

Avoid enormous network payloads to increase the performace score on page load.

### Screenshots

![Image](https://github.com/user-attachments/assets/f1d12b4c-2101-4511-b2a1-0e96a283fd32)

PR #1084, opened 2025-03-27
Replace large stake icons

gndelia:
### Description

I did a bit of analysis with [@next/bundle-analyzer](https://nextjs.org/docs/14/app/building-your-application/optimizing/bundle-analyzer) and found that a few icons in `Stake` were huge in size, although they were very small visible:

Particularly, 2 SVG point rewards were larger than the others, so I re-exported them from figma as PNG which takes less space

- 1025ac95bd10ee23b69207da8ce7774676cc1c54 replaces the `Babypie Jewels` icon. The whole token rewards icons are in the order of 500 kb

image

and this `babypieJewelsIcon.tsx` was responsible for like 60% of that. So I replaced it with a large PNG image.

| Before (SVG)  | After (PNG) |
| image | image |

(You need to focus on the small icon in the Rewards column)

- Then, in 0ad1e917114e12526130d2d55197a40bc7c51341 I did the same for the PumpBtc icon. Here, the difference wasn't that important in size, but I still did it as it was the 2nd largest icon.

| Before (SVG)  | After (PNG) |
| image | image |

The gains on `build` are on the order of ~300/400 KB

| Before  | After |
| image | image |

The original total size of the client code reported by the analyzer was `2.54` MB, and now it reports  `2.21` MB (gzipped both)

Looking forward to keep reducing this!

### Screenshots

No visible changes to users.

### Related issue(s)

Related to #1071

### Checklist

- [x] Manual testing passed.


ArturDolzan (approved):
Nice!

Issue #1061, opened 2025-03-24
Move subgraph calls into its own backend


gndelia:
We have several calls to Subgraphs. These are done directly from the FE to the backend. We should move these calls into a simple backend, letting the portal call this frontend.

Note that the backend should allow calls for both Testnet and Mainnet, and allow to be expanded for future subgraphs

The current existing subgraphs (6) are

- `hemi-stake-operations-subgraph` (testnet and mainnet)
- `hemi-tunnel-withdrawals-subgraph` (testnet and mainnet)
- `tunnel-deposits-subgraph` (Sepolia and Ethereum)

gabmontes:
Blocked until https://github.com/hemilabs/infrastructure/issues/474 is resolved.

PR #1097, opened 2025-03-28
Add first version of the subgraph API service

gabmontes:
### Description

This is the first step towards resolving #1061.


subgraph-api/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "scripts": {
+    "start": "tsx server.ts"
+  },
+  "dependencies": {
+    "@tsconfig/node20": "20.1.5",
+    "config": "3.3.12",
+    "cors": "2.8.5",
+    "dotenv": "16.4.7",
+    "express": "4.21.2",
```
gndelia, line 10: Just asking: I know express v5 has been out there in the wild for more than half a year. Any particular reason you wanted to stick with v4?

gndelia: Particularly, considering you instaled  `@types/express` - when installing `@types/*` packages, the major versions must match. So you either need `@types/express@4.x`, or bump express here

subgraph-api/src/api-server.ts
```diff
@@ -0,0 +1,178 @@
+import { hemi, hemiSepolia } from 'hemi-viem'
+import { mainnet, sepolia } from 'viem/chains'
+import config from 'config'
+import cors from 'cors'
+import express from 'express'
```
gndelia, line 5: On react, ESLint would throw a warning because these imports are not properly sorted 🤔  It may be a mismatch of rules.

Could it also be that `bloq/node` has more rules related to CommonJs than ESM?

subgraph-api/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "compilerOptions": {
+    "allowImportingTsExtensions": true,
+    "noEmit": true,
+    "resolveJsonModule": true,
+    "strict": false
```
gndelia, line 6: 😭

subgraph-api/src/api-server.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  if (skip) {
+    if (!isInteger(skip)) {
+      sendJsonResponse(res, 400, { error: 'skip must be a number' })
+      return
+    }
+    req.data.skip = Number.parseInt(skip)
+  }
+
+  next()
```
gndelia, line 72: we could check that all these numbers must be also positive

subgraph-api/src/api-server.ts
```diff
@@ -0,0 +1,178 @@
...
+        getWithdrawals = getBtcWithdrawals
+      }
+
+      // @ts-expect-error: req.data is populated by parseQueryParams
+      getWithdrawals({ address, ...req.data })
+        .then(function (withdrawals) {
+          sendJsonResponse(res, 200, { withdrawals })
+        })
+        // eslint-disable-next-line promise/no-callback-in-promise
+        .catch(next)
```
gndelia, line 143: Considering this is `express` and a common pattern when using this library, perhaps we could disable this rule for the entire `subgraph-api`. What do you think?

token-prices/api/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "token-prices-api",
```
gndelia, line 2: What changed? https://github.com/hemilabs/ui-monorepo/pull/857#discussion_r1963971742

subgraph-api/src/api-server.ts
gndelia: The rule that fails in the FE is `import/order`, it is configured for [bloq/next here](https://github.com/bloq/eslint-config-bloq/blob/c4f7b586abc08593dc93d8f75df840f2424fc66f/next.js#L14)

subgraph-api/src/subgraph.ts
```diff
@@ -0,0 +1,401 @@
+// Copied from webapp/utils/subgraph.ts and slightly modified
```
gndelia, line 1: Perhaps you may want to link here using the permalink, because once this is merged, that file will most likely be removed, and this comment won't make sense anymore

subgraph-api/src/subgraph.ts
```diff
@@ -0,0 +1,401 @@
...
+type Schema = {
+  query: string
+  variables?: Record<string, string | number>
+}
+
+type SuccessResponse<T> = { data: T }
+type ErrorResponse = { errors: { message: string }[] }
+type GraphResponse<T> = SuccessResponse<T> | ErrorResponse
+
+type ConfigType = typeof import('../config/default.json')
```
gndelia, line 23: nice, you're mastering the mystic art of typing here 😎

subgraph-api/src/subgraph.ts
```diff
@@ -0,0 +1,401 @@
...
+
+const request = <TResponse, TSchema extends Schema = Schema>(
+  url: string,
+  schema: TSchema,
+): Promise<TResponse> =>
+  fetch(url, {
+    body: JSON.stringify(schema),
+    headers: {
+      'Content-Type': 'application/json',
+      'Origin': 'https://app.hemi.xyz',
```
gndelia, line 50: should we move this into a ENV var?

subgraph-api/package.json
gabmontes: Express devs have just marked v5 as "latest" on Monday.

I will check the types tho.

subgraph-api/src/api-server.ts
gabmontes: That rule should be applied. Will check.

subgraph-api/tsconfig.json
gabmontes: 😅

subgraph-api/src/api-server.ts
gabmontes: Or return the promise. That may work too.

token-prices/api/package.json
gabmontes: Lerna was complaining. 🤷🏼‍♂️

subgraph-api/src/subgraph.ts
gabmontes: 😎

subgraph-api/package.json
gndelia: Ah, ok! The [5.0.1](https://www.npmjs.com/package/express/v/5.0.1) appears published 6 months ago, so I guess that confused me

subgraph-api/src/api-server.ts
gabmontes: The rule `import/order` was only set in `bloq/next`. I'm manually adding it.

subgraph-api/src/api-server.ts
gabmontes: Do not work. This is for `express@5` 🤷🏼‍♂️

gndelia (changes requested):

subgraph-api/src/api-server.ts
```diff
@@ -0,0 +1,176 @@
...
+    next()
+  } else {
+    next('route')
+  }
+}
+
+const isInteger = string => /^\d+$/.test(string)
+
+function validateQueryParams(req, res, next) {
+  const { fromBlock, limit, skip } = req.query
```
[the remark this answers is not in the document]
gabmontes: Currently, if the user does not specify `fromBlock`, the API will return 400. Do we want that or to default to 0?

subgraph-api/src/api-server.ts
```diff
@@ -0,0 +1,176 @@
...
+    },
+  )
+
+  app.use(function (req, res) {
+    sendJsonResponse(res, 404, { error: 'Not Found' })
+  })
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  app.use(function (err, req, res, next) {
+    if (config.get('DEBUG')) {
```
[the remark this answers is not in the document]
gabmontes: Oops!

gndelia (approved):

rafappelt (approved):

Issue #1013, opened 2025-03-11
Improve tunnel transactions tracking


gabmontes:
### Feature or change description

A user should know where their digital assets are at all times, even if they are "in flight", and even if they have multiple tunnel tx's simultaneously in flight.  Both [Superbridge](https://superbridge.app)[https://superbridge.app](https://superbridge.app/base) and Stargate provide the status of bridge transactions in flight, as some transactions might take ~7 days to complete.

In our ETH and BTC tunnels, in both Ethereum-to-Hemi and Hemi-to-Ethereum directions, there should never be a gap where transactions are invisibly processing, with zero visible UI feedback.

cc/ @dvnahuel @rmilrad

gndelia:
For Withdrawals, the "Claim" step automatically transfers the ETH/tokens, so nothing else should be done from that perspective. You can check the internal transactions (eth) or the logs (tokens) of `finalizeWithdrawal` transactions where you can see the 

The only thing that may be needed (to check) is that the appropriate balance query key (from `useQuery`) is being invalidated after the Claim is successful (as users may be seeing the old balance for a few seconds due to the react-query's cache).

gndelia:
For Bitcoin Deposits, if the user confirms it manually, The transaction itself sends the bitcoin, so we may just need to ensure that react-query's cache is being invalidated and re-fetched (this would be the balance of `hemiBTC`).
If the user does not confirm it manually, then it appears as pending until the vault operator operates on it, so nothing to do there

dvnahuel:
Hey team! Here's my proposal design-wise for this issue. For ETH deposits, we're going to add the 'Wait up to 3 minutes' message in the steps. As a nice-to-have, I’d also like to include a loading state animation on the 'Wait up to X minutes or hours' labels across all transactions (related to https://github.com/hemilabs/ui-monorepo/issues/695 as well).

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=11329-29982&t=GO4ewG6LfB0weN1K-1)
Screenshot:

<img width="567" alt="Image" src="https://github.com/user-attachments/assets/5d707ae6-8c7e-4d88-929e-9126ae9f8f73" />

Issue #37, opened 2024-02-01
Update OP stack (viem, wagmi, op-sdk and rainbox kit)


gndelia:
The [optimism-sdk](https://github.com/ethereum-optimism/optimism/tree/55a30f48b855dbcfb4cf77dac7b1864b026c20cd/packages/sdk) is on maintainance mode. It's not listed in their Npm, but it is on their github readme. They recommend going for [op-viem](https://viem.sh/op-stack).   
This works fine for us because we use `viem`, however, `op-viem` requires `viem 2.x` and we use `viem 1.x`... and we can't upgrade to that version because that breaks `RainboxKit 1.x`. We need for the publish of `Rainbox Kit 2.x` to upgrade `viem` and  `wagmi` to `2.x` and then replace the SDK for the new library

### Steps

- [x] Update `RainbowKit`, `Viem` and `Wagmi` to `2.x` https://github.com/BVM-priv/ui-monorepo/pull/106
- [ ] Replace `optimism-sdk` with `op-viem`. We may need an equivalent to [op-wagmi](https://github.com/base-org/op-wagmi) or develop our own wrappers

~~**Blocked** until [rainbox kit v2](https://www.npmjs.com/package/@rainbow-me/rainbowkit) is published~~

gndelia:
Accidentally closed - reopening

PR #1109, opened 2025-04-01
Add step to wait for Hemi deposit availability

ArturDolzan:
### Description

This PR introduces a new step in the deposit flow that informs the user to wait approximately three minutes for their funds to become available in Hemi.

To achieve this, we added a call to `crossChainMessenger.getMessageStatus` to check when the transaction has been fully relayed and acknowledged.

Additionally, an `invalidateQueries` call was added to `evmDepositsStatusUpdater.tsx` to ensure the balance is refreshed immediately after the deposit is confirmed by getMessageStatus.

With these changes, the user’s balance is now updated almost in real time.

### Screenshots

https://github.com/user-attachments/assets/35465e88-7b64-4769-ad3e-308a3d56a6db

### Related issue(s)

Related to #1013

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/types/tunnel.ts
+++ b/webapp/types/tunnel.ts
@@ -111,7 +111,11 @@ export const enum EvmDepositStatus {
   DEPOSIT_TX_FAILED = 4,
   // Approval failed
   APPROVAL_TX_FAILED = 5,
+  // Funds have been minted in Hemi
+  DEPOSIT_RELAYED = 6,
 }
+
+export const ExpectedWaitTimeMinutesGetFundsHemi = 3
 
 type CommonOperation = {
   amount: string
--- a/webapp/utils/watch/evmDeposits.ts
+++ b/webapp/utils/watch/evmDeposits.ts
@@ -1,8 +1,35 @@
+import { MessageStatus } from '@eth-optimism/sdk'
+import pMemoize from 'promise-mem'
 import { EvmDepositStatus, type EvmDepositOperation } from 'types/tunnel'
+import { findChainById } from 'utils/chain'
+import { createQueuedCrossChainMessenger } from 'utils/crossChainMessenger'
 import { getEvmBlock, getEvmTransactionReceipt } from 'utils/evmApi'
+import { createProvider } from 'utils/providers'
+import { Chain } from 'viem'
+
+const getCrossChainMessenger = pMemoize(
+  function (l1Chain: Chain, l2Chain: Chain) {
+    const l1Provider = createProvider(l1Chain)
+
+    const l2Provider = createProvider(l2Chain)
+
+    return createQueuedCrossChainMessenger({
+      l1ChainId: l1Chain.id,
+      l1Signer: l1Provider,
+      l2Chain,
+      l2Signer: l2Provider,
+    })
+  },
+  { resolver: (l1Chain, l2Chain) => `${l1Chain.id}-${l2Chain.id}` },
+)
 
 export const watchEvmDeposit = async function (deposit: EvmDepositOperation) {
   const updates: Partial<EvmDepositOperation> = {}
+
+  const l1Chain = findChainById(deposit.l1ChainId) as Chain
+  const l2Chain = findChainById(deposit.l2ChainId) as Chain
+
+  const crossChainMessenger = await getCrossChainMessenger(l1Chain, l2Chain)
 
   // check if it has completed on the background
   const receipt = await getEvmTransactionReceipt(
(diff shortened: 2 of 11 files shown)


webapp/types/tunnel.ts
```diff
@@ -93,6 +93,8 @@ export const enum EvmDepositStatus {
   DEPOSIT_TX_FAILED = 4,
   // Approval failed
   APPROVAL_TX_FAILED = 5,
+  // Relayed
```
gndelia, line 96: I'll add this for clarity
```suggestion
  // Funds have been minted in Hemi
```

gndelia (commented):
Do we need to update the "Status" column in the TX history to consider the new status?

image

I think we do... The Action should remain the same though ("View" button)

webapp/utils/watch/evmDeposits.ts
```diff
@@ -13,10 +40,22 @@ export const watchEvmDeposit = async function (deposit: EvmDepositOperation) {
...
-      : EvmDepositStatus.DEPOSIT_TX_FAILED
+
+  const status = await crossChainMessenger.getMessageStatus(
+    deposit.transactionHash,
+    0,
+    deposit.direction,
+  )
+
+  let newStatus: EvmDepositStatus
+  if (receipt.status === 'success' && status === MessageStatus.RELAYED) {
```
gndelia, line 51: strictly speaking, I think we can remove this check

```suggestion
  if (status === MessageStatus.RELAYED) {
```

If `status === MessageStatus.RELAYED`, then `receipt.status === 'success'` should always be true.

gndelia:
I left some suggestions. I haven't tested it, but from the video it looks really good to me!

webapp/utils/watch/evmDeposits.ts
ArturDolzan: Indeed. Simplified. Thanks!

webapp/utils/watch/evmDeposits.ts
```diff
@@ -1,9 +1,36 @@
+import { MessageStatus } from '@eth-optimism/sdk'
+import pMemoize from 'promise-mem'
 import { EvmDepositStatus, type EvmDepositOperation } from 'types/tunnel'
+import { findChainById } from 'utils/chain'
+import { createQueuedCrossChainMessenger } from 'utils/crossChainMessenger'
 import { getEvmBlock, getEvmTransactionReceipt } from 'utils/evmApi'
+import { createProvider } from 'utils/providers'
+import { Chain } from 'viem'
+
+const getCrossChainMessenger = pMemoize(
```
gabmontes, line 10: What are we going to do with this? I know we discussed this over Slack but what PR should we merge #1102 first or what?

In either case, we should keep one of the PRs in draft and review its status after the other is merged.

gndelia: I told artur to merge it as this feature is valuable, and I'd only have to rewrite it here

ArturDolzan: Sure, I can keep this PR open (as a draft, like you mentioned) until we have the new way to retrieve this data. The functionality should remain the same either way, so I’m good with whatever works best for you.

gabmontes: @gndelia @ArturDolzan are we updating this call right after the other PR is merged? If not or if we have the slightest chance of forgetting about this, let's open an issue and add a comment here with a link to it.

gndelia: > I told artur to merge it as this feature is valuable, and I'd only have to rewrite it here
To expand on this:

That PR replaces all usages with `getWithdrawalStatus`, but here, we need something like `getDepositStatus`. It is a different combination of functions in `viem`, and I need to implement it. And the first (and only) usage would be here so... let's better merge this first, so I can rework it after (and the feature becomes available)

gabmontes: Do we have an issue for that?

gndelia: yes, I'm doing this after this gets merged, but regardless, and replacing the whole `eth-optimism/sdk` as part of https://github.com/hemilabs/ui-monorepo/issues/37

gndelia (commented):
I just noticed that merging this PR  will break all past deposits  temporarily. As all deposits are in the state "DEPOSIT_TX_CONFIRMED", which is the past final state, but now, in order to be final, they need to move to the new state created as part of this PR to `DEPOSIT_RELAYED`

In the meantime, they will all appear as "wait up to 3 minutes"

image


The good news is that it should auto-fix, as the watcher will iterate each deposit, and update the new status. That was what happened while I was typing this comment for example, and now it looks like this

image


However, for users with many deposits, it may take a few minutes until all deposits are updated


cc @gabmontes

webapp/utils/watch/evmDeposits.ts
gndelia: @gabmontes @ArturDolzan another option is me sending a commit to this PR. We'll have to coordinate a bit though

gabmontes: If you guys don't forget, I'm ok.

gabmontes:
@gndelia I'm ok with this temporary auto-fixable glitch. If there is no safe workaround, let's keep it simple and let it auto fix itself.

gndelia (approved):

ArturDolzan:
> @ArturDolzan conflicts!

Rebased. Thanks!

gndelia:
I'm sorry, you will need to rebase this after merging #1111  😓

gndelia (approved):

gabmontes (approved):

PR #1112, opened 2025-04-02
Replace "getMessageStatus" when getting deposit status

gndelia:
### Description

This PR is a follow-up of #1109. It drops the added usage of `getMessageStatus` and replaces it with its equivalent call from `viem`.

- 95fd0420776e6ab1a15bb1658fb59c1627e3cbaa Does that replacement. In addition to that, it also adds the `l2TransactionHash`. We could not get it with the `eth-optimism/sdk`, but we can now get this hash with `viem`, so it is now stored in the deposit.
- cd1055207a360881308b9b52977897cea18fa41b Updates the drawer to show this hash in the last step
- 260ea03620ce31882679c92cf9b56c7248d6a3d6 As we learnt from the changes applied from Artur, the hemi balance is not updated as soon as the deposit is confirmed on the L1 chain. Samewise, the balance on the L1 is not updated as soon as the L2 withdrawal is initiated. Because of this, after initiating a deposit/withdrawal, there's no need to invalidate the balance on the target chain. This commit, then, removes the code that was doing exactly that. It also waits for the receipt.status to be defined
- e56e8b3517232f78dcd4f21cc91efefe71df0525 Removes the default status for EVM deposits when resyncing. Before #1109 the default status was `DEPOSIT_TX_CONFIRMED`, which was final. That is no longer the case. So when resyncing, all deposits would appear in their history as "Wait up to 3 minutes (Estimated)". By removing the status (letting the statusUpdater to pick up the appropriate status), a loading skeleton is shown, until the appropriate status is loaded.

### Screenshots

The Hemi transaction hash is now linked to the drawer in the last step

image

See now how the tx-history is reload

https://github.com/user-attachments/assets/ba5428f7-8944-4c82-9a69-329a99345bb6

### Related issue(s)

Related to #37
Related to #1013

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/tunnel/_components/reviewOperation/reviewEvmDeposit.tsx
+++ b/webapp/app/[locale]/tunnel/_components/reviewOperation/reviewEvmDeposit.tsx
@@ -123,10 +123,12 @@ const ReviewContent = function ({
 
   const getFundsHemiStep = (): StepPropsWithoutPosition => ({
     description: t('get-your-funds-on-hemi'),
+    explorerChainId: deposit.l2ChainId,
     status:
       depositStatus === EvmDepositStatus.DEPOSIT_RELAYED
         ? ProgressStatus.COMPLETED
         : ProgressStatus.NOT_READY,
+    txHash: deposit.l2TransactionHash,
   })
 
   // Show the approval only if it's a not native token and there is a approval.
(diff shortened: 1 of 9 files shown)


ArturDolzan (approved):
LGTM

gabmontes (approved):

Issue #1113, opened 2025-04-02
Move packages into folder "packages/*"


gndelia:
There are a few packages that are not websites nor APIs scattered in the root. We should move them into a `packages/*` folder, as this is standard in many monorepo tools. This will organize the code a little bit better

I believe:

- [x] btc-wallet
- [x] hemi-viem-stake-actions

should be moved

I think `sliding-block-window` will be removed once we implement [this](https://github.com/hemilabs/ui-monorepo/issues/1061), so I wouldn't move it.  
I also think we should finally get rid of `ui-common` (See #1183 )

PR #1115, opened 2025-04-04
Drop OP-SDK part 3

gndelia:
### Description

Another set of pre-changes related to removing the OP SDK. This may be easier to review by commit

- 24e7e0f7bbee24c80869ed0419d82d3fa1f9ac11 Updates the `useAllowance` to accept string instead of `Address`. When using this hook, we always had to add a `@ts-expect-error` because many times, the token using it may not be an ERC20 token. Now, this is handled inside the hook, so we no longer need to add that each time we use it
- ccbca3c0fdc0c012a65ab670906d1c1629836941 Renders a Skeleton in the TX-History when the timestamp is missing. The rationale is the following: When a tx confirms, the receipt has the block number, but the `timestamp` is in the block. For that, we had to use an `useEffect` to get the block and then save the timestamp. This made error handling more difficult. Plus, if the user left the component before the block was retrieved, we would lose the timestamp! For that, there was a backup mechanism to retrieve the missing info from the deposit using the web workers... But then I thought, what if we directly use the web workers for that? That way, the `deposit.tsx` would be cleaner, as it wouldn't have to do that. This commit, then, adds a `Skeleton` to show instead of "nothing" while the timestamp is not set. Note that I'm not removing that `useEffect` now, this will be done in an incoming PR. But this commit is useful to show something in those cases where the backup process hasn't kicked in. Once we drop the OP-SDK, the backup process will be the main process to get the timestamp of a transaction :)
- d866587c0bace41dfc7b8659314a3f0947f330de Creates a `viem-erc20` package that uses `.extend()` to extend a public (or a wallet) client with viem for doing some erc20 requests. This separated package is needed because the incoming package that replaces the OP-SDK needs these functions to run validations before depositing. So by having them in a separated package, I can use them both in the portal and in the OP-SDK replacement
- d54f90f06442c56f50f2c82fead531d2cb209421 Now removes the functions from the prior commit from the portal, and uses the package

### Screenshots

No visible changes to users

### Related issue(s)

Related to #37
Related to #1113  (as I have created the `packages/*` folder

### Checklist

- [x] Manual testing passed.


packages/viem-erc20/src/wallet/approve.ts
```diff
@@ -0,0 +1,36 @@
...
+    throw new Error('Invalid amount, must be greater than 0')
+  }
+
+  return writeContract(client, {
+    abi: erc20Abi,
+    // @ts-expect-error: TS is complaining about client.account definition, but this works
+    account: client.account,
+    address,
+    args: [spender, amount] as const,
+    // @ts-expect-error: TS is complaining about client.account definition, but this works
```
gabmontes, line 32: Wrong comment?

gndelia: Yes! [3081a7d](https://github.com/hemilabs/ui-monorepo/pull/1115/commits/3081a7d0967a2c35be3bc8043d1af6835a32f9ac)

gabmontes (approved):

rafappelt (approved):

PR #1119, opened 2025-04-07
Drop OP SDK - Part 4 - Creating a tunnel package (with deposit)

gndelia:
### Description

This PR creates a package with the `depositETH` and `depositErc20Token` functions. These will be used to deposit from the portal. The interface is an improvement over the `stake` functions as the `emitter` will let us to have a simpler interface.

The next PR will use this package from `evmDeposit.tsx` to make a deposit. It will update `useDeposit` by rewriting it, and dropping all the OP-SDK code related to depositing.

Eventually, this package could be published into npm

### Screenshots

No visible changes, this is just a package to be used

### Related issue(s)

Related to #37

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/packages/hemi-tunnel-actions/index.ts
+++ b/packages/hemi-tunnel-actions/index.ts
@@ -1,2 +1,2 @@
-export { depositErc20Token } from './src/depositErc20Token'
+export { depositErc20 } from './src/depositErc20'
 export { depositEth } from './src/depositEth'
--- a/packages/hemi-tunnel-actions/index.ts
+++ b/packages/hemi-tunnel-actions/index.ts
@@ -1,0 +1,2 @@
+export { depositErc20Token } from './src/depositErc20Token'
+export { depositEth } from './src/depositEth'
--- a/packages/hemi-tunnel-actions/package.json
+++ b/packages/hemi-tunnel-actions/package.json
@@ -1,0 +1,37 @@
+{
+  "name": "hemi-tunnel-actions",
+  "version": "1.0.0",
+  "description": "Hemi Tunnel Actions",
+  "keywords": [
+    "eth",
+    "ethereum",
+    "hemi",
+    "stake",
+    "viem"
+  ],
+  "license": "MIT",
+  "author": {
+    "email": "gonzalo@hemi.xyz",
+    "name": "Gonzalo D'Elia"
+  },
+  "scripts": {
+    "test": "vitest run",
+    "tsc:check": "tsc"
+  },
+  "dependencies": {
+    "hemi-viem": "2.3.2",
+    "viem-erc20": "1.0.0"
+  },
+  "devDependencies": {
+    "@types/node": "20.12.12",
+    "typescript": "5.6.3",
+    "viem": "2.22.10",
+    "vitest": "3.0.8"
+  },
+  "peerDependencies": {
+    "viem": "^2.x"
+  },
+  "exports": {
+    ".": "./index.ts"
+  }
+}
--- a/packages/hemi-tunnel-actions/src/depositErc20.ts
+++ b/packages/hemi-tunnel-actions/src/depositErc20.ts
@@ -163,6 +163,10 @@ const runDepositErc20 = ({
         emitter.emit('user-signing-deposit-error', error)
       })
 
+      if (!depositHash) {
+        return
+      }
+
       await handleWaitDeposit({
         emitter,
         hash: depositHash,
--- a/packages/hemi-tunnel-actions/src/depositErc20.ts
+++ b/packages/hemi-tunnel-actions/src/depositErc20.ts
@@ -108,7 +108,7 @@ const runDepositErc20 = ({
     })
 
     if (amount > allowance) {
-      emitter.emit('on-approve')
+      emitter.emit('pre-approve')
       const approveHash = await extendedL1WalletClient
         .approveErc20Token({
           address: l1TokenAddress,
@@ -120,7 +120,7 @@ const runDepositErc20 = ({
         })
 
       if (!approveHash) {
-        emitter.emit('on-deposit-settled')
+        emitter.emit('deposit-settled')
         return
       }
 
@@ -139,7 +139,7 @@ const runDepositErc20 = ({
       )
     }
 
-    emitter.emit('on-deposit')
+    emitter.emit('pre-deposit')
 
     // Using @ts-expect-error fails to compile so I need to use @ts-ignore
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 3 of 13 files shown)


packages/hemi-tunnel-actions/index.ts
```diff
@@ -0,0 +1,2 @@
+export { depositErc20Token } from './src/depositErc20Token'
```
gabmontes, line 1: All ERC-20 are tokens so `depositErc20Token` could easily be `depositErc20` with the exact same meaning but shorter. Just a note in the case we feel like shortening the names.

packages/hemi-tunnel-actions/src/types.ts
```diff
@@ -0,0 +1,21 @@
+import { type Hash, TransactionReceipt } from 'viem'
+
+export type DepositEvents = {
+  'deposit-failed': [Error]
+  'deposit-failed-validation': [string]
+  'on-deposit-settled': []
```
gabmontes, line 6: Events prefixed with `on-` might be renamed. The event handlers are usually prefixed that way but not the event names themselves.

packages/hemi-tunnel-actions/src/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+  hash,
+  publicClient,
+}: {
+  emitter: EventEmitter<EventMap<T>>
+  hash: Hash | undefined | void
+  publicClient: PublicClient
+}) {
+  if (!hash) {
+    return
+  }
```
gabmontes, line 63: This check should be hoisted to the caller as it is the way to handle the call to `writeContract()` failed. Same as done here in L74-76. It "logically" belongs with the previos call.

packages/hemi-tunnel-actions/src/depositEth.ts
```diff
@@ -0,0 +1,113 @@
...
+      emitter.emit('user-signing-deposit-error', error)
+    })
+
+    await handleWaitDeposit({
+      emitter,
+      hash: depositHash,
+      publicClient: l1PublicClient,
+    })
+
+    emitter.emit('on-deposit-settled')
```
gabmontes, line 109: This will be emitted even when `handleWaitDeposit` emits an error (as it just `return`s after doing so). Is that the intended behavior?

gndelia: Yes, this is an intended behavior.
This mimics the `onSettle` event from `useMutations` from `react-query` which always fires at the end, regardless of the outcome. It's used mostly for cleanup. I use it to invalidate balances so they are refetched in the background

packages/hemi-tunnel-actions/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "hemi-tunnel-actions",
```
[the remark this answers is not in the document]
gndelia: It could... it could also be published on its own. 

In the meantime, I preferred to work on it here for the speed of development. Once it's out there in the wild, it should be easier to either move it to hemi-viem or as a stand-alone package.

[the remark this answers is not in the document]
gndelia: Remember that a few more methods will be added (for withdraw, prove, claim). Having this in a separated repo would definitely slow down this process

packages/hemi-tunnel-actions/index.ts
gndelia: Renamed here [e4ee98a](https://github.com/hemilabs/ui-monorepo/pull/1119/commits/e4ee98a05f09e0c6a9657388dedb2559b7bafc6c)

packages/hemi-tunnel-actions/src/types.ts
gndelia: Events renamed here. [a346ac2](https://github.com/hemilabs/ui-monorepo/pull/1119/commits/a346ac2fe38d683aa2b5811488ce71ced0e16ef9)

packages/hemi-tunnel-actions/src/depositErc20Token.ts
```diff
@@ -0,0 +1,170 @@
...
+        emitter.emit('on-deposit-settled')
+        return
+      }
+
+      emitter.emit('user-signed-approve', approveHash)
+
+      const approveReceipt =
+        await extendedL1PublicClient.waitForTransactionReceipt({
+          hash: approveHash,
+        })
```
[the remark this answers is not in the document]
gndelia: Yes, Added here [d479ce5](https://github.com/hemilabs/ui-monorepo/pull/1119/commits/d479ce53b71f6d069e2fce043b6d23622becebef)

I also rewrote a little bit the code to ensure the `settled` always fire at the end

packages/hemi-tunnel-actions/src/utils.ts
gndelia: Done [03d3a22](https://github.com/hemilabs/ui-monorepo/pull/1119/commits/03d3a2230215d9db9837f7c15efbc6bd5397ac2d)

gabmontes (dismissed):

packages/hemi-tunnel-actions/src/depositErc20.ts
```diff
@@ -0,0 +1,181 @@
...
+  l1WalletClient: WalletClient
+  l2Chain: Chain
+  l2TokenAddress: Address
+}) =>
+  async function (emitter: EventEmitter<DepositErc20Events>) {
+    try {
+      const extendedL1PublicClient = l1PublicClient.extend(erc20PublicActions())
+      const extendedL1WalletClient = l1WalletClient.extend(erc20WalletActions())
+
+      const { canDeposit, reason } = await canDepositErc20({
```
ArturDolzan, line 88: It looks like if this call somehow fails (since there’s an await without a try/catch inside canDepositEth), the error would propagate here. but there’s no catch block in this function, only a finally. Should we handle unexpected errors explicitly here (e.g., by emitting an unexpected-error event), or is this the intended behavior? Same for erc20.

gndelia: you're correct, I will add a catch to handle unexpected errors.

gndelia: Applied here [93c15cb](https://github.com/hemilabs/ui-monorepo/pull/1119/commits/93c15cb4e110a69fdc29b55d1fc98a14f2c9bcd0)

I also applied a little change to `validateInputs` that will be needed for the withdrawal, so I'm bringing it here for easier review

ArturDolzan (approved):
LGTM

Issue #1124, opened 2025-04-08
Fee estimations are not correct


gndelia:
When running operations, estimations are not the same as what the wallet shows. For example, depositing ETH on testnet:

- Shows this estimation in the tunnel form

<img width="209" alt="Image" src="https://github.com/user-attachments/assets/7f90085b-1cce-4fb4-a59e-aac1ed9b7aea" />

- Show this estimation in the wallet prompt

<img width="354" alt="Image" src="https://github.com/user-attachments/assets/957d440d-002e-448e-b913-5f076d40ee74" />

- Has this transaction fee in the explorer

<img width="602" alt="Image" src="https://github.com/user-attachments/assets/a5ab3b3c-a94f-4549-ba06-057fec28dff4" />

Although there's a small acceptable difference between the wallet and the final gas spent, the estimation of the app is orders or magnitude smaller.

Now, there are 2 parts to calculate fees: Calculating the gas units and then estimating the price. Units * price give us the fees

Getting the gas units is currently hardcoded. See https://github.com/hemilabs/ui-monorepo/issues/866 I also have some work of that incoming as part of fixing #37

The rest comes from calculating the price. That is done in [useEstimateFees](https://github.com/hemilabs/ui-monorepo/blob/ac8071459571619865abaa4d842ecc90f218f7aa/webapp/hooks/useEstimateFees.ts#L49)

I did a quick check, and `baseFeePerGas` and `maxPriorityFeePerGas` seem to show values different from the ones calculated in the final transaction. We need to review this implementation to see what's wrong. It may also be that mainnet works ok and it is a testnet issue - needs to be reviewed

Note: This is a problem in all operations of the app that run either in Sepolia/Mainnet or in Hemi

Issue #866, opened 2025-02-17
Improve fee estimation across the app


gndelia:
Both the tunnel and the incoming Stake functionality use hardcoded unit gas limits to estimate fees. We should look into using `estimateGas` and `simulate` operations to get better gas estimations.

Related to https://github.com/hemilabs/ui-monorepo/issues/826 (We may need to close that one in favor of this)

gndelia:
Related https://viem.sh/docs/contract/estimateContractGas.html

PR #1125, opened 2025-04-09
Use hemi-tunnel-actions to deposit

gndelia:
### Description

This PR updates `evmDeposit.tsx` to use the new `depositETH` and `depositErc20` functions created to run these operations, instead of using the OP-SDK.

The `useDeposit` is largely updated, and it is created as an intermediary between the UI and the library that executes the transaction(s). Because of this, the `useDepositToken.ts` can be removed (both types of deposit are handled from the former). We no longer use `useEffects` which makes the code a bit easier to follow

The `encodeDepositETH` and `encodeDepositErc20` methods are created, and used for better estimations of gas units for these 2 operations. 

The code removed in `useL2Bridge.ts` is the OP-SDK code that can now be removed.

### Screenshots

There is no visible difference to users 😅  Visually, deposit works the same

### Related issue(s)

Related to #37 (Removing OP-SDK)
Related to #1124 (issue created, to improve fee estimations)
Related to #866 (improved gas units estimations for `depositETH` and `depositErc20`
Related to #539 (Partially removing the fixed gasLimit for sepolia)

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/hooks/useNeedsApproval.ts
+++ b/webapp/hooks/useNeedsApproval.ts
@@ -16,15 +16,17 @@ export const useNeedsApproval = function ({
   const {
     data: allowance = BigInt(0),
     isLoading,
-    status: allowanceStatus,
+    isError,
     queryKey: allowanceQueryKey,
+    status: allowanceStatus,
   } = useAllowance(address, {
     args: { owner, spender },
   })
 
   return {
     allowanceQueryKey,
-    isLoadingAllowance: isLoading,
+    isAllowanceError: isError,
+    isAllowanceLoading: isLoading,
     needsApproval: amount > allowance && allowanceStatus === 'success',
   }
 }
--- a/webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
+++ b/webapp/app/[locale]/tunnel/_components/evmDeposit.tsx
@@ -41,16 +41,16 @@ type EvmDepositProps = {
 
 const SubmitEvmDeposit = function ({
   canDeposit,
+  isAllowanceLoading,
   isRunningOperation,
   needsApproval,
   operationRunning,
-  isAllowanceLoading,
 }: {
   canDeposit: boolean
+  isAllowanceLoading: boolean
   isRunningOperation: boolean
   needsApproval: boolean
   operationRunning: OperationRunning
-  isAllowanceLoading: boolean
 }) {
   const t = useTranslations()
 
@@ -68,6 +68,7 @@ const SubmitEvmDeposit = function ({
     if (isAllowanceLoading) {
       return <Spinner size={'small'} />
     }
+    // TODO Se need to handle allowanceStatus === 'error, see https://github.com/hemilabs/ui-monorepo/pull/1125
     if (!isRunningOperation) {
       return texts[needsApproval ? 'approve' : 'deposit'].idle
     }
@@ -122,7 +123,7 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
 
   const l1StandardBridgeAddress = useL1StandardBridgeAddress(fromToken.chainId)
 
-  const { isLoadingAllowance, needsApproval } = useNeedsApproval({
+  const { isAllowanceLoading, needsApproval } = useNeedsApproval({
     address: fromToken.address,
     amount,
     spender: l1StandardBridgeAddress,
@@ -215,7 +216,7 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
       return (
         <SubmitEvmDeposit
           canDeposit={canDeposit}
-          isAllowanceLoading={isLoadingAllowance}
+          isAllowanceLoading={isAllowanceLoading}
           isRunningOperation={isRunningOperation}
           needsApproval={needsApproval}
           operationRunning={operationRunning}


webapp/hooks/useNeedsApproval.ts
```diff
@@ -0,0 +1,30 @@
...
+    status: allowanceStatus,
+    queryKey: allowanceQueryKey,
+  } = useAllowance(address, {
+    args: { owner, spender },
+  })
+
+  return {
+    allowanceQueryKey,
+    isLoadingAllowance: isLoading,
+    needsApproval: amount > allowance && allowanceStatus === 'success',
```
gabmontes, line 28: If `allowanceStatus` is not `success`, `needsApproval` will be `false`, which can be not the intended behavior. Perhaps the status should be exposed as `useAllowance` does, so the user decides what to do.

webapp/hooks/useL1WalletClient.ts
```diff
@@ -0,0 +1,13 @@
+import { Chain } from 'viem'
+import { useWalletClient } from 'wagmi'
+
+export const useL1WalletClient = function (chainId: Chain['id']) {
```
gabmontes, line 4: This hook seems to just rename "data" to "l1WalletClient". It may not worth having it at all and just use `useWalletClient` instead.

webapp/app/[locale]/tunnel/_hooks/useDeposit.ts
```diff
@@ -1,303 +1,221 @@
...
+  const { allowanceQueryKey } = useNeedsApproval({
+    address: fromToken.address,
+    amount,
+    spender: l1StandardBridgeAddress,
+  })
 
-      updateDeposit(deposit, {
-        status: EvmDepositStatus.DEPOSIT_TX_FAILED,
+  return useMutation({
+    mutationFn() {
```
Comment (gabmontes) [inline webapp/app/[locale]/tunnel/_hooks/useDeposit.ts:80]: What about using a name that is more descriptive?

webapp/app/[locale]/tunnel/_hooks/useDeposit.ts
```diff
@@ -1,303 +1,221 @@
...
+            l1Chain: findChainById(fromToken.chainId) as Chain,
+            l1PublicClient: getEvmL1PublicClient(fromToken.chainId),
+            l1WalletClient,
+            l2Chain: findChainById(toToken.chainId) as Chain,
+          })
+        : depositErc20({
+            account: address,
+            amount,
+            approvalAmount:
+              !depositingNative && extendedErc20Approval
```
Comment (gabmontes) [inline webapp/app/[locale]/tunnel/_hooks/useDeposit.ts:97]: `depositingNative` is always false here.

webapp/app/[locale]/tunnel/_hooks/useEstimateDepositFees.ts
```diff
@@ -0,0 +1,40 @@
...
+  fromToken,
+  toToken,
+}: {
+  amount: bigint
+  enabled?: boolean
+  fromToken: EvmToken
+  toToken: EvmToken
+}) {
+  const l1StandardBridge = useL1StandardBridgeAddress(fromToken.chainId)
+  const { data: gasUnits, isSuccess } = useEstimateGas({
```
Comment (gabmontes) [inline webapp/app/[locale]/tunnel/_hooks/useEstimateDepositFees.ts:20]: When calling `useEstimateGas` for an ETH deposit, `value` shall be sent too. For ERC-20 deposits, it is 0.

Comment (gndelia) [reply webapp/app/[locale]/tunnel/_hooks/useEstimateDepositFees.ts:21]: Added here ! [2da1bd3](https://github.com/hemilabs/ui-monorepo/pull/1125/commits/2da1bd36c4bbf84fd8d32953cd6fc1c8c1dab13c)

webapp/app/[locale]/tunnel/_hooks/useDeposit.ts
Comment (gndelia) [reply webapp/app/[locale]/tunnel/_hooks/useDeposit.ts]: Indeed, removed [0b23631](https://github.com/hemilabs/ui-monorepo/pull/1125/commits/0b236310ab704fd5877937ec4c6164ba85633df7)

webapp/app/[locale]/tunnel/_hooks/useDeposit.ts
Comment (gndelia) [reply webapp/app/[locale]/tunnel/_hooks/useDeposit.ts]: `mutationFn` is the name that `useMutation` expects 🤷🏽  can't do anything about it. However, I updated it so the inline function as a name [07b169d](https://github.com/hemilabs/ui-monorepo/pull/1125/commits/07b169d96299cc9f8d138d8808ce51c94d3a1c57)

webapp/hooks/useL1WalletClient.ts
gndelia: Granted, removed [fedc8e2](https://github.com/hemilabs/ui-monorepo/pull/1125/commits/fedc8e25823bbad804614fd902f1aec32e1f5f51)

webapp/hooks/useNeedsApproval.ts
gndelia: Improved here [19bf37e](https://github.com/hemilabs/ui-monorepo/pull/1125/commits/19bf37ec5b69aaa0926bf57bc744bfde67f8510d) created issue #1126 

Note that I still think that it is ok to check for `success`, as this hook may be called for native tokens. The query will be disabled for those scenarios, so `allowanceStatus` won't be `success`, and `needsApproval` result in false, which is correct.

Usages of this hook then will still need to check for `loading` and `error`, and that's why these 2 are now exposed (well, `loading` was already exposed heh)

gabmontes: For native tokens, this hook should just return `false`. No need do any actual query.

webapp/app/[locale]/tunnel/_hooks/useDeposit.ts
Comment (gabmontes) [reply webapp/app/[locale]/tunnel/_hooks/useDeposit.ts]: Oh! Didn't see it was an object property!

webapp/hooks/useNeedsApproval.ts
gndelia: The query is not running. See [useAllowance](https://github.com/hemilabs/ui-monorepo/blob/543ddc46c31939b82f58aa16e8b643faba3c1be1/webapp/hooks/useAllowance.ts#L30). But in order to check here that it did not run, I need to check for `allowanceStatus` (or to check something at least 😆 )

gabmontes (approved):

ArturDolzan (approved):

PR #1135, opened 2025-04-10
Drop OP SDK pt 6: Add initiateWithdraw to hemi-tunnel-actions

### LINKED ISSUE #37: Update OP stack (viem, wagmi, op-sdk and rainbox kit)

gndelia:
### Description

This PR creates the functions to initiate a withdraw using `viem`. These are added to `hemi-tunnel-actions`.

It works both for ETH and `erc20`. The next PR will be using this function to run the withdraw from the portal

### Screenshots

No visible changes to users

### Related issue(s)

Closes #37 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/packages/hemi-tunnel-actions/src/initiateWithdraw.ts
+++ b/packages/hemi-tunnel-actions/src/initiateWithdraw.ts
@@ -1,0 +1,232 @@
+import { EventEmitter } from 'events'
+import {
+  Address,
+  Chain,
+  encodeFunctionData,
+  PublicClient,
+  WalletClient,
+} from 'viem'
+import { writeContract } from 'viem/actions'
+import { erc20PublicActions } from 'viem-erc20'
+
+import { l2BridgeAbi } from './abis'
+import { WithdrawEvents } from './types'
+import { getL2BridgeAddress, toPromiseEvent, validateInputs } from './utils'
+
+const hasErc20Balance = ({
+  account,
+  publicClient,
+  tokenAddress,
+}: {
+  account: Address
+  publicClient: PublicClient
+  tokenAddress: Address
+}) =>
+  publicClient.extend(erc20PublicActions()).getErc20TokenBalance({
+    account,
+    address: tokenAddress,
+  })
+
+const hasEthBalance = ({
+  account,
+  publicClient,
+}: {
+  account: Address
+  publicClient: PublicClient
+}) => publicClient.getBalance({ address: account })
+
+const canInitiateWithdraw = async function ({
+  account,
+  amount,
+  checkBalance,
+  l1Chain,
+  l2Chain,
+}: {
+  account: Address
+  amount: bigint
+  checkBalance: () => Promise<string | undefined>
+  l1Chain: Chain
+  l2Chain: Chain
+}): Promise<{
+  canWithdraw: boolean
+  reason?: string
+}> {
+  const reason = validateInputs({
+    account,
+    amount,
+    l1Chain,
+    l2Chain,
+  })
+  if (reason) {
+    return { canWithdraw: false, reason }
+  }
+
+  const balanceReason = await checkBalance()
+
+  if (balanceReason) {
+    return { canWithdraw: false, reason: balanceReason }
+  }
+  return { canWithdraw: true }
+}
+
+type InitiateWithdraw = {
+  account: Address
+  amount: bigint
+  checkBalance: () => Promise<string | undefined>
+  l1Chain: Chain
+  l2Chain: Chain
+  l2PublicClient: PublicClient
+  l2TokenAddress: Address
+  l2WalletClient: WalletClient
+  value?: bigint | undefined
+}
+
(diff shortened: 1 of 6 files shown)


packages/hemi-tunnel-actions/src/initiateWithdraw.ts
```diff
@@ -0,0 +1,232 @@
...
+      amount,
+      async checkBalance() {
+        const balance = await hasEthBalance({
+          account,
+          publicClient: l2PublicClient,
+        })
+        // using >= because we need to consider gas fees
+        if (amount >= balance) {
+          return 'insufficient balance'
+        }
```
gabmontes, line 196: We should use ">" instead of ">=". And could we have a raw estimate of the fees so we can "amount >= balance + fee"? And I think the comment is wrong. Double check.

packages/hemi-tunnel-actions/src/initiateWithdraw.ts
```diff
@@ -0,0 +1,232 @@
...
+      amount,
+      async checkBalance() {
+        const balance = await hasErc20Balance({
+          account,
+          publicClient: l2PublicClient,
+          tokenAddress: l2TokenAddress,
+        })
+        if (amount > balance) {
+          return 'insufficient balance'
+        }
```
gabmontes, line 225: If amount === balance, we can execute the withdrawal. Can we also check if the user has enough ETH to pay the (guesstimated) fees?

gndelia: Yes, I think I inverted the conditions between initiateWithdrawEth and initiateWithdrawErc20 🤦🏽

packages/hemi-tunnel-actions/src/initiateWithdraw.ts
gndelia: > And could we have a raw estimate of the fees so we can "amount >= balance + fee"? 

We could, but I thought it would not be very easy.

packages/hemi-tunnel-actions/src/initiateWithdraw.ts
```diff
@@ -0,0 +1,232 @@
...
+    address: tokenAddress,
+  })
+
+const hasEthBalance = ({
+  account,
+  publicClient,
+}: {
+  account: Address
+  publicClient: PublicClient
+}) => publicClient.getBalance({ address: account })
```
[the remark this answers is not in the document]
gndelia: Renamed [054e18c](https://github.com/hemilabs/ui-monorepo/pull/1135/commits/054e18cee92d926dbe6a9d751ee644ff27ca049c)

packages/hemi-tunnel-actions/src/initiateWithdraw.ts
gndelia: > Can we also check if the user has enough ETH to pay the (guesstimated) fees?

I can check it has balances

gndelia: Added check for balance here [9c7b658](https://github.com/hemilabs/ui-monorepo/pull/1135/commits/9c7b658968de1b317167195d350be0a0898e38f6)

packages/hemi-tunnel-actions/src/initiateWithdraw.ts
gndelia: plus the wallet already does this for us 😄

gndelia:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **TypeError: Cannot read properties of undefined (reading 'call')** `/en/stake/` [View Issue](https://hemi-labs.sentry.io/issues/38299333/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

See https://github.com/hemilabs/ui-monorepo/issues/1093

PR #1146, opened 2025-04-16
Get timestamp for failed withdrawals

### LINKED ISSUE #1127: Failed transactions are missing the timestamp
I have a failed withdrawal in my TX history

Image

The timestamp is missing. This may be ok (there are certain circumstances, for example, the user closing the page in a specific moment) where this may happen. That's why the web workers that work in the background also check if the timestamp is missing from withdrawals, and they add it if needed. See [here](https://github.com/hemilabs/ui-monorepo/blob/543ddc46c31939b82f58aa16e8b643faba3c1be1/webapp/utils/watch/evmWithdrawals.ts#L68).

However, it seems failed withdrawals are being skipped, so the timestamp is never added

If you need to recreate a failed withdraw, you can manipulate the local code to send a withdrawal with `value: 0`, so it reverts

gndelia:
### Description

This PR fixes failed withdrawals that do not show the timestamp in the TX history.

### Screenshots

image

### Related issue(s)

Closes #1127

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/webapp/app/[locale]/tunnel/_hooks/useWithdraw.ts
+++ b/webapp/app/[locale]/tunnel/_hooks/useWithdraw.ts
@@ -111,6 +111,7 @@ export const useWithdraw = function ({
         track?.('evm - init withdraw failed', { chain: networkType })
 
         updateWithdrawal(withdrawal, {
+          blockNumber: Number(receipt.blockNumber),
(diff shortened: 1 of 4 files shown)


alvarengathomas (approved):

gabmontes (approved):
LGTM

PR #1147, opened 2025-04-16
Drop OP SDK part 10: Add action to finalize a withdrawal

gndelia:
### Description

We're getting there! This is the PR 10 of the series to drop the OP SDK. The next one is the last one!

This PR adds the `finalize` action into `hemi-tunnel-actions`. This will be used in the Claim step. Note that I use _Finalize_ because that's the method name in the contract (And in `viem`). Claiming is something from our UI.

Next PR will use this action to claim withdrawals.

### Screenshots

No visible changes - this just adds an action.

### Related issue(s)

Related to #37

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/packages/hemi-tunnel-actions/src/finalizeWithdrawal.ts
+++ b/packages/hemi-tunnel-actions/src/finalizeWithdrawal.ts
@@ -61,12 +61,14 @@ const canFinalizeWithdrawal = async function ({
     chain: l1WalletClient.chain,
     receipt,
     targetChain: l2PublicClient.chain,
-  })
+  }).catch(() => null)
 
   if (withdrawalStatus !== 'ready-to-finalize') {
     return {
       canFinalize: false,
-      reason: `Withdrawal status is not ready-to-finalize, current status: ${withdrawalStatus}`,
+      reason: withdrawalStatus
+        ? `Withdrawal status is not ready-to-finalize, current status: ${withdrawalStatus}`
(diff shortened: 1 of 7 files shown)


packages/hemi-tunnel-actions/src/finalizeWithdrawal.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  // Using @ts-expect-error fails to compile so I need to use @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore because it works on IDE, and when building on its own, but fails when compiling from the portal through next
+  const withdrawalStatus = await getWithdrawalStatus(l1WalletClient, {
+    chain: l1WalletClient.chain,
+    receipt,
+    targetChain: l2PublicClient.chain,
+  })
```
gabmontes, line 64: What if this call fails? It looks like the function could be more consistent if it always return an object with `canFinalize` and `reason` instead if also being able to throw. It makes the user have to consider more scenarios.

gndelia: Applied here [e870f3d](https://github.com/hemilabs/ui-monorepo/pull/1147/commits/e870f3dc65b0346a3faf5da703fa192fcb752d5c)

rafappelt (approved):

ArturDolzan (approved):

PR #1158, opened 2025-04-23
Prevent subgraph api to get "limit" parameter larger than 5000

### LINKED ISSUE #1145: Resyncing the TX-history may break after 5000 tunnel operations
As part of running a script that downloads all the information from a subgraph, I faced this error

> bad indexers: {0xbdfb5ee5a2abf4fc7bb1bd1221067aef7f9de491: BadResponse(unattestable response: The skip argument must be between 0 and 5000, but is 5100)}

When we resync the tx-history from scratch, we start from a block, and we request operations in chunks of size 100 (which is also the max number allowed per The Graph). Each chunk returns 100 entities, and increases the "skip" value by 100.

It turns out, as that error shows, that skip's max value can be 5000. After that, the query throws an error. If a user had more than 5000 tunnel operations when resyncing from scratch, they would face this issue.

One solution is, once we're near to hitting 5000 skip value, to set it to 0 and use that last known `fromBlock` value from the last response. There may be some transaction hash collisions (as we would be querying a block that we may have partially requested), so duplicates may need to be removed (I think the tx-history already handles duplicates, though).

gndelia:
### Description

This PR solves a corner case of the resync history process, in which when a user has more than 5000 tunneling operations, the subgraphs API would crash because graphQL does not allow `skip` values larger than 5000. When (if) we hit that limit, the solution is to increase the `fromBlock`, and reset `skip` to 0. 
Note that duplicates are already handled in our history reducer, so we don't need to worry about that in the worker.

The solution was inspired in the [stake-operations-script repo](https://github.com/hemilabs/stake-operations-script) where I originally found this issue

### Screenshots

No visible changes to users

### Related issue(s)

Closes #1145

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/utils/sync-history/bitcoin.ts
+++ b/portal/utils/sync-history/bitcoin.ts
@@ -36,6 +36,7 @@ import { getBtcWithdrawals, getLastIndexedBlock } from 'utils/subgraph'
 import { type Address, decodeFunctionData, type Hash, toHex } from 'viem'
 
 import { chainConfiguration } from './chainConfiguration'
+import { calculateSkip } from './common'
 import { createSlidingTransactionList } from './slidingTransactionList'
 import { type HistorySyncer } from './types'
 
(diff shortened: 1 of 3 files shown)


ArturDolzan (approved):

gndelia:
I added [7d8f7e8](https://github.com/hemilabs/ui-monorepo/pull/1158/commits/7d8f7e8901063b64b176ef62abb40625229625b5) to apply one of Copilot's suggestions

PR #1160, opened 2025-04-23
Apply version tags to API Docker images

### LINKED ISSUE #1159: Use version tags for the API docker images
In order to support automated deployments that will be implemented soon, the three API docker images need to be tagged with versions. Current tags by branch/short-sha are not good for the automation scripts to identify new images and trigger the deploys.

gabmontes:
### Description

Read the version from `package.json` and set version tags to the built images.

### Related issue(s)

Closes #1159

### DIFF
--- a/.github/actions/docker-build-push/action.yml
+++ b/.github/actions/docker-build-push/action.yml
@@ -16,6 +16,9 @@ inputs:
   images:
     description: 'The Docker image name'
     required: true
+  version:
+    description: 'The version of the image'
+    required: true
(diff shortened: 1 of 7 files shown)


gndelia (approved):
LGTM - does that mean that on every change, the version of the package.json of these APIs needs to be updated?

gabmontes:
@gndelia yes. I will try to automate that in another PR.

ArturDolzan (approved):

PR #1168, opened 2025-04-25
Added sorting to stake tables

### LINKED ISSUE #981: Improve default sorting on Dashboard and Stake tables

alvarengathomas:
### Description

Added sorting to stake tables as follows:
- `sortTokens.ts` code and unit tests done by @qgatssdev on PR #1049;
- Dashboard table sorting based on the amount staked;
- Stake table sorting based on the token wallet balance;
- Fixed tables re-sorting after initial load;

### Related issue(s)

Closes #981

### Checklist

- [X] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/stake/_hooks/useWalletBalance.ts
+++ b/portal/app/[locale]/stake/_hooks/useWalletBalance.ts
@@ -1,0 +1,29 @@
+import { useQueries } from '@tanstack/react-query'
+import { useHemiClient } from 'hooks/useHemiClient'
+import { useStakeTokens } from 'hooks/useStakeTokens'
+import { StakeToken } from 'types/stake'
+import { useAccount } from 'wagmi'
+
+export const useWalletPositions = function () {
+  const { address: account } = useAccount()
+  const hemiClient = useHemiClient()
+  const stakeTokens = useStakeTokens()
+
+  return useQueries({
+    combine: results => ({
+      loading: results.some(({ isLoading }) => isLoading),
+      tokensWalletBalance: results
+        .filter(({ status }) => status === 'success')
+        .map(({ data }) => data),
+    }),
+    queries: stakeTokens.map(token => ({
+      queryFn: async () =>
+        hemiClient.getErc20TokenBalance({
+          account,
+          address: token.address as `0x${string}`,
+        }),
+      queryKey: ['wallet-token-balance', token.chainId, token.address],
+      select: (balance: bigint) => ({ ...token, balance }) satisfies StakeToken,
+    })),
+  })
+}
--- a/portal/app/[locale]/stake/_hooks/useWalletBalances.ts
+++ b/portal/app/[locale]/stake/_hooks/useWalletBalances.ts
@@ -2,6 +2,7 @@ import { useQueries } from '@tanstack/react-query'
 import { useHemiClient } from 'hooks/useHemiClient'
 import { useStakeTokens } from 'hooks/useStakeTokens'
 import { StakeToken } from 'types/stake'
+import { isNativeToken } from 'utils/nativeToken'
 import { useAccount } from 'wagmi'
 
 export const useWalletBalances = function () {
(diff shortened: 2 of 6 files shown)


gabmontes:
@alvarengathomas @gndelia should we close #1049 after this one?

portal/app/[locale]/stake/_hooks/useWalletBalance.ts
```diff
@@ -0,0 +1,29 @@
+import { useQueries } from '@tanstack/react-query'
+import { useHemiClient } from 'hooks/useHemiClient'
+import { useStakeTokens } from 'hooks/useStakeTokens'
+import { StakeToken } from 'types/stake'
+import { useAccount } from 'wagmi'
+
+export const useWalletPositions = function () {
```
Comment (gndelia) [inline portal/app/[locale]/stake/_hooks/useWalletBalance.ts:7]: I think the name of this file should match the hook name. So either this hook needs to be renamed to `useWalletBalance` , or the file to `useWalletPositions`.

I think I prefer `useWalletBalances` (note the plural) everywhere

Comment (alvarengathomas) [reply portal/app/[locale]/stake/_hooks/useWalletBalance.ts]: fixed

gndelia:
> I could not experience the problem of seeing the rows reordered after the initial load described by @gndelia on PR https://github.com/hemilabs/ui-monorepo/pull/1049 in my local environment, but I think we can address that in another issue/PR.


I think I've managed to reproduce this. On mainnet, you could hardcode the staked balance as `1` for all tokens. That is like if the user had staked 1 hemiBTC, 1 USD, 1 ETH, etc...
For that, you can update [here](https://github.com/hemilabs/ui-monorepo/blob/7049d9e752818b06de5c6f37fa778b154677b98a/portal/app/%5Blocale%5D/stake/_hooks/useStakedBalance.ts#L37) this function, to only return

```ts
// Will return 1000000 for btc, but 1000000000000000000 for ETH
return Promise.resolve(BigInt('1'.concat('0'.repeat(token.decimals))))
```

Once that happens, reordering does take place. In speedy internet, it is not noticeable, but adding throttling to Slow 4g, it becomes very visible

See 00:19

https://github.com/user-attachments/assets/86330b26-65e4-47c4-a86e-fcd6fc11bdd6


(Note: Of course, the whole page needs to be generally improved in loading times, but there are other tickets for that)

alvarengathomas:
> > I could not experience the problem of seeing the rows reordered after the initial load described by @gndelia on PR #1049 in my local environment, but I think we can address that in another issue/PR.
> 
> I think I've managed to reproduce this. On mainnet, you could hardcode the staked balance as `1` for all tokens. That is like if the user had staked 1 hemiBTC, 1 USD, 1 ETH, etc... For that, you can update [here](https://github.com/hemilabs/ui-monorepo/blob/7049d9e752818b06de5c6f37fa778b154677b98a/portal/app/%5Blocale%5D/stake/_hooks/useStakedBalance.ts#L37) this function, to only return
> 
> ```ts
> // Will return 1000000 for btc, but 1000000000000000000 for ETH
> return Promise.resolve(BigInt('1'.concat('0'.repeat(token.decimals))))
> ```
> 
> Once that happens, reordering does take place. In speedy internet, it is not noticeable, but adding throttling to Slow 4g, it becomes very visible
> 
> See 00:19
> 
>  Screen.Recording.2025-04-25.at.6.44.36.PM.mov 
> (Note: Of course, the whole page needs to be generally improved in loading times, but there are other tickets for that)
> 
> I think if we do not render the table until all the info is loaded (which should be achievable playing with `loading` booleans), this should be fixed. Bonus points for rendering loading cells (Like in the TX history)
> 
> Example in TX history

@gndelia I updated code to check for prices API loading flag, please test it again.

gndelia (approved):

ArturDolzan (approved):

PR #1169, opened 2025-04-25
Replace 'javascript-time-ago' with custom function

gndelia:
### Description

This PR intends to replace [javascript-time-ago](https://www.npmjs.com/package/javascript-time-ago) with a small custom implementation, loosely based on what Blockscout uses (except for the fact that Blockscout uses [dayjs](https://www.npmjs.com/package/dayjs)

As we just use this to list the "Time ago" in the TX history, I think it was best to remove this library which only add little to offer, and it was only used here. Instead, for formatting, I'm using [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) which is widely supported and offers the same functionality

### Screenshots

No visible changes to users

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/utils/format.ts
+++ b/portal/utils/format.ts
@@ -43,7 +43,7 @@ export const formatPastTime = function (
     minute: 59, // 59 seconds
     second: 1, // 1 second
   }
-  /* eslint-disable sort-keys */
+  /* eslint-enable sort-keys */
 
   const rtf = new Intl.RelativeTimeFormat(locale, { style: 'long' })
 
--- a/portal/utils/format.ts
+++ b/portal/utils/format.ts
@@ -23,6 +23,42 @@ export const formatNumber = (value: number | string) =>
 export const formatFiatNumber = (value: number | string) =>
   fiatRounder(value, { shouldFormat: true })
 
+export const formatPastTime = function (
+  secondsFromNow: number,
+  locale: string,
+) {
+  // We need to iterate from the largest unit to the smallest unit - so I prefer
+  // so sort them like that
+  /* eslint-disable sort-keys */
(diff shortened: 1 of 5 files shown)


portal/utils/format.ts
```diff
@@ -23,6 +23,42 @@ export const formatNumber = (value: number | string) =>
...
+    // See https://github.com/hemilabs/blockscout-frontend/blob/4f993ce5e62382ea98f42c6d79818630af331b9d/lib/date/dayjs.ts#L13
+    year: 28944000, // 11 months, assuming 365 days - 30 days
+    month: 2419200, // 4 weeks
+    week: 518400, // 4 days
+    day: 82800, // 23 hours
+    hour: 3540, // 59 minutes
+    minute: 59, // 59 seconds
+    second: 1, // 1 second
+  }
+  /* eslint-disable sort-keys */
```
ArturDolzan, line 46: Should it not be `/* eslint-enable sort-keys */` here?
```suggestion
  /* eslint-enable sort-keys */
```

ArturDolzan (commented):
Minor comment otherwise LGTM.

portal/utils/format.ts
gndelia: you're correct! Good catch. Fixed here  [2a48184](https://github.com/hemilabs/ui-monorepo/pull/1169/commits/2a481848db67ebe97a21dc3f21bb6df9fa1e66c2)

ArturDolzan (approved):

gabmontes (approved):
uACK

Issue #1195, opened 2025-05-05
Improve initial render after statically building


gndelia:
Related to #1031 #1065  and #1066 

Our page is a static page. This means when we build the page (using `npm run build`) before we deploy, Next.js generates a static version of our page.  While the app is large (and we're working on reducing the bundle size in the related issues), the requested page is initially... blank.

This is the preview on staging when requesting `/tunnel`

<img width="1500" alt="Image" src="https://github.com/user-attachments/assets/5b71ae85-6639-4f42-aad0-5e51ae508608" />

☝  This is what it gets rendered on the first request (which is `/tunnel.html`)

If we use the Performance tab to check how the page is rendered when the user access `/tunnel`, we see the following:

https://github.com/user-attachments/assets/89bffe46-6412-4188-b310-5ca400084823

(Note that the video was using throttling with `Fast 4g`)

There is a problem with our current app: While it is a static page, some things depend on the request context, so they can't be built statically. For example, the content may vary depending on whether the user is on Mainnet or Testnet (which is defined by the query string `networkType`)

In these scenarios, React renders everything up to the point it can, and then it leaves the rest to be rendered by the browser, using Client Side Rendering (CSR), at the moment of the user request.

We can also show a fallback while this is rendered, using `Suspense` - this generates a faster impression to the user.

If we're bailing out to client-side rendering whenever we depend on Testnet / Mainnet, with the current code, due to [this Suspense in the root layout](https://github.com/hemilabs/ui-monorepo/blob/09de2e13a20b71e89356a7af00ca464d9588f3cd/portal/app/%5Blocale%5D/layout.tsx#L75), it means the entire App is being rendered as client-side, while nothing is generated on the static build as a preview! This suspense is needed because some of the Context files in the layout use [query string to determine if the user is on mainnet / testnet](https://github.com/hemilabs/ui-monorepo/blob/09de2e13a20b71e89356a7af00ca464d9588f3cd/portal/hooks/useNetworkType.ts#L12), causing that React Node, down below the rest of the React Tree, to render it using CSR

This explains why in the video above, the page is fully blank and all of a sudden, it is "fully rendered".

Note, due to the Suspense location (needed because of how some Contexts are defined), this causes _all pages_ to be fully rendered

However, if we think about the different app components, there are sections that are rendered exactly the same, regardless of the user being on Mainnet or Testnet.

[...]

PR #1217, opened 2025-05-08
Static rendering improvements pt1

gndelia:
### Description

This PR applies some minor improvements to the initial render when statically building the site. Here, it's just moving around some files, I think the next PR will have more noticeable improvements when I split [useSyncHistory.ts](https://github.com/hemilabs/ui-monorepo/blob/main/portal/hooks/useSyncHistory/index.ts)

You can read the details of why these changes are needed, but the summary is that reading query string (in our case, to detect if the network is `mainnet` or `testnet`) "hangs" the rendering on build time, and lets the app to resolve the output on client side rendering time. Thus, some checks are causing that on build time, all we get is a static blank page.

It's easier to review by commit:

- e7998e3f4f4f203b29de2509703968b0363a9e75 Updates the `BtcWalletContext` not to depend on query string. We used to only enable the appropriate Bitcoin chain depending on query string. However, we can just enable both mainnet and testnet. This config is only to determine which chains are valid to connect to. Testnet wallets can be added and everything works as before - we already were doing this on [the EVM config](https://github.com/hemilabs/ui-monorepo/blob/05e239652946cf005b766f37222425eff8f4d2c5/portal/context/evmWalletContext.tsx#L42) and worked wonderfully, so we can get rid of this. By doing so, we can move `WalletsContext` out of the Suspense tree
- 802b12535f90bb461a6563252c7e94b563510d65 Moves the `ConnectWalletDrawerProvider` outside of the Suspense, as it does not require it, and it doesn't really depend on `TunnelHistoryProvider`, which is the trickiest one. So better to just move it out of the Suspense tree!
- a89d982bd9aa99527a26e7e31379bf7bf4d73a2e Moves the workers that watch deposits and withdrawals out of `portal/context/tunnelHistoryContext/index.tsx`. This way, they can be all moved outside of that file, and load them dynamically once. By load them with `ssr: false`, they no longer need Suspense.
- 3d8ee288d04fd519d212cc37349c734c1d90e855 Does the same as the previous commit, but for the Sync History workers. With all of that, the `portal/context/tunnelHistoryContext/index.tsx` only now contains the hook, next to be split in the next PR
- 8daf108d50101e2ef690f3e5ae5d83a09381cb07 Splits the `Analytics` part that depended on query string (the `GlobalTracking` file, now moved into a new component) so it can be wrapped around a Suspense (And not relying on the one in the root `layout`
- 6ddda50c40b51a5d666447c206047b3506d13675 After the previous commit, we can move `Analytics` out of the `Suspense` in the root `layout.tsx`

### Screenshots

No visible changes to users

### Related issue(s)

Related to #1195 

### Checklist

- [x] Manual testing passed.


portal/components/connectedWallet/connectedAccount.tsx
```diff
@@ -231,12 +231,8 @@ export const ConnectedBtcChain = function () {
     // As only one btc chain is supported at the moment, this will work.
     // Once there are multiple chains, we may need to show a dropdown or something
     // to select the chain to connect to.
-    return (
-      <WrongNetwork
-        onClick={() => switchChain({ chainId: chains[0].id })}
-        type="BTC"
-      />
-    )
+    const chainId = chains[0].id === chain.id ? chains[1].id : chains[0].id
```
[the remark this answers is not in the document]
gndelia: This is statically configured [here](https://github.com/hemilabs/ui-monorepo/blob/6ddda50c40b51a5d666447c206047b3506d13675/portal/context/btcWalletContext.tsx#L9) as part of this very same PR.

portal/components/tunnelStatusUpdaters/index.tsx
```diff
@@ -0,0 +1,17 @@
+import { BitcoinDepositsStatusUpdater } from './bitcoinDepositsStatusUpdater'
+import { BitcoinWithdrawalsStatusUpdater } from './bitcoinWithdrawalsStatusUpdater'
+import { EvmDepositsStatusUpdater } from './evmDepositsStatusUpdater'
+import { EvmWithdrawalsStateUpdater } from './evmWithdrawalsStateUpdater'
+
+export const TunnelStatusUpdaters = () => (
+  <>
+    {/* Track updates on bitcoin deposits, in bitcoin or in Hemi */}
+    {<BitcoinDepositsStatusUpdater />}
```
gabmontes, line 9: Extra curly braces?

portal/components/tunnelStatusUpdaters/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+    {/* Track updates on bitcoin withdrawals, from Hemi to Bitcoin */}
+    {<BitcoinWithdrawalsStatusUpdater />}
```
gabmontes, line 11: Same?

portal/components/tunnelStatusUpdaters/index.tsx
gndelia: 🤦🏽  copy-paste error. Removed in [c72c4c1](https://github.com/hemilabs/ui-monorepo/pull/1217/commits/c72c4c13bd098ea74f2d283c19dae97c43159a7d)

portal/components/tunnelStatusUpdaters/index.tsx
gndelia: See [here](https://github.com/hemilabs/ui-monorepo/pull/1217#discussion_r2080268611)

gabmontes (approved):

ArturDolzan (approved):

Issue #1229, opened 2025-05-13
Reduce the amount of Sentry errors


gndelia:
Besides logging individual issues for bugs for Sentry, this is like a "Catch All" for reducing / improving / filtering  errors on Sentry. We already had one that is closed, but I think there's still more work to do.

gabmontes:
Let's close this issue and open new ones for each specific problem we spot in Sentry.

PR #1230, opened 2025-05-13
Filter errors from Price and Points API

gndelia:
### Description

Both errors from Points and Prices API are handled in the frontend, so there is no more value in log these in Sentry. They are only consuming our quota.

`denyUrls` is added because we're also logging errors from Ajax requests.

### Related issue(s)

Related to #1229

### Checklist

- [x] Manual testing passed.


gabmontes (approved):

ArturDolzan (approved):
uACK

PR #1256, opened 2025-05-27
Preserve query params on locale change

### LINKED ISSUE #1249: Switching language does not preserve the chain if the user was in Testnet
### Description

If the user is in Testnet, when switching language, it gets redirected to Mainnet. See video

https://github.com/user-attachments/assets/50ad4e99-3ae2-406b-9d2f-72e2c2cdcdc0

### Steps to reproduce

Enter the portal, and switch to testnet. Then switch the language

Expected: The user should continue to be on Testnet
Actual: The user is now on mainnet

### Environment

- **OS**: _e.g., Windows, macOS, Linux_
- **Runtime/Browser/Version**: _e.g., Node 22, Chrome 120_
- **Wallet**: _e.g., MetaMask, Trust, Coinbase Wallet_

ArturDolzan:
### Description

This PR ensures that all existing query parameters (e.g. `networkType=testnet`) are preserved when switching between locales. Previously, changing the language via the locale selector would reset the query string, causing a fallback to `mainnet` due to the absence of the `networkType` parameter.

### Screenshots

https://github.com/user-attachments/assets/ebc44ea7-d8ce-424d-935f-20e514985831

### Related issue(s)

Closes #1249
Fixes #1249 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/_components/navbar/_components/help.tsx
+++ b/portal/app/[locale]/_components/navbar/_components/help.tsx
@@ -9,6 +9,7 @@ import { useOnClickOutside } from 'hooks/useOnClickOutside'
 import { usePathnameWithoutLocale } from 'hooks/usePathnameWithoutLocale'
 import { useRouter } from 'i18n/navigation'
 import { locales } from 'i18n/routing'
+import { useSearchParams } from 'next/navigation'
 import { Locale, useLocale, useTranslations } from 'next-intl'
 import { ComponentProps, ReactNode, RefObject, useState } from 'react'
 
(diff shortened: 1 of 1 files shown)


gabmontes (approved):

portal/app/[locale]/_components/navbar/_components/help.tsx
```diff
@@ -135,10 +136,14 @@ const ItemWithSubmenu = function ({
...
   const pathname = usePathnameWithoutLocale()
+  const searchParams = useSearchParams()
   const router = useRouter()
 
   const onClick = function (locale: Locale) {
-    router.push(pathname, { locale })
+    const query = searchParams.toString()
+    const fullPath = query ? `${pathname}?${query}` : pathname
+
+    router.push(fullPath, { locale })
```
Comment (gndelia) [inline portal/app/[locale]/_components/navbar/_components/help.tsx:146]: I know this was already like that, but I just noticed it was incorrect

```suggestion
    const query = searchParams.toString()
    const fullPath = query ? `${pathname}?${query}` : pathname

    router.push(fullPath)
```

`locale` is a path Parameter, not a query string (which is the second argument of `router.push`). It should be removed

gndelia (approved):

PR #1261, opened 2025-05-28
Input error improvements

### LINKED ISSUE #895: Prevent users to tunnel less than the minimum amount allowed by the token decimals

### LINKED ISSUE #1126: Show errors on tunnel
There are 2 types of errors we need to show:

- Input errors: For example, the user entered an invalid amount, something that is not an amount, or a different option. Note here that the code mostly prevents users from entering letters or other stuff, but we need to consider that users could find a way to bypass this. In addition to that, see https://github.com/hemilabs/ui-monorepo/issues/895
- Form errors unrelated to the user: For example, when depositing ERC20, we fail to query the blockchain to see if the allowance is enough; we can determine if an approval step is needed or not.

We need to tackle these 2 kind of scenarios

### LINKED ISSUE #1263: Adjust disabled state opacity for the main CTA in the tunnel
The disabled state of the main CTA currently has low opacity, making it less readable for users.

We need to update the opacity when disabled to: **0.5**

gndelia:
### Description

This PR improves the way we show errors both in the Tunnel and in Stake. Now errors are shown in the CTA button. In addition to that, the input is shown in red when there's an error in the input. I also made some tweaks as part of this PR. It may be better to review it by commit

- 20f6b0ee84757eaa67e3e2f945f83ceab9999601 This commit just moves the `tokenInput.tsx`  into a folder, so a new file can be added under `components/tokenInput/`
- 7cdea124c8d663fc4f7e66a44edb9347de81e27a Updates `connectEvmWallet` by moving it from a Portal component to the `components` folder, as this will be used in Stake too. In addition to that, it allows for customization of the text shown.
- 3adb47e114ca4c3081e3fb446dee5c726025daac Replaces usages of [viem's parseUnits](https://viem.sh/docs/utilities/parseUnits.html) with a custom function, `parseTokenUnits`. This is because `parseUnits` tends to round up (by design), but this causes problems (in the scenario of #895). For example, if a token has `decimals` 6, the minimum value that can be represented is `0.000001`. But `parseUnits` would round `0.0000009` to 0.000001` (note that the former has one extra zero than the latter), thus causing that in that border case, a user without sufficient balance could deposit. Now this function mimics what the contracts do: They truncate everything after the non-relevant decimals. So if the minimum amount is  `0.000001`, a user depositing  `0.123456789101112` will actually deposit `0.123456`. I 
- 52d30171a210eac474eefcd6d6c30ef832f54d75 Makes some improvements on how some components used for the CTA are loaded
- c92021c98e3322b53181a1fe045aa21c353d4215 Adds a generic function for input validation. Note that we had similar functions for Stake and the Tunnel, so these will be replaced by this unified new one.
- 3234074dbf1b996da1bc669093d64c464f134b93 Just like the errors, the "Minimum amount to operate" will also be shown in the CTA. So we no longer need to show the Minimum Input Message in gray (This was used to show the minimum bitcoin to operate with). This commit removes that component
- 72f60bc87031c76ef6fb2dfab53eb057e944af1a Updates the tunnel to use the new input validation function, and shows the errors on the CTA
- 66d4ee214bc3613ab6e61a5032babe6f86f2edbe Same as before, but for Stake
- f32d50bfa18497fc0f16b0787f14c85bc5c673e2 For the EVM Deposit, it shows an error if the allowance failed to load (something that's been pending for a long time)
- [a3d893c](https://github.com/hemilabs/ui-monorepo/pull/1261/commits/a3d893ce6a29d0c7f0c14e490666496b527d9263) Adjusts opacity for disabled buttons

### Screenshots

Some videos for the tunnel:

https://github.com/user-attachments/assets/814f1971-d0c3-4748-af84-ab605a9ca0fb

And for stake

https://github.com/user-attachments/assets/a0ac0da8-a507-4356-bf7f-b74c3b76ac72

### Related issue(s)

Closes #1263 
Closes #1126
Closes #895

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/tunnel/_components/evmDeposit.tsx
```diff
@@ -130,41 +68,47 @@ export const EvmDeposit = function ({ state }: EvmDepositProps) {
     updateFromInput,
   } = state
```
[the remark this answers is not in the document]
Comment (gndelia) [reply portal/app/[locale]/tunnel/_components/evmDeposit.tsx:70]: The comment is already inside `parseTokenUnits`. No need to add it here

portal/app/[locale]/tunnel/_components/submitEvmWithdrawal.tsx
```diff
@@ -0,0 +1,52 @@
+import { Button } from 'components/button'
```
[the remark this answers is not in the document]
Comment (gndelia) [reply portal/app/[locale]/tunnel/_components/submitEvmWithdrawal.tsx:3]: Applied in [fd5eca7](https://github.com/hemilabs/ui-monorepo/pull/1261/commits/fd5eca758a7c92ab6c94a08de0557e071eff07a6)

portal/app/[locale]/tunnel/_components/submitWithTwoWallets.tsx
```diff
@@ -50,8 +58,21 @@ export const SubmitWithTwoWallets = function ({ disabled, text }: Props) {
...
+          {validationError ? (
+            <Button disabled type="button">
+              {validationError}
+            </Button>
+          ) : (
+            <Button disabled={disabled} type="submit">
+              {text}
+            </Button>
+          )}
+        </>
```
Comment (ArturDolzan) [inline portal/app/[locale]/tunnel/_components/submitWithTwoWallets.tsx:74]: This extra Fragment can be removed
```suggestion
          {validationError ? (
            <Button disabled type="button">
              {validationError}
            </Button>
          ) : (
            <Button disabled={disabled} type="submit">
              {text}
            </Button>
          )}
```

gabmontes (dismissed):

portal/components/tokenInput/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+      isValid: false,
+    }
+  }
+  if (!isNativeToken(token) && amount > balance) {
+    return {
+      error: t('common.insufficient-balance', { symbol: token.symbol }),
+      errorKey: 'insufficient-balance',
+      isValid: false,
+    }
+  }
```
ArturDolzan: Just a suggestion, but you could simplify the native vs non-native token balance check by unifying both conditions into a single if..
```suggestion
  if (amount > balance || (isNativeToken(token) && amount === balance)) {
    return {
      error: t('common.insufficient-balance', { symbol: token.symbol }),
      errorKey: 'insufficient-balance',
      isValid: false,
    }
  }
```

ArturDolzan (commented):
I left some comments otherwise LGTM.

gabmontes (dismissed):

portal/components/tokenInput/utils.ts
gabmontes: That could be an oversimplification. I personally find it difficult to read even when it should work just perfectly fine.

gndelia: agree, I like this even more. I applied it in [eb376ae](https://github.com/hemilabs/ui-monorepo/pull/1261/commits/eb376aec8c7a9edb58b583b3507991518b615da8).

gndelia: 😆

ArturDolzan (approved):

gabmontes (approved):

PR #1264, opened 2025-05-28
Simplify test configuration and hooks

gndelia:
### Description

The `clearAllMocks` function can be set up at the configuration level, so we don't need to manually call it in every `beforeEach` when mocking deps in tests. This simplifies our suite a bit. See [docs](https://vitest.dev/config/#clearmocks).

### Screenshots

No visible changes to users

### Checklist

- [x] Manual testing passed.


gabmontes (approved):
🎉

ArturDolzan (approved):

PR #1269, opened 2025-05-29
Remove NetworkSwitch from header

### LINKED ISSUE #1231: Remove "Wrong Network" Toast
Currently, when a user is on the Tunnel section of the Portal and connected to the wrong network (e.g., Ethereum), clicking on the "From Network" selector to switch to Hemi triggers a toast saying the user is on the wrong network.

After discussing with @gndelia  and reviewing the approach used by Superbridge, we’ve decided to handle this differently. We'll implement logic within the Tunnel to automatically switch the user to the correct network based on the selected "From Network". This eliminates the need for the toast and makes the experience smoother.

Additionally, we will remove the network selector from the header, as it will no longer be necessary.

- Criteria:

1. Remove the "Wrong Network" toast from the Tunnel flow
2. Add automatic network switching based on the selected "From Network"
3. Remove the network selector from the header

Related to https://github.com/hemilabs/ui-monorepo/issues/1011 cc: @gabmontes

gndelia:
### Description

This PR removes the Network Switch shown in the Tunnel page from the header (and from the mobile footer).
It also removes the "Wrong Network" toast. We are only leaving the CTA "Connect to X" in the form, so it matches the Stake and the Drawers' behaviour.

We discussed once more with Nahuel about switching chains when toggling Deposit <-> Withdraw, but it is a bit clunky. Plus, we still need to handle the scenario where the user changes the chain manually from the wallet directly.

This way, it is still an improvement from what we had

### Screenshots

Goodbye Footer, and toast on mobile!

| Before  | Now |
| image  | image |

See how interactions are now

https://github.com/user-attachments/assets/8444c086-13ac-4f24-9189-eb444991a987

### Related issue(s)

Closes #1231 

### Checklist

- [x] Manual testing passed.


gabmontes (approved):

ArturDolzan (approved):
IMO the interactions are much better now 👏

Issue #721, opened 2024-12-27
Monitor vault configuration on Bitcoin


gndelia:
Bitcoin deposits and Withdrawals depend on the existence of vaults, and these should be properly configured, with deposited amounts, etc. There may be scenarios where deposits or withdrawals may not be able to be run. We could add some monitoring to these contracts (it shouldn't be necessarily in this repo) and detect errors if any

gabmontes:
The monitor should run continuously as a backend process/cron job and should even send alerts to Slack when something goes off ranges. That process could populate a small database with metrics that a simple API and web page could render for better "human readability".

PR #1322, opened 2025-06-18
Vaults monitor API and cron

gabmontes:
### Description

This PR creates the two backend components for the bitcoin vaults monitor: the API and the cron job.

A simple web page to quickly see the status of the tunnel and all the vaults will be added in another PR.

### Related issue(s)

Related to #721

### DIFF
--- a/vaults-monitor/README.md
+++ b/vaults-monitor/README.md
@@ -18,13 +18,13 @@ These control the `api` server:
 | CACHE_MINUTES | The time to cache the status in minutes.                        | 5                      |
 | CHAIN         | The Hemi chain to monitor: mainnet or testnet.                  | mainnet                |
 | ORIGINS       | Comma-separated list of allowed origins. '\*' is not supported. | `http://localhos:3000` |
-| PORT          | The HTTP port the server listens for requests.                  | 3003                   |
+| PORT          | The HTTP port the server listens for requests.                  | 3004                   |
 
 These environment variables control how the `cron` job behaves:
 
 | Variable             | Description                                                                         | Default                 |
 | -------------------- | ----------------------------------------------------------------------------------- | ----------------------- |
-| API_URL              | The URL of the API service.                                                         | `http://localhost:3003` |
+| API_URL              | The URL of the API service.                                                         | `http://localhost:3004` |
 | REFRESH_INTERVAL_SEC | How frequently the cache will be refreshed. If set to 0, it will run once and exit. | 300                     |
 | SLACK_MENTION        | The user to tag when sending alerts                                                 |                         |
 | SLACK_WEBHOOK_URL    | The full URL of the webhook to send the alerts to.                                  |                         |
@@ -42,6 +42,6 @@ Note: Environment variables can also be set in a `.env` file
 Querying the API is straightforward (response was edited):
 
 ```console
-$ curl http://localhost:3003
+$ curl http://localhost:3004
 {"bitcoinChainData":{"bitcoin":{"height":901676},"bitcoinKit":{"height":901674,"blockHash":"0x0000000000000000000107cb02a6c19084b54a1d0e8504d3a96d8fdd10d2f62b","version":536944640,"previousBlockHash":"0x00000000000000000000a30d9d5025c11b17b2c44b176de0e284c84f59dacebb","merkleRoot":"0x05c2f62905f83beeb75131a245c1851705814d09165fdac50b74e0e4814607b2","timestamp":1750196887,"bits":386021892,"nonce":1816880925}},"tunnelManagerData":{"withdrawalsPaused":false},"vaultsData":[{"status":5,"vaultAddress":"0x3DA10b74Bd339E69C1De9408020ce640B012e8cc","balanceSats":0,"bitcoinCustodyAddress":"18AVmm853HVhibPHMc3JRLXMynzKAbj6Po","pendingWithdrawalAmountSat":0,"pendingWithdrawalCount":0},...]}
 ```
--- a/vaults-monitor/api/server.js
+++ b/vaults-monitor/api/server.js
@@ -9,7 +9,7 @@ const { getBtcVaultsData } = require('./src/btc-vaults')
 const cacheMinutesStr = process.env.CACHE_MINUTES || '5'
 const chain = process.env.CHAIN || 'mainnet'
 const originsStr = process.env.ORIGINS || `http://localhost:3000`
-const portStr = process.env.PORT || '3003'
+const portStr = process.env.PORT || '3004'
 
 const chainId = chain === 'mainnet' ? hemi.id : hemiSepolia.id
 const maxAge = Number.parseInt(cacheMinutesStr) * 60 * 1000
@@ -25,3 +25,4 @@ app.get('/', async function (req, res) {
 })
 
 app.listen(Number.parseInt(portStr))
+console.log(`API server listening on port ${portStr}`)
--- a/vaults-monitor/api/src/btc-vaults.js
+++ b/vaults-monitor/api/src/btc-vaults.js
@@ -18,7 +18,7 @@ const viem = require('viem')
 
 function getHemiClient(chainId) {
   const chain = { [hemi.id]: hemi, [hemiSepolia.id]: hemiSepolia }[chainId]
-  const transport = viem.http(chain.rpcUrls.default.http[0])
+  const transport = viem.http()
   return viem.createPublicClient({ chain, transport })
 }
 
--- a/vaults-monitor/cron/src/check-vaults.js
+++ b/vaults-monitor/cron/src/check-vaults.js
@@ -1,8 +1,20 @@
 'use strict'
 
 const fetchJson = require('./fetch-json')
-const safeAsyncFn = require('./safe-async-fn')
 const postMessageToSlack = require('./post-message-to-slack')
+
+// These values match the Status enum used in the BitcoinTunnelManager and
+// BitcoinVault contracts.
+/* eslint-disable sort-keys */
+const VaultStatus = {
+  CREATED: 0,
+  INITIALIZING: 1,
+  LIVE: 2,
+  CLOSING_INIT: 3,
+  CLOSING_VERIF: 4,
+  CLOSED: 5,
+}
+/* eslint-enable sort-keys */
 
 const toBtc = sats => (sats / 100000000).toFixed(8)
 
@@ -25,17 +37,23 @@ function analyzeVaultsData({
   if (!vaultsData.length) {
     alerts.push('There are no vaults configured')
   }
-  if (!vaultsData.some(data => data.status === 2)) {
+  if (!vaultsData.some(data => data.status === VaultStatus.LIVE)) {
     alerts.push('No vaults are in LIVE state')
   }
   vaultsData.forEach(function (data, i) {
-    if (data.status === 0 || data.status === 1) {
-      // Vault is CREATED or INITIALIZING
+    if (
+      data.status === VaultStatus.CREATED ||
+      data.status === VaultStatus.INITIALIZING
+    ) {
       return
     }
 
-    if (data.status === 2 || data.status === 3) {
-      // Vault is LIVE or CLOSING_INIT
+    if (
+      data.status === VaultStatus.LIVE ||
+      data.status === VaultStatus.CLOSING_INIT
+    ) {
+      // This check shall be improved.
+      // See https://github.com/hemilabs/ui-monorepo/issues/1324
       if (data.balanceSats < data.pendingWithdrawalAmountSat) {
         alerts.push(
           `Vault ${i} at ${data.vaultAddress} does not have enough BTC in ${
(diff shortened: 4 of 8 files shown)


vaults-monitor/README.md
```diff
@@ -0,0 +1,47 @@
...
+### Configuration
+
+These control the `api` server:
+
+| Variable      | Description                                                     | Default                |
+| ------------- | --------------------------------------------------------------- | ---------------------- |
+| CACHE_MINUTES | The time to cache the status in minutes.                        | 5                      |
+| CHAIN         | The Hemi chain to monitor: mainnet or testnet.                  | mainnet                |
+| ORIGINS       | Comma-separated list of allowed origins. '\*' is not supported. | `http://localhos:3000` |
+| PORT          | The HTTP port the server listens for requests.                  | 3003                   |
```
gndelia, line 21: It's unlikely that both are running at the same time, but should we default to port 3004? 3003 is used by the [subgraph-api](https://github.com/hemilabs/ui-monorepo/blob/079326424dc51f418d4c34f8a5a924d029e61b49/subgraph-api/config/default.json#L4)

vaults-monitor/api/server.js
```diff
@@ -0,0 +1,27 @@
...
+const app = express()
+app.use(express.cors({ origin: originsStr.split(',') }))
+
+app.get('/', async function (req, res) {
+  const vaultsData = await limitedGetBtcVaultsData(chainId)
+  res.writeHead(200, { 'Content-Type': 'application/json' })
+  res.end(JSON.stringify(vaultsData))
+})
+
+app.listen(Number.parseInt(portStr))
```
gndelia, line 27: I would add a log info message after this
```suggestion
app.listen(Number.parseInt(portStr))

// eslint-disable-next-line no-console
console.info(`API server listening on port ${port}`)
```


it has happened to me a few times that I run the server locally, and as nothing happens (as there is no logs), I thought that something hang or was broken because I expected some output in the terminal 😅

vaults-monitor/api/src/btc-vaults.js
```diff
@@ -0,0 +1,119 @@
...
+  getVaultStatus,
+  hemi,
+  hemiSepolia,
+} = require('hemi-viem')
+const viem = require('viem')
+
+function getHemiClient(chainId) {
+  const chain = { [hemi.id]: hemi, [hemiSepolia.id]: hemiSepolia }[chainId]
+  const transport = viem.http(chain.rpcUrls.default.http[0])
+  return viem.createPublicClient({ chain, transport })
```
gndelia, line 22: This is the default value (See [here](https://github.com/wevm/viem/blob/4ccc559d243796387c671f69694f475fc5edc312/src/clients/transports/http.ts#L110)) so we can omit it

```suggestion
  const transport = viem.http()
  return viem.createPublicClient({ chain, transport })
```

vaults-monitor/cron/src/check-vaults.js
```diff
@@ -0,0 +1,74 @@
...
+      if (data.balanceSats < data.pendingWithdrawalAmountSat) {
+        alerts.push(
+          `Vault ${i} at ${data.vaultAddress} does not have enough BTC in ${
+            data.bitcoinCustodyAddress
+          } to complete ${
+            data.pendingWithdrawalCount
+          } pending withdrawals: (${toBtc(data.balanceSats)}) < (${toBtc(
+            data.pendingWithdrawalAmountSat,
+          )})`,
+        )
```
gndelia, line 48: This seems correct as an initial implementation. But for a future PR/issue, There are two collateralization thresholds that we may want to monitor. According to the contract comments:

```sol
[COLLATERALIZATION SYSTEM]
 * Each Vault is collateralized by its respective operator with a quantity of the approved ERC20
 * token. Vault operators can deposit additional collateral or withdraw excess collateral at any
 * time.
 *
 * The value of the collateral versus the custodied BTC is determined by an Oracle, which is an
 * implementation of the IAssetPriceOracle which is configured in the SimpleGlobalVaultConfig,
 * meaning that the same Oracle implementation (and thus, price feeds) are used by all
 * SimpleBitcoinVaults.
 *
 * There are two collateralization thresholds:
 *   1. The "soft collateralization threshold", below which no additional deposits can be accepted
 *      (and any attempted deposits will be rejected and the operator will be required to send
 *      the deposit back to the originator minus fees)
 *   2. The "hard collateralization threshold", below which the vault is subject to liquidation.
 *
 * The soft collateralization threshold is by default set to the maximum allowed value at vault
 * construction, which is provided by the SimpleGlobalVaultConfig and used by the
 * SimpleBitcoinVaultFactory. Operators can optionally choose to increase this threshold to protect
 * against liquidations.
 *
 * For example, if the minimum soft collateralization threshold is 140, then an operator could set
 * it to 150, and deposits would not be accepted when the deposit would increase the value of BTC
 * custodied by the vault such that the value of the collateral is less than 150% that of the BTC
 * custodied.
 *
 * The hard collateralization threshold is set to a single value provided by the
 * SimpleGlobalVaultConfig and cannot be changed by the operator.
 * 
 ```


I'm just pointing it out to keep it in mind in case you want to create an issue

File: vaults-monitor/api/src/btc-vaults.js [hunk 5]
```diff
@@ -0,0 +1,119 @@
...
+  const client = esploraClient({ network })
+  const height = await client.bitcoin.blocks.getBlocksTipHeight()
+  return { height }
+}
+
+async function getBitcoinChainData(client) {
+  const [bitcoin, bitcoinKit] = await Promise.all([
+    getBitcoinChainHeight(client.chain.testnet ? 'testnet' : 'mainnet'),
+    getBitcoinChainLastHeader(client),
+  ])
```
gndelia, line 35:
```suggestion
  const [bitcoinHeight, bitcoinKit] = await Promise.all([
    getBitcoinChainHeight(client.chain.testnet ? 'testnet' : 'mainnet'),
    getBitcoinChainLastHeader(client),
  ])
```

Should this be `bitcoinHeight`? With the current naming, it sounds like the chain definition for Bitcoin 🤔

gndelia: Nevermind, I understood better what you wanted to achieve when reading

```js
 bitcoinChainData.bitcoin.height - bitcoinChainData.bitcoinKit.height
```

vaults-monitor/cron/src/check-vaults.js
```diff
@@ -0,0 +1,74 @@
...
+  })
+  return alerts
+}
+
+function sendAlertsToSlack({ alerts, slackMention, slackWebhookUrl }) {
+  const intro =
+    'The following problems were found when analyzing the bitcoin vaults:'
+  const message = `${intro}\n\n${alerts.map(alert => `- ${alert}\n`)}`
+  if (slackWebhookUrl) {
+    postMessageToSlack(message, slackWebhookUrl, slackMention)
```
gndelia, line 60: In line 70, where `sendAlertsToSlack`  is called, there's an `await`.
`sendAlertsToSlack`  is not async, nor is it returning a Promise 🤔  perhaps a `return` is missing here?

```suggestion
    return  postMessageToSlack(message, slackWebhookUrl, slackMention)
```

Or the `await` below is not needed? This also goes in line with [this comment](https://github.com/hemilabs/ui-monorepo/pull/1322/files#r2155527915) about the response not being used

vaults-monitor/README.md
gabmontes: Oh! I thought 3003 was free. Changing to 3004...

vaults-monitor/api/package.json
```diff
@@ -0,0 +1,14 @@
...
+    "esplora-client": "1.2.0",
+    "hemi-viem": "2.5.0",
+    "not-express": "4.1.0",
+    "promise-mem": "1.0.4",
+    "viem": "2.31.3"
+  },
+  "engines": {
+    "node": ">=20"
+  }
+}
```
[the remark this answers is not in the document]
gabmontes: It is up there...

vaults-monitor/cron/src/post-message-to-slack.js
```diff
@@ -0,0 +1,24 @@
...
+  const messageText = { text: `${mention ? `<${mention}> ` : ''}${message}.` }
+  const res = await fetch(webHookUrl, {
+    body: JSON.stringify(messageText),
+    headers: { 'content-type': 'application/json' },
+    method: 'POST',
+  })
+  if (!res.ok) {
+    throw new Error(`Failed to post to Slack: ${res.statusText}`)
+  }
+  return res.text()
```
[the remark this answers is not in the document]
gabmontes: Slack responds `"ok"` in plain text. No use but I would like to make the function generic enough. Could be published as a package too. We used it in several places.

vaults-monitor/api/package.json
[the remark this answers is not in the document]
gndelia: I Swear I reviewed it many times and it wasn't there 😨  perhaps I should take a break... 😆

vaults-monitor/cron/src/check-vaults.js
```diff
@@ -0,0 +1,74 @@
+'use strict'
+
+const fetchJson = require('./fetch-json')
+const safeAsyncFn = require('./safe-async-fn')
+const postMessageToSlack = require('./post-message-to-slack')
+
+const toBtc = sats => (sats / 100000000).toFixed(8)
```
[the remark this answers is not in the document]
gabmontes: Given that the maximum amount of sats (21 million times 10^8) is lower than MAX_SAFE_INTEGER and that BigInts cannot be JSON stringified -a problem for the API- I switched every bitcoin/sats amount to Number.

vaults-monitor/cron/src/check-vaults.js
gabmontes: That's why I dislike `async/await` so much...

vaults-monitor/api/src/btc-vaults.js
```diff
@@ -0,0 +1,119 @@
...
+async function getBtcVaultsData(chainId) {
+  const client = getHemiClient(chainId)
+  const isTestnet = client.chain.testnet
+  const [bitcoinChainData, tunnelManagerData, vaultsData] = await Promise.all([
+    getBitcoinChainData(client),
+    // @ts-ignore ts(2345)
+    getTunnelManagerStatus(client),
+    getAllVaultsData(client),
+  ])
+  return { bitcoinChainData, isTestnet, tunnelManagerData, vaultsData }
```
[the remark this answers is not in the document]
gndelia: My point was to monitor that in order to remember to consolidate some utxos from time to time. Perhaps it's not relevant though. I also thought that I was using the `/address` elsewhere, but it seems the portal is not using it.

vaults-monitor/cron/src/check-vaults.js
gabmontes: No. I know that. The point is that `async/await` hides hides so many things happening in the background. Using a proimse chain, everything is explicit. You always return the next promise.

gndelia: Skill issue

⣿⣿⣿⣿⣿⣿⡿⠛⣛⣛⣛⣛⣛⣛⣛⣛⣛⣛⡛⠛⠿⠿⢿⣿⣿⣿⣿⣿⣿
⣿⣿⣿⣿⡿⢃⣴⣿⠿⣻⢽⣲⠿⠭⠭⣽⣿⣓⣛⣛⣓⣲⣶⣢⣍⠻⢿⣿⣿
⣿⣿⣿⡿⢁⣾⣿⣵⡫⣪⣷⠿⠿⢿⣷⣹⣿⣿⣿⢲⣾⣿⣾⡽⣿⣷⠈⣿⣿
⣿⣿⠟⠁⣚⣿⣿⠟⡟⠡⠀⠀⠀⠶⣌⠻⣿⣿⠿⠛⠉⠉⠉⢻⣿⣿⠧⡙⢿
⡿⢡⢲⠟⣡⡴⢤⣉⣛⠛⣋⣥⣿⣷⣦⣾⣿⣿⡆⢰⣾⣿⠿⠟⣛⡛⢪⣎⠈
⣧⢸⣸⠐⣛⡁⢦⣍⡛⠿⢿⣛⣿⡍⢩⠽⠿⣿⣿⡦⠉⠻⣷⣶⠇⢻⣟⠟⢀
⣿⣆⠣⢕⣿⣷⡈⠙⠓⠰⣶⣤⣍⠑⠘⠾⠿⠿⣉⣡⡾⠿⠗⡉⡀⠘⣶⢃⣾
⣿⣿⣷⡈⢿⣿⣿⣌⠳⢠⣄⣈⠉⠘⠿⠿⠆⠶⠶⠀⠶⠶⠸⠃⠁⠀⣿⢸⣿
⣿⣿⣿⣷⡌⢻⣿⣿⣧⣌⠻⢿⢃⣷⣶⣤⢀⣀⣀⢀⣀⠀⡀⠀⠀⢸⣿⢸⣿
⣿⣿⣿⣿⣿⣦⡙⠪⣟⠭⣳⢦⣬⣉⣛⠛⠘⠿⠇⠸⠋⠘⣁⣁⣴⣿⣿⢸⣿
⣿⣿⣿⣿⣿⣿⣿⣷⣦⣉⠒⠭⣖⣩⡟⠛⣻⣿⣿⣿⣿⣿⣟⣫⣾⢏⣿⠘⣿
⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣶⣤⣍⡛⠿⠿⣶⣶⣿⣿⣿⣿⣿⣾⣿⠟⣰⣿
⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣷⣶⣶⣤⣭⣍⣉⣛⣋⣭⣥⣾⣿⣿

gndelia (approved):

ArturDolzan (approved):

Issue #1318, opened 2025-06-17
Improve workers intervals and priorities


gndelia:
### Feature or change description

The different [workers](https://github.com/hemilabs/ui-monorepo/tree/main/portal/components/tunnelStatusUpdaters) have different intervals to poll for deposits/withdrawals. And each one, in the [worker definition](https://github.com/hemilabs/ui-monorepo/tree/main/portal/workers) uses its own priority system.

For example, on EVM withdrawals, the [priority](https://github.com/hemilabs/ui-monorepo/blob/0f9932f5bb5db74798e15c4893edf8d6c010347c/portal/workers/watchEvmWithdrawals.ts#L48) depends on which fields are missing, and the [interval](https://github.com/hemilabs/ui-monorepo/blob/0f9932f5bb5db74798e15c4893edf8d6c010347c/portal/components/tunnelStatusUpdaters/evmWithdrawalsStateUpdater.tsx#L23) on the status. However, if there is info that's missing, the interval should be probably smaller

This issue is to improve this system, with a better defined (and even tested) priority system and reduced intervals.

Furthermore, if an operation is opened in a drawer, we could prioritize that particular deposit/withdrawal

### Problem or use case addressed

The goal should be to allow loading the missing information faster, or prioritize those that are more recent or that the user is currently watching. We should also be able to unit-test the priority system

### Proposed implementation

TBD by the implementer

PR #1323, opened 2025-06-18
Added dynamic wait times for EVM withdrawals

### LINKED ISSUE #118: Check if we can remove hardcoded waiting time
When proving / claiming there are a few hardcoded period times

image

image

These were provided by @max-sanchez . However, the op-sdk has a few methods like [estimateMessageWaitTimeSeconds](https://sdk.optimism.io/classes/crosschainmessenger#estimateMessageWaitTimeSeconds) or [getChallengePeriodSeconds](https://sdk.optimism.io/classes/crosschainmessenger#getChallengePeriodSeconds) that may provide this information. We can check if these provide a more accurate estimate (we also need to ensure that they work ok on Hemi)

alvarengathomas:
### Description

Added dynamic wait times for EVM withdrawals

### Screenshots

When the transaction is pending or the request returns an error it shows the hardcoded wait times as before:

pending

---

When the transaction is confirmed it shows the exact waiting time for prove and claim:

<screenshot missing because my latest transaction got stuck in pending status, but I did a lot of tests before and it was working fine>

Please test it in your local environment.

--- 

When all the steps are done and there is no more wait times it show "Waiting completed":

ready-to-claim

### Related issue(s)

Closes #118

### Checklist

- [X] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx
+++ b/portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx
@@ -1,0 +1,106 @@
+import { useNetworkType } from 'hooks/useNetworkType'
+import { useTranslations } from 'next-intl'
+import Skeleton from 'react-loading-skeleton'
+import { MessageStatus, ToEvmWithdrawOperation } from 'types/tunnel'
+import { secondsToHours, secondsToMinutes } from 'utils/time'
+
+import {
+  useEvmWithdrawTimeToFinalize,
+  useEvmWithdrawTimeToProve,
+} from '../_hooks/useWaitTime'
+
+type Props = {
+  withdrawal: ToEvmWithdrawOperation
+}
+
+const ExpectedWithdrawalWaitTimeSecondsTestnet = 20 * 60
+const ExpectedWithdrawalWaitTimeSecondsMainnet = 40 * 60
+const ExpectedProofWaitTimeSecondsTestnet = 3 * 60 * 60
+const ExpectedProofWaitTimeSecondsMainnet = 24 * 60 * 60
+
+const renderTime = function (
+  timeInSeconds: number,
+  tCommon: ReturnType<typeof useTranslations>,
+) {
+  const minuteInSeconds = 60
+  const hourInSeconds = 60 * 60
+
+  if (timeInSeconds > hourInSeconds) {
+    return tCommon.rich('wait-hours', {
+      hours: () =>
+        tCommon('hours', {
+          hours: Math.ceil(secondsToHours(timeInSeconds)),
+        }),
+    })
+  }
+
+  if (timeInSeconds > minuteInSeconds) {
+    return tCommon('wait-minutes', {
+      minutes: Math.ceil(secondsToMinutes(timeInSeconds)),
+    })
+  }
+
+  return tCommon('wait-minutes', {
+    minutes: 1,
+  })
+}
+
+export const EvmWithdrawalWaitTimeToProve = function ({ withdrawal }: Props) {
+  const [networkType] = useNetworkType()
+  const tCommon = useTranslations('common')
+  const { data, isError, isPending } = useEvmWithdrawTimeToProve(withdrawal)
+
+  if (isPending) {
+    return <Skeleton className="w-12" />
+  }
+
+  if (isError) {
+    return (
+      <span>
+        {renderTime(
+          networkType === 'mainnet'
+            ? ExpectedWithdrawalWaitTimeSecondsMainnet
+            : ExpectedWithdrawalWaitTimeSecondsTestnet,
+          tCommon,
+        )}
+      </span>
+    )
+  }
+
+  if (data === 0) {
+    return <span> {tCommon('waiting-completed')} </span>
+  }
+
+  return <span> {renderTime(data, tCommon)} </span>
+}
+
+export const EvmWithdrawalWaitTimeToFinalize = function ({
+  withdrawal,
+}: Props) {
+  const [networkType] = useNetworkType()
+  const tCommon = useTranslations('common')
+  const { data, isError, isPending } = useEvmWithdrawTimeToFinalize(withdrawal)
+
+  if (isPending) {
+    return <Skeleton className="w-12" />
+  }
+
+  if (isError || withdrawal.status < MessageStatus.READY_FOR_RELAY) {
+    return (
+      <span>
+        {renderTime(
+          networkType === 'mainnet'
+            ? ExpectedProofWaitTimeSecondsMainnet
+            : ExpectedProofWaitTimeSecondsTestnet,
+          tCommon,
+        )}
+      </span>
+    )
+  }
+
+  if (data === 0) {
+    return <span> {tCommon('waiting-completed')} </span>
+  }
+
+  return <span> {renderTime(data, tCommon)} </span>
+}
--- a/portal/app/[locale]/tunnel/_hooks/useWaitTime.ts
+++ b/portal/app/[locale]/tunnel/_hooks/useWaitTime.ts
@@ -1,0 +1,22 @@
+import { useQuery } from '@tanstack/react-query'
+import { ToEvmWithdrawOperation } from 'types/tunnel'
+import {
+  getTimeToProveInSeconds,
+  getTimeToFinalizeInSeconds,
+} from 'utils/wait-times/evmWithdrawals'
+
+export const useEvmWithdrawTimeToProve = (withdrawal: ToEvmWithdrawOperation) =>
+  useQuery({
+    enabled: !!withdrawal,
+    queryFn: () => getTimeToProveInSeconds(withdrawal),
+    queryKey: ['time-to-prove-withdraw', withdrawal.transactionHash],
+  })
+
+export const useEvmWithdrawTimeToFinalize = (
+  withdrawal: ToEvmWithdrawOperation,
+) =>
+  useQuery({
+    enabled: !!withdrawal,
+    queryFn: () => getTimeToFinalizeInSeconds(withdrawal),
+    queryKey: ['time-to-finalize-withdraw', withdrawal.transactionHash],
+  })
--- a/portal/app/[locale]/tunnel/transaction-history/_components/withdrawStatus.tsx
+++ b/portal/app/[locale]/tunnel/transaction-history/_components/withdrawStatus.tsx
@@ -3,9 +3,12 @@ import Skeleton from 'react-loading-skeleton'
 import {
   BtcWithdrawStatus,
   MessageStatus,
+  ToEvmWithdrawOperation,
   WithdrawTunnelOperation,
 } from 'types/tunnel'
 import { isToEvmWithdraw } from 'utils/tunnel'
+
+import { EvmWithdrawalWaitTimeToProve } from '../../_components/evmWithdrawalWaitTime'
 
 import { TxStatus } from './txStatus'
 
@@ -15,17 +18,21 @@ type Props = {
 
 const EvmWithdrawStatus = function ({ withdrawal }: Props) {
   const t = useTranslations()
-  const waitMinutes = t('common.wait-minutes', { minutes: 20 })
+  const waitTime = (
+    <EvmWithdrawalWaitTimeToProve
+      withdrawal={withdrawal as ToEvmWithdrawOperation}
+    />
+  )
 
   const statuses = {
     // This status should never be rendered, but just to be defensive
     // let's render the next status:
     [MessageStatus.UNCONFIRMED_L1_TO_L2_MESSAGE]: (
-      <TxStatus.InStatus text={waitMinutes} />
+      <TxStatus.InStatus text={waitTime} />
     ),
     [MessageStatus.FAILED_L1_TO_L2_MESSAGE]: <TxStatus.Failed />,
     [MessageStatus.STATE_ROOT_NOT_PUBLISHED]: (
-      <TxStatus.InStatus text={waitMinutes} />
+      <TxStatus.InStatus text={waitTime} />
     ),
     [MessageStatus.READY_TO_PROVE]: (
       <TxStatus.InStatus text={t('transaction-history.ready-to-prove')} />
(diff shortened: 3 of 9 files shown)


portal/app/[locale]/tunnel/_hooks/useWaitTime.ts
```diff
@@ -0,0 +1,22 @@
...
+  })
+
+export const useEvmWithdrawTimeToFinalize = (
+  withdrawal: ToEvmWithdrawOperation,
+) =>
+  useQuery({
+    enabled: !!withdrawal,
+    queryFn: () => getTimeToFinalizeInSeconds(withdrawal),
+    queryKey: ['time-to-finalize-withdraw', withdrawal.transactionHash],
+  })
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_hooks/useWaitTime.ts:22]: I think we need a `refetchInterval` here for both. It's not needed to check every block, but perhaps every 2 blocks? (That would be 24 seconds, as each block is ~12 seconds)

portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx
```diff
@@ -0,0 +1,106 @@
...
+      </span>
+    )
+  }
+
+  if (data === 0) {
+    return <span> {tCommon('waiting-completed')} </span>
+  }
+
+  return <span> {renderTime(data, tCommon)} </span>
+}
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx:75]: (The following comment applies both to `EvmWithdrawalWaitTimeToFinalize` and `EvmWithdrawalWaitTimeToProve`)

I think `withdrawal.status` should be taken into consideration in these components to decide what to render. The withdrawal object you receive comes from [the TunnelContext](https://github.com/hemilabs/ui-monorepo/blob/079326424dc51f418d4c34f8a5a924d029e61b49/portal/context/tunnelHistoryContext/index.tsx#L102), available to the entire app. 
In the background, there are some web workers that are polling for every withdrawal, and those update the `status` field if they detect a change

Whenever a withdrawal is updated, that updated withdrawal is passed down the whole hierarchy, re-rendering the `reviewEvmWithdrawal.tsx`, which means it would re-render this component again here.  
Now, let's assume that the  `useEvmWithdrawTimeToProve` refetches, and now the waiting time to prove is 0.  This component would render "Waiting time completed". However, if the web worker did not update the `status` to `READY_TO_PROVE`, the drawer will render "Waiting time complete" but the "Prove on Ethereum" action won't be enabled. Note that polling for withdrawals may take [more than one minute](https://github.com/hemilabs/ui-monorepo/blob/0f9932f5bb5db74798e15c4893edf8d6c010347c/portal/components/tunnelStatusUpdaters/evmWithdrawalsStateUpdater.tsx#L23). (There's an issue to work on this https://github.com/hemilabs/ui-monorepo/issues/1318)

If there are only one or two withdrawals, it may not be noticeable as the workers can quickly update them, but if there are more than a few, polling may be delayed even more than what the intervals specify, and I think that this scenario becomes very plausible.


How would I solve it? Well, I'd only render "Waiting complete" if the status is "Ready to Prove". This way, the `status` is the mandatory field to mark as "completed". That's consistent with [how the status is shown as completed or not ready for each step](https://github.com/hemilabs/ui-monorepo/blob/079326424dc51f418d4c34f8a5a924d029e61b49/portal/app/%5Blocale%5D/tunnel/_components/reviewOperation/reviewEvmWithdrawal.tsx#L200). 

If the status does not show that the withdrawal is ready to prove, even though the `timeToProve` returns 0, I would show that there's a minimum waiting (like 1 minute or so). This way, we give time to the background worker to update the withdrawal status.  After all, if there's no more time to wait, it's highly unlikely that the withdrawal won't be updated in a few seconds/minutes.

IMHO that delay is acceptable as the user is not staring at the screen waiting (They're just doing stuff, they will be notified in the nav bar that the withdrawal has a new state), and they can wait 1 minute or 2 and have a consistent experience.  
I think with that (and after fixing #1318) this would be even less of a problem, as polling in the background should become faster.

portal/app/[locale]/tunnel/_hooks/useWaitTime.ts
Comment (alvarengathomas) [reply portal/app/[locale]/tunnel/_hooks/useWaitTime.ts]: added

portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx
Comment (alvarengathomas) [reply portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx:75]: added withdrawal status to both

portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx
```diff
@@ -0,0 +1,106 @@
...
+const ExpectedWithdrawalWaitTimeSecondsMainnet = 40 * 60 // 40 minutes
+const ExpectedProofWaitTimeSecondsTestnet = 3 * 60 * 60 // 3 hours
+const ExpectedProofWaitTimeSecondsMainnet = 24 * 60 * 60 // 24 hours
+
+const renderTime = function (
+  timeInSeconds: number,
+  tCommon: ReturnType<typeof useTranslations>,
+) {
+  const minuteInSeconds = 60
+  const hourInSeconds = 60 * 60
```
Comment (gabmontes) [inline portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx:26]: These constants could be defined outside the function.

portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx
```diff
@@ -0,0 +1,106 @@
...
+type Props = {
+  withdrawal: ToEvmWithdrawOperation
+}
+
+const ExpectedWithdrawalWaitTimeSecondsTestnet = 20 * 60 // 20 minutes
+const ExpectedWithdrawalWaitTimeSecondsMainnet = 40 * 60 // 40 minutes
+const ExpectedProofWaitTimeSecondsTestnet = 3 * 60 * 60 // 3 hours
+const ExpectedProofWaitTimeSecondsMainnet = 24 * 60 * 60 // 24 hours
+
+const renderTime = function (
```
Comment (gabmontes) [inline portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx:21]: It looks like this function could better be a component that renders the `<span>` with the text inside. It could also call `useTranslations` instead of receiving the `tCommon` as a parameter. In addition, to be reused, it could receive the type of time: withdrawal or proof.

I think it is abstracting some logic but the function/component could do better in that regard.

portal/messages/en.json
```diff
@@ -45,6 +45,7 @@
     "unsupported-chain-heading": "You're not connected to a supported chain",
     "wait-hours": "Wait ~<hours></hours>",
     "wait-minutes": "Wait ~{minutes} minutes",
```
gndelia, line 47: In `portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx` this is called with

```tsx
{tCommon('wait-minutes', {
  minutes: 1,
})}
```

That means it will render `Wait ~1 minutes`, when it should be `Wait ~1 minute` (notice the singular version)

We need to implement the plural / singular form  (see example on line 49). I think it should be something like this

```suggestion
    "wait-minutes": "Wait ~{minutes, plural, =1 {# minute} other {# minutes}}",
```

This is how it should look like for Spanish

```js
"wait-minutes": "Espere hasta ~{minutes, plural, =1 {# minuto} other {# minutos}}",
```

I'll let you the do Portuguese version

See [next-intl docs for plural usage](https://next-intl.dev/docs/usage/messages#cardinal-pluralization)

portal/app/[locale]/tunnel/_hooks/useWaitTime.ts
```diff
@@ -0,0 +1,24 @@
+import { useQuery } from '@tanstack/react-query'
+import { ToEvmWithdrawOperation } from 'types/tunnel'
+import {
+  getTimeToProveInSeconds,
+  getTimeToFinalizeInSeconds,
+} from 'utils/wait-times/evmWithdrawals'
+
+export const useEvmWithdrawTimeToProve = (withdrawal: ToEvmWithdrawOperation) =>
+  useQuery({
+    enabled: !!withdrawal,
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_hooks/useWaitTime.ts:10]: If the withdrawal is defined, this will query the `getTimeToProveInSeconds` despite the status of the withdrawal. But it's a call that's only meaningful in certain states. We could improve this (and `useEvmWithdrawTimeToFinalize`) by running the query when we actually need to. Something like

```suggestion
    enabled: !!withdrawal && [MessageStatus.STATE_ROOT_NOT_PUBLISHED, MessageStatus.READY_TO_PROVE].includes(withdrawal.status),
```

I think we could even get rid of `READY_TO_PROVE`, but let's be defensive and keep it.

Similar in line 17, but with `IN_CHALLENGE_PERIOD` and  `READY_FOR_RELAY` .


Note that `useQuery` will return `undefined` when disabled (or a cached value if any), so you may need to handle those scenarios.


With the current implementation, if a withdrawal is ready to claim, it is still calling the `getTimeToProveInSeconds`, which we know for a fact that won't be rendered, as at this stage, the wait should be completed (otherwise, it wouldn't be ready for claim!)

portal/app/[locale]/tunnel/transaction-history/_components/withdrawStatus.tsx
```diff
@@ -15,17 +18,21 @@ type Props = {
 
 const EvmWithdrawStatus = function ({ withdrawal }: Props) {
   const t = useTranslations()
-  const waitMinutes = t('common.wait-minutes', { minutes: 20 })
+  const waitTime = (
+    <EvmWithdrawalWaitTimeToProve
+      withdrawal={withdrawal as ToEvmWithdrawOperation}
+    />
+  )
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/tunnel/transaction-history/_components/withdrawStatus.tsx]: Improved here a5f6062aff29fc34eaa7b52d5ff18d544fe42909

portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+export const EvmWithdrawalWaitTimeToFinalize = function ({
+  withdrawal,
+}: WithdrawalProps) {
+  const [networkType] = useNetworkType()
+  const tCommon = useTranslations('common')
+  const { data, isError, isFetching, isPending } =
+    useEvmWithdrawTimeToFinalize(withdrawal)
+
+  if (isFetching && isPending) {
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx:102]: I'm not entirely sold on using `isFetching` here. That flag would be true on background refetches for revalidations. I think in those cases, I'd rather still show the previous data. What are your thoughts?

gabmontes (commented):
It looks like there is a problem with some types.

app/[locale]/tunnel/transaction-history/_components/table.tsx:364:27

portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/tunnel/_components/evmWithdrawalWaitTime.tsx]: Done here 6db5caad2c53a64c521e6b4ecd82ecfcf10a9fe3

gndelia:
I think there's something in the logic that needs to be improved. For example, see this Withdrawal that's ready to Prove. The `waiting completed` should appear straight away. We know that the status is `ready to prove`, so we don't need to show the skeleton, nor we care about the wait time. It should be `Waiting complete` straight away!  
The same happens with a "Ready to Claim" withdrawal


https://github.com/user-attachments/assets/c27254e3-dbbf-4a1d-b711-2aee2430b02b


For the first "Waiting X time", the loading should appear when the withdrawal is not yet ready to prove, after the withdrawal was initiated.
For the second "Waiting X time", the loading should appear after the withdrawal has been proven, but it is not ready to claim

ArturDolzan:
> I think there's something in the logic that needs to be improved. For example, see this Withdrawal that's ready to Prove. The `waiting completed` should appear straight away. We know that the status is `ready to prove`, so we don't need to show the skeleton, nor we care about the wait time. It should be `Waiting complete` straight away! The same happens with a "Ready to Claim" withdrawal
> 
>  Screen.Recording.2025-06-23.at.3.51.14.PM.mov 
> For the first "Waiting X time", the loading should appear when the withdrawal is not yet ready to prove, after the withdrawal was initiated. For the second "Waiting X time", the loading should appear after the withdrawal has been proven, but it is not ready to claim

Improvements implemented here f74f740449c202ec9c2676b9a4f7a39c06f02b0a. Please check it again.

portal/app/[locale]/tunnel/transaction-history/_components/withdrawStatus.tsx
```diff
@@ -3,29 +3,37 @@ import Skeleton from 'react-loading-skeleton'
...
 type Props = {
   withdrawal: WithdrawTunnelOperation
 }
 
-const EvmWithdrawStatus = function ({ withdrawal }: Props) {
+const EvmWithdrawStatus = function ({
+  withdrawal,
+}: {
+  withdrawal: ToEvmWithdrawOperation
+}) {
```
Comment (gabmontes) [inline portal/app/[locale]/tunnel/transaction-history/_components/withdrawStatus.tsx:24]: `Props` was replaced but is still defined in L16. Is that correct?

gndelia:
When starting a withdrawal on Sepolia (which hasn't confirmed yet)

image

I see the `wait ~1 minute`. This seems wrong, I think it should show the 20~minutes

After the TX is confirmed, I see the expected `wait ~20 minutes` (Correct)

image


And from there, the countdown goes as expected

On prove, while the tx is not confirmed, it works as expected

portal/app/[locale]/tunnel/transaction-history/_components/withdrawStatus.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/tunnel/transaction-history/_components/withdrawStatus.tsx:24]: Yeah it's still being used [here](https://github.com/hemilabs/ui-monorepo/blob/f74f740449c202ec9c2676b9a4f7a39c06f02b0a/portal/app/%5Blocale%5D/tunnel/transaction-history/_components/withdrawStatus.tsx#L103) on the exported component.

ArturDolzan:
> When starting a withdrawal on Sepolia (which hasn't confirmed yet)
> 
> image
> I see the `wait ~1 minute`. This seems wrong, I think it should show the 20~minutes
> 
> After the TX is confirmed, I see the expected `wait ~20 minutes` (Correct)
> 
> image
> And from there, the countdown goes as expected
> 
> On prove, while the tx is not confirmed, it works as expected

Good catch! Fixes done here 41e6dfe3ab1df6e42b5c29ebf6bb55c6f56d6741. Thanks!

Issue #1336, opened 2025-06-25
Add API to read TVL


gndelia:
### Feature or change description

Add an API that returns the TVL, and read it from there

### Problem or use case addressed

We're currently reading the TVL from Databox, but this may not be entirely reliable as we're overloading Databox with multiple calls that could be cached. In addition to that, Databox is not controlled by us

### Proposed implementation

Add a new API that returns the calculated TVL. Include an in-memory cache layer

PR #1337, opened 2025-06-25
Add a new cache for external APIs

gabmontes:
### Description

The Portal queries external APIs which can return errors or fail. An API cache service is introduced in this PR. Once deployed, a new PR will change i.e. the code at `portal/hooks/useTvl` to take advantage of the cache.

### Related issue(s)

Related to #1336

### DIFF
--- a/external-api-cache/README.md
+++ b/external-api-cache/README.md
@@ -6,17 +6,23 @@ This service is a middleman between the Portal and external APIs, caching the re
 
 These environment variables control how the cache works:
 
-| Variable           | Description                                                     | Default                |
-| ------------------ | --------------------------------------------------------------- | ---------------------- |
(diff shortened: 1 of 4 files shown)


ArturDolzan (dismissed):

package.json
```diff
@@ -26,6 +26,7 @@
   },
   "private": true,
   "workspaces": [
+    "external-api-cache",
```
gndelia, line 29: not for this PR, but I'm starting to wonder if we should create an `apis/*` folder or similar (some workspaces definitions group apis and app under `apps/*`, similar to `packages/*`

gndelia (approved):

package.json
gabmontes: Or consolidate all APIs in a single service. Let's keep thinking about this. The current state is not ideal at all.

PR #1338, opened 2025-06-25
Use API cache to get the TVL

### LINKED ISSUE #1336: Add API to read TVL

gabmontes:
### Description

With #1337 merged and the service deployed, this PR changes `useTvl` to take advantage of the cache.

### Related issue(s)

Closes #1336

### DIFF
--- a/portal/hooks/useTvl.ts
+++ b/portal/hooks/useTvl.ts
@@ -2,44 +2,14 @@ import { useQuery } from '@tanstack/react-query'
 import fetch from 'fetch-plus-plus'
 import { isValidUrl } from 'utils/url'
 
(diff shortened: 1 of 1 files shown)


ArturDolzan (approved):

gndelia (approved):
LGTM - but we should hold this until the `NEXT_PUBLIC_API_CACHE_URL` is set, right?

PR #1340, opened 2025-06-25
Improve EVM withdrawal polling and priority logics

ArturDolzan:
### Description

This PR extracts and unifies the logic for polling and priority based on:

- focusedWithdrawalHash, which gives max priority to withdrawals currently being viewed by the user (e.g., when the drawer is open).
- Missing withdrawal data (timestamp, status, prove/claim tx)
- Withdrawal message status
- Chain-specific fallback intervals

It also adds unit tests covering all combinations of priority and interval cases.

### Screenshots

No UI changes

### Related issue(s)

Related to #1318 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/test/utils/watch/pollings/analyzeEvmWithdrawalPolling.test.ts
+++ b/portal/test/utils/watch/pollings/analyzeEvmWithdrawalPolling.test.ts
@@ -1,0 +1,86 @@
+import {
+  MessageDirection,
+  MessageStatus,
+  type ToEvmWithdrawOperation,
+} from 'types/tunnel'
+import { hemiSepolia, sepolia } from 'viem/chains'
+import { describe, expect, it } from 'vitest'
+import {
+  analyzeEvmWithdrawalPolling,
+  EvmWithdrawalPriority,
+} from 'workers/pollings/analyzeEvmWithdrawalPolling'
+
+const getSeconds = (seconds: number) => seconds * 1000
+const getMinutes = (minutes: number) => getSeconds(minutes * 60)
+
+const createWithdrawal = (
+  overrides: Partial<ToEvmWithdrawOperation> = {},
+): ToEvmWithdrawOperation =>
+  // @ts-expect-error Only adding the minimum required properties for testing
+  ({
+    claimTxHash: '0x456',
+    direction: MessageDirection.L2_TO_L1,
+    l1ChainId: sepolia.id,
+    l2ChainId: hemiSepolia.id,
+    proveTxHash: '0x123',
+    status: MessageStatus.RELAYED,
+    timestamp: 123456,
+    transactionHash: '0xabc',
+    ...overrides,
+  })
+
+describe('analyzeEvmWithdrawalPolling', function () {
+  it('returns MAX priority and 6s interval for focused withdrawal', function () {
+    const withdrawal = createWithdrawal()
+    const result = analyzeEvmWithdrawalPolling({
+      focusedWithdrawalHash: '0xabc',
+      withdrawal,
+    })
+    expect(result.priority).toBe(EvmWithdrawalPriority.MAX)
+    expect(result.interval).toBe(getSeconds(6))
+  })
+
+  it('returns HIGH priority and 8s interval if timestamp is missing', function () {
+    const withdrawal = createWithdrawal({ timestamp: undefined })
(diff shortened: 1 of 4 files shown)


gabmontes (approved):

portal/workers/pollings/analyzeEvmWithdrawalPolling.ts
```diff
@@ -0,0 +1,94 @@
...
+      priority: EvmWithdrawalPriority.MEDIUM,
+    }
+  }
+
+  return {
+    interval:
+      refetchInterval[withdrawal.l2ChainId]?.[withdrawal.status] ?? fallback,
+    priority: EvmWithdrawalPriority.LOW,
+  }
+}
```
gndelia, line 94: Blocks are produced in nearly ~12 seconds, so I think no interval should be less than that, as the response will be the same

gndelia (changes requested):

portal/test/utils/watch/pollings/analyzeEvmWithdrawalPolling.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    claimTxHash: '0x456',
+    direction: MessageDirection.L2_TO_L1,
+    l1ChainId: sepolia.id,
+    l2ChainId: hemiSepolia.id,
+    proveTxHash: '0x123',
+    status: MessageStatus.RELAYED,
+    timestamp: 123456,
+    transactionHash: '0xabc',
+    ...overrides,
+  })
```
gndelia, line 30: For tests, I like to use `zeroHash` and `zeroAddress` from `viem` (Assuming that the fact that they are the same value do not affect the test)

```suggestion
    claimTxHash: zeroHash,
    direction: MessageDirection.L2_TO_L1,
    l1ChainId: sepolia.id,
    l2ChainId: hemiSepolia.id,
    proveTxHash: zeroHash,
    status: MessageStatus.RELAYED,
    timestamp: 123456,
    transactionHash: zeroHash,
    ...overrides,
  })
```

portal/workers/pollings/analyzeEvmWithdrawalPolling.ts
```diff
@@ -0,0 +1,94 @@
...
+    [MessageStatus.READY_TO_PROVE, MessageStatus.READY_FOR_RELAY].includes(
+      // @ts-expect-error status is of type MessageStatus
+      withdrawal.status,
+    )
+  ) {
+    return {
+      interval: getSeconds(10),
+      priority: EvmWithdrawalPriority.MEDIUM,
+    }
+  }
```
gndelia, line 87: I know this was in the original implementation - I would just remove them

```suggestion
```

This causes that if there are 2 (due to queue concurrency) withdrawals in either of those states, they block status updates on the other withdrawals.

I'd just leave all the status updates at the same priority level (except the `status` undefined which takes higher priority)

portal/workers/watchEvmWithdrawals.ts
```diff
@@ -3,16 +3,16 @@ import PQueue from 'p-queue'
...
-import { isWithdrawalMissingInformation } from 'utils/tunnel'
 import { type EnableWorkersDebug } from 'utils/typeUtilities'
 import { hasKeys } from 'utils/utilities'
 import { watchEvmWithdrawal } from 'utils/watch/evmWithdrawals'
 import { typeWorker } from 'utils/workers'
 import { type Hash } from 'viem'
 
+import { analyzeEvmWithdrawalPolling } from './pollings/analyzeEvmWithdrawalPolling'
+
 const queue = new PQueue({ concurrency: 2 })
```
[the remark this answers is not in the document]
ArturDolzan: Absolutely. Changed here 0276ecd397cf403a93720ddf2ec5b60b4e2349cc

portal/test/utils/watch/pollings/analyzeEvmWithdrawalPolling.test.ts
ArturDolzan: TIL! Improved here 0276ecd397cf403a93720ddf2ec5b60b4e2349cc. Thanks!

gabmontes (approved):

gndelia (approved):

PR #1374, opened 2025-07-07
Allow the token prices job to terminate

gabmontes:
### Description

For the token prices cron to run as a cron (refreshes the prices and gracefully exits), the connection to Redis must be closed after the task is done. Otherwise, the open connection will prevent Node to exit.

### Related issue(s)

Related to #1360

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal-backend/cron/token-prices/src/refresh-prices.js
+++ b/portal-backend/cron/token-prices/src/refresh-prices.js
@@ -26,18 +26,20 @@ async function fetchPrices() {
 }
 
 const client = redis.createClient(config.get('redis'))
-client.connect()
 
 const expiration = config.get('cacheExpirationMin') * 60
 
-const storePrices = prices =>
-  Promise.all(
+async function storePrices(prices) {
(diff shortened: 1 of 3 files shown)


ArturDolzan (approved):

portal-backend/cron/token-prices/src/refresh-prices.js
```diff
@@ -26,18 +26,20 @@ async function fetchPrices() {
...
+  client.connect()
+  await Promise.all(
     Object.entries(prices)
       .map(([symbol, price]) =>
         client.set(`price:${symbol}`, price, { EX: expiration }),
       )
       .concat(client.set('time', Date.now())),
   )
+  client.quit() // Release the connection to allow the process to exit
+}
```
gndelia, line 42: If there's an error, would the connection still be open?
Perhaps you can use a try / finally and close the connection there.

ArturDolzan (approved):

PR #1401, opened 2025-07-22
Split stake toast from toast component

gndelia:
### Description

We will need the toast component _somewhere else_ (_wink wink_), and we had a toast component used exclusively for the Stake page. This PR splits that component into 2: A generic component that accepts the props, and updates the Stake component to use it.

In addition to this, I updated the toast to be loaded dynamically as it won't be shown after user interaction.

### Screenshots

Other than the lazy loading, there are no visible changes to users

### Related issue(s)

Related to _mystery issue_ (?)

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/stake/_components/stakeToast.tsx
+++ b/portal/app/[locale]/stake/_components/stakeToast.tsx
@@ -18,26 +18,33 @@ export const StakeToast = function ({ chainId, txHash, type }: Props) {
 
   const t = useTranslations('stake-page.toast')
 
+  const tx = {
+    href: `${blockExplorer.url}/tx/${txHash}`,
+    label: formatEvmHash(txHash),
+  }
+
+  if (type === 'stake') {
+    return (
+      <Toast
+        description={t('here-your-stake-tx', {
+          type: t('staking'),
+        })}
+        goTo={{
+          href: '/stake/dashboard',
+          label: t('go-staking-dashboard'),
+        }}
+        title={t('staking-successful')}
+        tx={tx}
+      />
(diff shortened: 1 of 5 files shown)


ArturDolzan (approved):

portal/app/[locale]/stake/_components/stakeToast.tsx
```diff
@@ -1,89 +1,43 @@
...
-      <button className="h-5 w-5" onClick={() => setClosedToast(true)}>
-        <CloseIcon className="h-full w-full [&>path]:hover:stroke-white" />
-      </button>
-    </div>
+    <Toast
+      description={t('here-your-stake-tx', {
+        type: t(type === 'stake' ? 'staking' : 'unstaking'),
+      })}
+      goTo={
+        type === 'stake'
```
Comment (gabmontes) [inline portal/app/[locale]/stake/_components/stakeToast.tsx:27]: Sometimes I prefer to do the check at a higher level, once, and i.e. render two different Toast components depending on that check. That's usually more readable than having a single component and check in the props that may change.

gabmontes (approved):

portal/app/[locale]/stake/_components/stakeToast.tsx
Comment (gndelia) [reply portal/app/[locale]/stake/_components/stakeToast.tsx]: Implemented in [41f4ba8](https://github.com/hemilabs/ui-monorepo/pull/1401/commits/41f4ba89e2996d103c7f83ddfdc20aae0b4c8586)

ArturDolzan (approved):

gabmontes (approved):

Issue #1334, opened 2025-06-24
Iterate the portal header left and right padding


dvnahuel:
We need to update the left and right padding of the portal header for better alignment with the overall layout.

Current aligment: 

<img width="174" alt="Image" src="https://github.com/user-attachments/assets/290e28e1-c4a6-4b1f-97b0-2f9e85dc99cf" />
<img width="222" alt="Image" src="https://github.com/user-attachments/assets/55c7db9d-13dd-4687-9799-130e4b16dcd3" />

Proposal:

Iterate from 24px to 12px (left and right padding)

PR #1407, opened 2025-07-25
Apply new button style guide

### LINKED ISSUE #1376: Implement new CTA design system.
The buttons are defined by three main heights: 28px, 32px, and 44px, and three main styles: primary, secondary, and tertiary. You can check all states for each style in [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=13427-14401&t=K66ZqETOKz5CO13b-1).

Also note: **when a CTA has only an icon and no label, the container width should match the height.**

### LINKED ISSUE #1389: Places where to upgrade the CTA's on the portal.
**Tunnel:**

- The connect wallet CTA should be: **Secondary XS**

Image

- The main tunnel CTA should be: **Primary XL**

Image

- TX table CTA's should be: **Primary XS** or **Secondary XS** (depending the state of the item)

Image
Image

- TX table _drawer_ CTA's should be: **Primary S** or **Secondary S** (depending the state of the item)

- TX table _reload_ CTA should be: **Primary XS**

Image

- Wallet drawer, when the Unisat wallet is not connected should be: **Tertiarty XS**

Image

**Left main navbar:**

- The '?' should be: **Secondary XS**

Image

**Get started page:**

- Learn more in the _tutorials_ section CTA should be: **Primary XS**

Image

ArturDolzan:
### Description

This PR implements the new button style guide, adjusting variants, sizes, and paddings.

- Updated Button and ButtonLink components to follow the new style guide.
- Introduced small, xSmall, and xLarge size options with proper spacing and typography.
- Simplified the internal logic for computing classes and added a utility to merge Tailwind classes with conflict resolution.
- Adjusted button usage across the app, especially inside drawers, where smaller buttons (xSmall, small) are now applied for better visual balance.

### Screenshots

https://github.com/user-attachments/assets/59947b5c-6676-4f14-b120-491148925c78

### Related issue(s)

Closes #1376
Closes #1389
Related to #1334  

### Checklist

- [x] Manual testing passed.


gabmontes (changes requested):
@ArturDolzan conflicts!

portal/utils/simpleMergeTailwindClasses.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Simple function to merge multiple Tailwind CSS class lists into a single string, deduplicating classes
+ * based on specified property prefixes. If multiple classes share the same prefix (e.g., 'px-', 'py-'),
+ * only the last occurrence is kept in the result.
+ *
+ * @param classLists - An array of class list strings (may include undefined or falsy values).
+ * @param deduplicationPrefixes - An optional array of prefixes to use for deduplication (defaults to ['px-', 'py-']).
+ * @returns A single string containing the merged and deduplicated Tailwind CSS classes.
+ */
+export function simpleMergeTailwindClasses({
```
gabmontes, line 23: How is this related to [tailwind-merge](https://www.npmjs.com/package/tailwind-merge)? Should we use that library instead, as is a more complete merging solution?

portal/components/button.tsx
```diff
@@ -1,65 +1,80 @@
...
-    focus:shadow-button-primary-focused`,
-  secondary: `text-neutral-950 bg-white border-neutral-300/55 hover:bg-neutral-100
-    disabled:bg-neutral-100 shadow-button-secondary focus:shadow-button-secondary-focused`,
+  primary: `primary-button border-button-primary bg-button-primary text-white 
+  transition-all duration-200 hover:bg-button-primary-hovered 
+  disabled:bg-button-primary-disabled disabled:shadow-button-primary-disabled shadow-button-primary 
+  focus:shadow-button-primary-focused border border-solid`,
+  secondary: `text-neutral-950 bg-white hover:bg-neutral-50 disabled:bg-neutral-50 
+  shadow-button-secondary focus:shadow-button-secondary-focused disabled:shadow-button-secondary-disabled`,
+  tertiary: `text-neutral-950 bg-transparent hover:bg-neutral-100 focus:bg-neutral-100 focus:shadow-button-tertiary-focused`,
```
gabmontes, line 18: You may want to play with the indentation here to make this more readable. Or just use plain single quotes. Same for `commonCss` above.

portal/utils/simpleMergeTailwindClasses.ts
gndelia: I know it's not _exactly_ the same, but in general, for these components that have a very defined set of options (Variants), I'd prefer avoiding exposing the `className` to the caller, because if we would, the consumer could potentially override any option... I think we shouldn't allow that, as it does not favor the consistency of the components across the app. If an option becomes so important, maybe it needs to be part of the props.

dvnahuel:
Can we add a transition so the hovering interactions feels more smooth? let's try something like `transition: all 0.3s ease;
`

portal/utils/simpleMergeTailwindClasses.ts
ArturDolzan: @gndelia I agree that avoiding className exposure helps enforce consistency.. there’s just one edge case still open: we have some buttons (actually just two) that render only a single icon, like  the reload tx history, with no label or additional content. For those, the padding is different (px-1.5 -> 6px).

In this case, removing the className exposure, I was thinking of introducing a prop like isIconOnly to handle that scenario explicitly. Alternatively, I tried to infer it internally (e.g., checking if children is a single valid React element without children), but it’s not very reliable..

Do you think isIconOnly as a prop is acceptable in this case? Or do you have a better idea on how we could detect and handle this padding variation without breaking the style contract?

dvnahuel:
CTA's focus state should trigger only on keyboard focus, not mouse click.

portal/components/button.tsx
```diff
@@ -1,65 +1,80 @@
...
 }
 
-type ButtonProps = ComponentProps<'button'> & Variant & FontSize & Height
+export type ButtonSize = 'xSmall' | 'small' | 'xLarge'
+
+const sizeMap: Record<ButtonSize, string> = {
+  small: 'h-8 text-sm gap-x-2 px-3',
+  xLarge: 'h-11 text-mid gap-x-2 px-4',
+  xSmall: 'h-7 text-xs gap-x-1 px-2.5',
+} as const
```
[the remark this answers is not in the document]
ArturDolzan: Well, it was discussed with @dvnahuel and he mentioned that it was not needed to create, for example, the Large one for now, but it could be done in the future.. so to follow his definitions [here](https://github.com/hemilabs/ui-monorepo/issues/1389), I used basically the same nomenclature. 

Please let me know if you guys prefer another nomenclature.. I'm good with both.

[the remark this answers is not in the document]
gndelia: no, it's ok, but I got curious 😅

dvnahuel:
XS style CTA's have should have 6px border radius.

portal/utils/simpleMergeTailwindClasses.ts
gndelia: yeah, I understand what you're saying. For that reason I made (and exists). the `ButtonLink` component, where it looks like a button, but it's actually not a button (it's a link!)

Maybe you can use a `ButtonIcon` component, and it does something like this

```tsx
const (children, buttonProps) => (

<div className="[&>button]:px-1.5>
  <Button {...buttonProps}>{children}</Button>
/>
```

it's not that far from `isIconOnly`, that's correct, but it keeps the `Button` API simpler.

_actually_, I think that rather than setting `px-.1.5`, you need to set `px-0` (Removing it) and make the height and width match.. So basically I think this component would use its own set of sizeMaps:



```tsx

const (children, size, ...buttonProps) => (

<div className="[&>button]:px-1.5>
  <Button {...buttonProps}>{children}</Button>
/>
```

dvnahuel:
Reload CTA here should be styles as 'secondary'
Screenshot 2025-07-25 at 12 25 31 PM

dvnahuel:
The menu CTA on mobile and tablet views should also be styled as secondary.
Screenshot 2025-07-25 at 12 29 06 PM

gndelia:
> Reload CTA here should be styles as 'secondary' Screenshot 2025-07-25 at 12 25 31 PM

The issue says the other way around! https://github.com/hemilabs/ui-monorepo/issues/1389

image

dvnahuel:
Something we need to keep in mind is the following: CTAs with icons should use the same colors as shown in Figma.
For example, in the case of the 'Connect Wallets' CTA, the wallet icons should be set (in the default state) to grayscale/500, and on hover, to grayscale/950.
Screenshot 2025-07-25 at 12 51 29 PM

portal/components/button.tsx
ArturDolzan: Improved here 5370e454620d9b73d863e7bb6179f0e9401e9069

portal/components/button.tsx
```diff
@@ -1,65 +1,80 @@
...
 import { ComponentProps } from 'react'
+import { simpleMergeTailwindClasses } from 'utils/simpleMergeTailwindClasses'
 import { isRelativeUrl } from 'utils/url'
 
 import { ExternalLink } from './externalLink'
 
-const commonCss = `box-content flex items-center justify-center
-  rounded-lg border border-solid px-3 py-1.5 font-medium disabled:opacity-50`
+const commonCss = `box-border flex items-center justify-center
+  rounded-lg py-1.5 font-medium disabled:opacity-50`
```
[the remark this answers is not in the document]
ArturDolzan: Fixed here 5370e454620d9b73d863e7bb6179f0e9401e9069

ArturDolzan:
@dvnahuel this commit 5370e454620d9b73d863e7bb6179f0e9401e9069 addresses several feedback points related to CTA styles and behavior:

- Focus state: CTA focus styles now appear only when navigating via keyboard (focus-visible), not on mouse click.
- Primary button shadow: Adjusted shadows to better match the design in Figma.
- XS CTAs: Ensured xSmall buttons have a 6px (rounded-md) border radius.
- Reload CTA: Updated to use the secondary style, both in the corresponding screen and mobile menu.
- Oops screen CTA: Now uses the small primary style as expected.
- Icon colors: Updated icons inside buttons (e.g. Connect Wallet) to match the expected grayscale values: text-neutral-500 by default, text-neutral-950 on hover.

portal/components/connectWallets/wallets.tsx
```diff
@@ -58,8 +58,8 @@ const ConnectWalletButton = function ({
       {icon}
       <span className="text-base font-medium text-neutral-950">{text}</span>
       {rightIcon || (
-        <div className="ml-auto">
-          <Chevron.Right />
+        <div className="group-hover ml-auto">
+          <Chevron.Right className="group-hover:[&>path]:fill-neutral-950" />
```
gndelia, line 62: I think this is incorrect. Check out the [group-hover](https://v3.tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state) docs

```suggestion
        <div className="group ml-auto">
          <Chevron.Right className="group-hover:[&>path]:fill-neutral-950" />
```

The parent uses `group` and the children uses `group-hover`. They can also be named, like 


```suggestion
        <div className="group/container ml-auto">
          <Chevron.Right className="group-hover/container:[&>path]:fill-neutral-950" />
```

but apparently that's not mandatory

dvnahuel:
The primary CTA shadows are off
Screenshot 2025-07-29 at 10 45 16 AM

gndelia:
@dvnahuel  a minor question, but

image


Should we update the image with the "Deploy" button, so the button matches our new design? To be fair, I don't think it's even matching our previous design entirely.

(Note this could be done in a separate PR / issue)

dvnahuel:
The Connect Wallet icon should use neutral/500 by default, and change to neutral/950 on hover.
Screenshot 2025-07-29 at 10 56 10 AM

portal/components/connectWallets/wallets.tsx
ArturDolzan: Fixed here 3d67e7a7c9bf96bd276d8319e7c91fe8ccd79f89. Thanks!

portal/app/[locale]/tunnel/transaction-history/_components/reloadHistory.tsx
```diff
@@ -29,12 +30,13 @@ export const ReloadHistory = function () {
...
+    <ButtonIcon
       onClick={onResync}
+      size="xSmall"
       type="button"
+      variant="secondary"
     >
-      <ReloadIcon className="[&>path]:group-hover/reload-history:fill-neutral-950" />
-    </button>
+      <ReloadIcon />
+    </ButtonIcon>
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/tunnel/transaction-history/_components/reloadHistory.tsx]: Fixed here 3d67e7a7c9bf96bd276d8319e7c91fe8ccd79f89. Thanks!

dvnahuel:
The 'hamburger' CTA should have the same interacion as the secondary CTA (on-hover the background-color and icon color changes to natural-950)
Screenshot 2025-07-29 at 1 49 17 PM

dvnahuel:
> > @dvnahuel a minor question, but
> > image
> > Should we update the image with the "Deploy" button, so the button matches our new design? To be fair, I don't think it's even matching our previous design entirely.
> > (Note this could be done in a separate PR / issue)
> 
> Makes sense, let's put it as a new issue

Issue [here](https://github.com/hemilabs/ui-monorepo/issues/1416)

gndelia (dismissed):
Code-wise, it LGTM. I'll defer to Nahuel to approve the whole UX side (btw, it looks fantastic!)

portal/components/button.tsx
```diff
@@ -4,51 +4,99 @@ import { isRelativeUrl } from 'utils/url'
...
+    icon: 'h-7 text-xs rounded-md min-w-7',
+    regular: 'h-7 text-xs gap-x-1 px-2.5 rounded-md',
+  },
+  small: {
+    icon: 'h-8 text-sm rounded-lg min-w-8',
+    regular: 'h-8 text-sm gap-x-2 px-3 rounded-lg',
+  },
+  xLarge: {
+    icon: 'h-11 text-mid rounded-lg min-w-11',
+    regular: 'h-11 text-mid gap-x-2 px-4 rounded-lg',
```
gabmontes, line 56: Why are we jumping from "small" right to "xLarge"?

ArturDolzan: @gabmontes the naming wasn’t based on `small`, `medium`, and `large` because during the initial discussion with @dvnahuel , we aligned on only implementing a specific size for now, and following the naming from his reference to keep it consistent. That said, I’m totally open to renaming them if using a more generic convention like small, medium, large would make things clearer.

Let me know what you prefer..

dvnahuel: This is how I’ve been structuring the designs, mainly because, in the current portal, there isn’t really a place where a ‘large (40px)’ CTA fits naturally. The x-large (44px) button is mostly used in the tunnel form and was almost specifically created for that. Everywhere else across the portal we’re mostly using xSmall or small button sizes.

gabmontes: Got it. Thanks for the explanation.

gabmontes (approved):

gndelia (approved):

PR #1422, opened 2025-07-31
fix: Buttonlink hover effect

UttamShetty19:
Added missing hover background effect to ButtonLink so its behavior matches the Button component.
Fixes inconsistent UI on elements like the “Learn more” button.


ArturDolzan (changes requested):

ArturDolzan (approved):
LGTM! Thanks for your contribution!

PR #1424, opened 2025-08-01
Replace Inter Display with Inter Variable

### LINKED ISSUE #1384: Replace Inter Display with Inter Variable in 'Get Started' texts
Update the font in the following sections from Inter Display to Inter Variable for consistency.

Image
Image

arielmergen:
### Description

This commit applies Inter Variable to h3 and h4; other headings will remain in Inter Display.

### Screenshots

Captura de pantalla 2025-08-01 a la(s) 4 58 01 p  m

Captura de pantalla 2025-08-01 a la(s) 4 57 44 p  m

### Related issue(s)

Closes #1384

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/portal/styles/globals.css
+++ b/portal/styles/globals.css
@@ -36,10 +36,13 @@ like "next/router" or "next/navigation" inside the error pages.
   @apply font-inter-variable;
 }
 
+h3,
+h4 {
+  @apply font-inter-variable;
+}
+
 h1,
 h2,
-h3,
-h4,
 h5,
 h6 {
   @apply font-inter-display;


gndelia (commented):
This will update _all h3 and h4 headings_ in the portal.


I think @dvnahuel only wanted to update those 2 particular cases. I'll let him confirm. 

Assuming he says yes, if that's the case, for the "Add your network to your wallet", we could update it so it uses a `p` instead of a heading (or even a `span`). 
It's not clear that one is a heading, so this change feels safe.

As of the other one, IMHO _it should_ be a heading. If Nahuel wants it to be rendered with a different font, then we can directly apply the font style there

portal/styles/globals.css
```diff
@@ -36,10 +36,13 @@ like "next/router" or "next/navigation" inside the error pages.
   @apply font-inter-variable;
 }
 
+h3,
+h4 {
+  @apply font-inter-variable;
+}
+
```
gabmontes, line 43: Let's move "h3,h4" below "h1,h2,h5,h6" for better readability.

gabmontes (commented):
Is this change *not* generating any undesired side-effect. Correct?

arielmergen:
> This will update _all h3 and h4 headings_ in the portal.
> 
> I think @dvnahuel only wanted to update those 2 particular cases. I'll let him confirm.
> 
> Assuming he says yes, if that's the case, for the "Add your network to your wallet", we could update it so it uses a `p` instead of a heading (or even a `span`). It's not clear that one is a heading, so this change feels safe.
> 
> As of the other one, IMHO _it should_ be a heading. If Nahuel wants it to be rendered with a different font, then we can directly apply the font style there

Hi @gndelia, I am sending this pr to confirm what you mentioned. Thanks for the suggestions.

@gabmontes, Side effect: some headings that required Inter Display will display Inter Variable.

dvnahuel:
> > Side effect: some headings that required Inter Display will display Inter Variable
> 
> @arielmergen is this what we want? @dvnahuel ?

Inter Display should only be used for titles. In the screenshots you shared, the 'add network to your wallet' should actually be using Inter Variable instead. And let's keep Inter Display for the other example (the h3 one)

portal/styles/globals.css
arielmergen: I've reverted these changes

gabmontes (approved):

PR #1471, opened 2025-09-03
Add increaseAmount to ve-hemi-actions

### LINKED ISSUE #1470: Add increaseAmount to ve-hemi-actions

We need to support the increaseAmount in the `ve-hemi-actions` package. This should let users add more HEMI to an existing lock without creating a new one. The flow is pretty similar to createLock, just reusing the approval handling and pointing to the increaseAmount function on the contract.

ArturDolzan:
### Description

This PR adds the `increaseAmount` flow so users can top up their existing lock positions. While doing that I also refactored the approval logic into a shared handleApproval helper, since both createLock and increaseAmount needed the same checks and events. It also adds some unit tests.

### Screenshots

No UI changes.

### Related issue(s)

Closes #1470 

### Checklist

- [x] Manual testing passed.


packages/ve-hemi-actions/actions/wallet/veHemi.ts
```diff
@@ -75,6 +83,139 @@ const canCreateLock = async function ({
...
+          address: hemiTokenAddress,
+        }),
+    )
+
+    if (additionalAmount > tokenBalance) {
+      return { canIncrease: false, reason: 'insufficient balance' }
+    }
+
+    return { canIncrease: true }
+  } catch {
```
gndelia, line 137: I think here we're not validating that the Position isn't expired already - see [this check](https://github.com/hemilabs/veHEMI/blob/04054f2eae6508e2b563547ea58f038faa0ef505/src/VeHemi.sol#L201)

gndelia (changes requested):

packages/ve-hemi-actions/actions/wallet/veHemi.ts
```diff
@@ -75,6 +83,139 @@ const canCreateLock = async function ({
...
+  veHemiAddress,
+  walletClient,
+}: {
+  account: Address
+  amount: bigint
+  approvalAmount?: bigint
+  walletClient: WalletClient
+  veHemiAddress: Address
+  hemiTokenAddress: Address
+  emitter: EventEmitter<ApprovalEvents>
```
gndelia, line 157: let's sort this alphabetically (I'm surprised the linter did not catch this)

packages/ve-hemi-actions/utils.ts
```diff
@@ -50,9 +79,32 @@ export const validateCreateLockInputs = function ({
...
+  chainId: number
+  tokenId: bigint
+}) {
+  const commonError = validateCommonInputs({
+    account,
+    amount: additionalAmount,
+    approvalAmount: approvalAdditionalAmount,
+    chainId,
+  })
+  if (commonError) return commonError
```
gabmontes, line 104: In `eslint-config-bloq` we had a rule enabled to prevent `if () return` in one line and always force using curly brackets. We should keep doing so in TS code.

packages/ve-hemi-actions/actions/wallet/veHemi.ts
```diff
@@ -110,60 +251,18 @@ const runCreateLock = ({
...
-      // @ts-ignore because it works on IDE, and when building on its own, but fails when compiling from the portal through next
-      const allowance = await getErc20TokenAllowance(walletClient, {
-        address: hemiTokenAddress,
-        owner: account,
-        spender: veHemiAddress,
+      const approved = await handleApproval({
+        account,
+        amount,
+        approvalAmount,
+        emitter: emitter as EventEmitter<ApprovalEvents>,
```
gndelia, line 258: In order to avoid this casting, I think we can do something similar to [this](https://github.com/hemilabs/ui-monorepo/blob/137abbcd0e1fb864338a7222bd492d47ec2f138d/packages/hemi-tunnel-actions/src/utils.ts#L78). Check the usages of `handleWaitDeposit` and see how it automatically infers the correct type of `EventEmitter<T>`

packages/ve-hemi-actions/actions/wallet/veHemi.ts
ArturDolzan: You are correct. Added the check here 23656193192048190d45f29d034931e3b5576a8d. Thanks!

packages/ve-hemi-actions/actions/wallet/veHemi.ts
ArturDolzan: Nice! applied here 23656193192048190d45f29d034931e3b5576a8d and the it infers automatically. Thanks!

gndelia: soo... if it infers automatically, can we remove the casting?

The goal should be to be able to just do

```suggestion
        emitter: emitter,
```

packages/ve-hemi-actions/actions/public/veHemi.ts
```diff
@@ -23,3 +23,24 @@ export const getHemiTokenAddress = async function (
     functionName: 'HEMI',
   })
 }
+
+export const getLockedBalance = async function (
+  client: PublicClient | WalletClient,
```
gndelia, line 28: Does this work?

```suggestion
  client: Client,
```

gndelia (approved):

PR #1519, opened 2025-09-17
Fix token selector modal UI

### LINKED ISSUE #1420: There's an extra scrollbar in the token selector
### Description

When opening the token selector, if the user has tokens with balance, there's an extra horizontal scrollbar

See video

https://github.com/user-attachments/assets/39df1663-e3d1-4061-884e-3a8e64313fa8

Notice that this doesn't happen when the user is disconnected, so it must be related with the section for user's balances

### Steps to reproduce

1. Connect your wallet using an address that contain balances for a few tokens (testnet should be fine)
2. Open the token selector
3. As you scroll, you notice an horizontall scrollbar that dissapears when the "Your tokens" section is no longer visible

### Proposed fix

That scrollbar should not be there

gndelia:
### Description

This PR fixes a small horizontal overflow in the token selector UI

### Screenshots

| Before  | Now |
| image | image |

Notice how the horizontal scrollbar is now gone

### Related issue(s)

Closes #1420

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/components/tokenSelector/list.tsx
+++ b/portal/components/tokenSelector/list.tsx
@@ -85,7 +85,7 @@ function AllTokensHeader({
       }}
     >
       {hasTokens && (
-        <div className="absolute inset-x-0 top-4 h-px border-t border-neutral-300/55 md:-mx-6" />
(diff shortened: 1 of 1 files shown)


gabmontes (approved):
🎉

PR #1555, opened 2025-10-01
Ignore cross-spawn

gabmontes:
### Description

The Docker image scans report `cross-spawn@7.0.3` has a vulnerability but it cannot be solved as `@graphprotocol/graph-cli@0.97.1` has its dependencies and sub-dependencies fixed and the maintainers have been ignoring for months the [reports about this](https://github.com/graphprotocol/graph-tooling/issues/2043).

The alternative for us is to ignore the alert and move on as this is not a problem for our specific use case.

Note: There is no need to update the versions and create new images. Nothing relevant changed.

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -1,1 +1,3 @@
+# @graphprotocol/graph-cli@0.97.1 > gluegun@5.2.0 > cross-spawn@7.0.3
+# https://github.com/graphprotocol/graph-tooling/issues/2043
 CVE-2024-21538


arielmergen (dismissed):

ArturDolzan (approved):

.trivyignore
```diff
@@ -0,0 +1 @@
+CVE-2024-21538
```
gndelia, line 1: Can we link the issue here?

```suggestion
# See https://github.com/graphprotocol/graph-tooling/issues/2043
CVE-2024-21538
```

So at least whenever we find this file, we have an easy way to see if it's been fixed (I think that's how comments are added)

gndelia (approved):

arielmergen (approved):

Issue #1505, opened 2025-09-15
Implement new designs for staking governance


ArturDolzan:
### Feature or change description

[Design link](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=15008-14287)

PR #1562, opened 2025-10-03
Add management features

ArturDolzan:
### Description

This PR implements new features that let users manage their existing staking positions. Users can now increase their staked amount or extend the lockup time without having to create a new position. It built drawer flows for both operations with step-by-step progress tracking so you can follow along as transactions happen. There's also a handy action menu on each table row to quickly access these options. While working on this, I extracted some components to keep things organized and reusable.

**Note**: The Add time to his lockup interface will be improved by @dvnahuel  following today's usability discussion we had after the daily. Current implementation, including text labels, is work in progress.

### Screenshots

https://github.com/user-attachments/assets/28d322c2-79cd-4b3f-86dc-f76c2f8f6e48

Mobile
https://github.com/user-attachments/assets/abe47d47-845b-4b56-aebf-94eeda7f5220

### Related issue(s)

Related to #1505

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/staking-dashboard/_hooks/useIncreaseAmount.ts
```diff
@@ -0,0 +1,174 @@
...
+            ),
+        )
+
+        // fees
+        updateNativeBalanceAfterFees(receipt)
+        // HEMI balance
+        queryClient.setQueryData(
+          hemiBalanceQueryKey,
+          (old: bigint) => old + amount,
+        )
```
Comment (gndelia) [inline portal/app/[locale]/staking-dashboard/_hooks/useIncreaseAmount.ts:142]: I think here is the other way around. You're adding $HEMI to your locked position, thus reducing your HEMI balance in your wallet

```suggestion
        queryClient.setQueryData(
          hemiBalanceQueryKey,
          (old: bigint) => old - amount,
        )
```

gndelia (changes requested):

portal/types/stakingDashboard.ts
```diff
@@ -49,6 +43,18 @@ export type StakingPosition = {
...
+  approvalTxHash?: Hash
+  input?: string
+  inputDays?: string
+  lockupDays?: number
+  transactionHash?: Hash
+  stakingPosition?: Partial<
+    Pick<StakingPosition, 'amount' | 'tokenId' | 'lockTime'>
+  >
+  status?: StakingDashboardStatus
+}
```
gndelia, line 56: A minor suggestion, but if you expect everything to be optional (particularly, if you expect it when adding properties in the future), you can use `Partial`
```suggestion
export type StakingDashboardOperation = Partial<{
  approvalTxHash: Hash
  input: string
  inputDays: string
  lockupDays: number
  transactionHash: Hash
  stakingPosition: Partial<
    Pick<StakingPosition, 'amount' | 'tokenId' | 'lockTime'>
  >
  status: StakingDashboardStatus
}>
```

Is every field optional for a reason?

portal/app/[locale]/staking-dashboard/_hooks/useStakingDashboardState.ts
```diff
@@ -40,59 +40,74 @@ type Actions =
   | UpdateStakingDashboardOperation
   | UpdateUnlockingDashboardOperation
 
-// the _:never is used to fail compilation if a case is missing
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const compilationError = function (_: never): never {
-  throw new Error('Missing implementation of action in reducer')
+const compilationError = function (value: string): never {
+  throw new Error(`Missing implementation of action in reducer: ${value}`)
```
Comment (gndelia) [inline portal/app/[locale]/staking-dashboard/_hooks/useStakingDashboardState.ts:44]: This will only give a compilation error if using a `switch`, as we do in [the tunnel state](https://github.com/hemilabs/ui-monorepo/blob/0b6ff07e9c792810c02a874bb40ef11cac31db53/portal/app/%5Blocale%5D/tunnel/_hooks/useTunnelState.ts#L139C14-L139C30)

For example:

Try creating an action 

```ts
type TestOperation = Action<'testOperation'> & NoPayload
```

and add it at the end of `type Actions`

`compilationError` does not give a compilation error. So if you go for using an object instead of the `switch`, you can remove this function. After all, you'll get the compilation error in line 56

```
// Now you get a TS error as you're missing to add the action handler here
const actionHandlers: ActionHandlers = {
```

portal/app/[locale]/staking-dashboard/_hooks/useIncreaseAmount.ts
```diff
@@ -0,0 +1,174 @@
...
+
+        queryClient.setQueryData(
+          stakingPositionQueryKey,
+          (old: StakingPosition[] | undefined = []) =>
+            old.map(position =>
+              position.tokenId === tokenId
+                ? { ...position, amount: position.amount + amount }
+                : position,
+            ),
+        )
```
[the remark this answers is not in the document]
Comment (gndelia) [reply portal/app/[locale]/staking-dashboard/_hooks/useIncreaseAmount.ts:134]: I did not test it locally. Perhaps you fixed it without noticing after recording the video? 😅

portal/app/[locale]/staking-dashboard/_hooks/useStakingDashboardState.ts
Comment (ArturDolzan) [reply portal/app/[locale]/staking-dashboard/_hooks/useStakingDashboardState.ts]: Applied here bd7f889c80d24bdcda8d86f6f9bbfd25e8c90062. Thanks!

portal/app/[locale]/staking-dashboard/_components/stakeReview/increaseAmount/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+    isError:
+      isIncreaseAmountFeesError ||
+      (needsApproval && isApprovalTokenGasFeesError),
+    label: t('common.network-gas-fee', { network: hemi.name }),
+    token: getNativeToken(hemi.id),
+  })
+
+  const getTotalIncreaseAmount = () =>
+    getTotal({
+      fees: increaseAmountFees,
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/staking-dashboard/_components/stakeReview/increaseAmount/index.tsx]: Indeed. Applied here bd7f889c80d24bdcda8d86f6f9bbfd25e8c90062. Thanks!

portal/app/[locale]/staking-dashboard/_hooks/useIncreaseAmount.ts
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/staking-dashboard/_hooks/useIncreaseAmount.ts:134]: It could be. I did some small changes after recording those videos. Please let me know if you can reproduce it later.

gabmontes (commented):
The veHEMI contract calculates the new time lock as follows:

```solidity
        uint256 _unlockTime = ((block.timestamp + lockDuration_) / SIX_DAYS) * SIX_DAYS; // unlock time is rounded down to SIX_DAYS
        if (_unlockTime > block.timestamp + MAX_TIME) revert LockDurationTooLong();
        if (_unlockTime <= _oldLocked.end) revert NewLockDurationNotGreater();
```

So the new lock duration sent to [`increaseUnlockTime`](https://github.com/hemilabs/veHEMI/blob/d72d12017a088cbb13b57a2f3f5fc09b1063e853/src/VeHemi.sol#L212) is added to the current time (the current block timestamp), normalized and then checked. If the new time is more than 4 years in the future, it fails. If it is less than the current unlock date, it fails too.

I didn't see in the code where these two conditions are checked. The first is just done by preventing a new lock duration larger than 4 years. But the second is more tricky.

gndelia (dismissed):
uACK

portal/types/stakingDashboard.ts
```diff
@@ -49,6 +43,18 @@ export type StakingPosition = {
...
+export type StakingDashboardOperation = Partial<{
+  approvalTxHash: Hash
+  input: string
+  inputDays: string
+  lockupDays: number
+  transactionHash: Hash
+  stakingPosition: Partial<
+    Pick<StakingPosition, 'amount' | 'tokenId' | 'lockTime' | 'timestamp'>
+  >
+  status?: StakingDashboardStatus
```
gndelia, line 55: A minor comment - This can be removed, as there's already a `Partial`
```suggestion
  status: StakingDashboardStatus
```

gndelia (changes requested):

ArturDolzan:
> Even though I know the UX will be improved, I think we could fix this:
> 
> image
> The dropdown should have a shadow - it should be `shadow-lg` (cc @dvnahuel )

It used to have a shadow (we can se it on videos).. It was introduced after rebasing I guess.. will check

gndelia:
@ArturDolzan  yes - probably after https://github.com/hemilabs/ui-monorepo/pull/1564 😅

portal/app/[locale]/staking-dashboard/_hooks/useIncreaseUnlockTime.ts
```diff
@@ -0,0 +1,145 @@
...
+            (old: StakingPosition[] | undefined = []) =>
+              old.map(position =>
+                position.tokenId === tokenId
+                  ? {
+                      ...position,
+                      lockTime: daysToSeconds(BigInt(lockupDays - step)),
+                    }
+                  : position,
+              ),
+          )
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/staking-dashboard/_hooks/useIncreaseUnlockTime.ts:120]: You're right, there was a bug. Fixed here f206ef5a1eea95942b02143b5805f233513d7671 by calculating the new unlock time the same way the contract does (current time + duration, rounded to 6-day increments), then subtract the original start timestamp. Now the optimistic update should matches what shows up after the subgraph syncs. Thanks for pointing this out!

gndelia (approved):
uACK

Issue #1536, opened 2025-09-22
Improve tunnel form component to match the design system


dvnahuel:

gndelia:
For item 1 I think we should first implement #1525 and #1528

Issue #1526, opened 2025-09-18
Componentize shadow styles across the portal


dvnahuel:
Currently, we have many manually applied shadows across different components in the portal, which isn’t ideal. This issue aims to standardize shadows by:

1. Creating variables for each shadow group (`bs`, `sm`, `md`, `lg`, `xl`)
2. Applying those new shadow variables to every component in the portal that uses shadows

Shadows variables found [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=15213-10991&t=0GMrbU5acUzQ2tJ2-1)

dvnahuel:
These should be all the places where we need to add the new shadows variables.

**1 - Tunnel**

<img width="183" height="119" alt="Image" src="https://github.com/user-attachments/assets/9fca75e6-368e-4c8c-a30e-1414f4a3d145" />
<img width="517" height="61" alt="Image" src="https://github.com/user-attachments/assets/78605fba-0acf-4d85-8454-a66aeb4cfaad" />
<img width="501" height="58" alt="Image" src="https://github.com/user-attachments/assets/e2f37f38-ca8a-4a19-8937-614f0a579e42" />

All component shared above should be with the `sm` shadow variable.



<img width="158" height="108" alt="Image" src="https://github.com/user-attachments/assets/9d3b3b92-ddf8-4294-8254-6eb44e0df1f1" />

All dropdowns menus should have the `lg` shadow variable.



<img width="472" height="584" alt="Image" src="https://github.com/user-attachments/assets/56642007-a615-4e87-ad90-bc3d1c0061c6" />
<img width="155" height="117" alt="Image" src="https://github.com/user-attachments/assets/a0263d0c-a5d7-4f3e-a1e1-f63ca46cf2ef" />

The token selection modal should have the **xl** shadow variable
The ETH-USDC-USDT cards should also be set in **s** shadow variable



<img width="130" height="66" alt="Image" src="https://github.com/user-attachments/assets/7b09c413-a0e7-45b6-bd86-e3c64c4e1360" />

The segment control active CTA should be set in **s** shadow variable.



<img width="491" height="804" alt="Image" src="https://github.com/user-attachments/assets/5d1db28e-8f62-44c1-b231-fbc85924f6a2" />

All drawers should be set in **xl** shadow variable.



[...]

PR #1570, opened 2025-10-07
Update buttons styles

### LINKED ISSUE #1525: Update the primary CTA component style
We need to update the primary CTA component style to better match the Token page and the new brand book. Even though this won’t be a 1:1 match with the Token page (which uses a blue hover color that doesn’t follow UX heuristics for applications), we still want the component to feel more consistent and aligned.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=13427-14401&t=0GMrbU5acUzQ2tJ2-1)

### LINKED ISSUE #1528: Update the secondary CTA shadows values.
Now that we have shadow variables, we need to update the secondary CTA component to use them.

[Secondary CTA's should have the `sm` shadow variable.](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=13427-14401&t=0GMrbU5acUzQ2tJ2-1)

gndelia:
### Description

This PR updates the styles for the `Button` and `ButtonLink` components to match the `hemi.xyz` site.

Note that I also updated the duplicated styles in the Error Page and Not Found page. These styles were duplicated because the pages can be rendered when the Root Layout crashes, but that can be easily solved by just loading the CSS again. Regardless, the `Button` component can't be directly imported (because for the `ButtonLink` component, we automatically prepend the `networkType` value, and that implies reading the QueryString, which depends on `nuqs` (which requires an adapter that's on the Root Layout... ok that was long to write!). The way to solve this (without duplicating styles again), is to create CSS classes and use them from the `Button` and `ButtonLink`, as well as from the plain `button` and `a` elements in the error Pages. Back to basics!

### Screenshots

Unhandled error in the tunnel form (Root layout has not crashed here)

image

Unhandled error in the root layout (here, the whole page has crashed uncontrollably)

image

Not found page

image

And then general screenshots where the buttons have been updated

image

image

(With focus)

image

image

image

image

Hover effects:

https://github.com/user-attachments/assets/fb572e85-5fb1-4096-a35d-63a36d92f6cc

https://github.com/user-attachments/assets/3fcfe695-9e93-4a73-a667-f24804d6132c

### Related issue(s)

Closes #1528 
Closes #1525 
Related to #1526  and #1536 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/not-found.tsx
+++ b/portal/app/[locale]/not-found.tsx
@@ -27,7 +27,10 @@ export default function NotFound() {
             {t('not-found.description')}
           </h3>
         </div>
-        <a className="button button-primary" href="/">
+        <a
+          className="button--base button-primary button-small button-regular"
+          href="/"
+        >
           {t('not-found.action')}
         </a>
       </div>


portal/components/button/button.css
```diff
@@ -0,0 +1,62 @@
...
+.button--small--regular {
+  @apply h-8 gap-x-2 rounded-lg px-3 text-sm;
+}
+
+.button--x-large--icon {
+  @apply text-mid h-11 min-w-11 rounded-lg;
+}
+
+.button--x-large--regular {
+  @apply text-mid h-11 gap-x-2 rounded-lg px-4;
```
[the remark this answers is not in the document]
gndelia: it does exist in the tailwind config

portal/components/button/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  />
+)
+
+export const ButtonIcon = ({
+  size = 'small',
+  variant = 'primary',
+  ...props
+}: ButtonProps) => (
+  <button
+    className={`button--base  ${buttonSizePresets[size].icon} ${variants[variant]}`}
```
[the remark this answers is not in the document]
gndelia: Fixed in [f4c6384](https://github.com/hemilabs/ui-monorepo/pull/1570/commits/f4c638447bb2b792bcb72f4eec6ba6986f92200b)

dvnahuel:
LGTM! this issue also addressed the different main CTA that we have on the deposits / withdraws, right? (such as prove TX)

gndelia:
> LGTM! this issue also addressed the different main CTA that we have on the deposits / withdraws, right? (such as prove TX)

Not quite sure I follow - here it is how it looks in the Drawers

image

image

image

image

portal/components/button/button.css
```diff
@@ -0,0 +1,62 @@
+/*
+The following ".button" global classes were created because the <Button>
+and <ButtonLink> components of "components/button.tsx" file cannot be used
+by the error pages "app/not-found.tsx" and "app/global-error.tsx". These pages would be rendered if the root layout
+crashes, so Adapters for Nuqs are not available. We don't need them, but the button.tsx component needs them for the ButtonLink component.
+Because of that, we can add all the styles here, and then use plain buttons in those error pages.
+*/
+.button--base {
```
gabmontes, line 8: Is this following any known naming convention? It looks like BEM but is not quite that. If we can stick to a known standard, much better.

portal/components/button/button.css
```diff
@@ -0,0 +1,62 @@
...
+  before:bg-neutral-50 disabled:bg-neutral-50 disabled:opacity-55;
+}
+
+.button--tertiary {
+  @apply focus-visible:shadow-button-tertiary-focused bg-transparent
+    text-neutral-950 before:bg-neutral-100 focus:bg-neutral-100;
+}
+
+/* Button Sizes */
+.button--x-small--icon {
```
gabmontes, line 40: Perhaps these classes could be split into `button--xs`, `button--xs`, `button--lg` and `button--icon`, leaving the "regular" option in the base class if possible.

portal/components/button/button.css
gndelia: I assumed `primary`, `secondary`, and `tertiary` to be all variants. As well as the sizes (both icons and regular). None of them impacts any other child element. Because of this, I thought all of them would be modifiers, hence the usage of `--`

portal/components/button/button.css
gndelia: This is the styles that I am supposed to be implementing (Like, this is the [current implementation](https://github.com/hemilabs/ui-monorepo/blob/main/portal/components/button.tsx#L46))

```ts
const buttonSizePresets = {
  xSmall: {
    icon: 'h-7 text-xs rounded-md min-w-7',
    regular: 'h-7 text-xs gap-x-1 px-2.5 rounded-md',
  },
  small: {
    icon: 'h-8 text-sm rounded-lg min-w-8',
    regular: 'h-8 text-sm gap-x-2 px-3 rounded-lg',
  },
  xLarge: {
    icon: 'h-11 text-mid rounded-lg min-w-11',
    regular: 'h-11 text-mid gap-x-2 px-4 rounded-lg',
  },
}
```

I'm open to an alternative implementation to the current one As icon and regular are not combined, I thought this implementation would prevent using "half" of the styles. Icon needs to be implemented to the 3, so I am unsure `button--icon` will work. But yes, we could do `button--x-small` the `regular` and then `button--x-small--icon`, and do the same for the other 2. What do you think?

Issue #1619, opened 2025-11-05
Add hemiBTC Staking page


gndelia:
Umbrella issue for the hemiBTC Staking page. Desings TBD

dvnahuel:
These are the 'integrations' we want to showcase (we're still missing one) 

1. Spectra
2. Gearbox
3. TBD (!)

dvnahuel:
[Another issue related](https://github.com/hemilabs/ui-monorepo/issues/1653)

PR #1626, opened 2025-11-07
Hemibtc staking page pt1

gndelia:
### Description

This PR adds the bitcoin-yield page, behind a feature flag. The URL is TBD. This PR may be easier to read commit by commit

- a28f504ab7bd498869c957202eb36ef9b5276180 removes some redundant styles in `pageTitle`, as these styles were already defined in [globals.css](https://github.com/hemilabs/ui-monorepo/blob/43d99b2ccf9642a89313746ff86f33033e2ba939/portal/styles/globals.css#L21)
- 11a4cde33adcf16f4abd5838f1a35df64696246c updates the `Card` component to allow customization of the shadow. Prior to this, all cards had the same shadow. But the Cards used here need a different one.
- aa63b6085892ea68bbb2c9c2c0eff236c5708ab7 Moves the background from the `/stake` page into the `components` folder, so it can be used both in that page and in here
- 949e520eb47ba6b0261fce5f872eaf024cbc673a Updates the `hemi-btc-staking-actions` to allow reading  the [pricePerShare](https://github.com/vesperfi/yield-vault/blob/c988391def18748f4215a45684a99695e98fc540/src/YieldVault.sol#L206) from the YieldVault contract
- 4230563a473f524d045ba120eb36a33c2efa5555 Adds the Bitcoin Yield page. The table and integrations section are missing
- 98d908a38e830829b1d5b3da351ea999b3313ad0 Adds a feature flag, so we can enable this on staging

### Screenshots

<details>
<summary>When the feature flag is disabled, the user gets a 404</summary>
image
</details>

<details>
<summary>Page when the feature flag is enabled</summary>

image

</details>

Note that some of the values of the cards are hardcoded and will be developed in future PRs. Note also that the contract is not deployed yet with real values to read from

Note also that the mobile implementation is missing

### Related issue(s)

Related to #1619

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/hooks/useHemiBtcToken.ts
+++ b/portal/hooks/useHemiBtcToken.ts
@@ -3,28 +3,28 @@ import { useMemo } from 'react'
 import { EvmToken } from 'types/token'
 import { getTokenByAddress } from 'utils/token'
 import { type Address, type Chain } from 'viem'
-import { hemi as hemiMainnet, hemiSepolia } from 'viem/chains'
+import { hemi, hemiSepolia } from 'viem/chains'
 
 const HEMI_BTC_ADDRESSES: Record<Chain['id'], Address> = {
-  [hemiMainnet.id]: '0xAA40c0c7644e0b2B224509571e10ad20d9C4ef28',
+  [hemi.id]: '0xAA40c0c7644e0b2B224509571e10ad20d9C4ef28',
   [hemiSepolia.id]: '0x36Ab5Dba83d5d470F670BC4c06d7Da685d9afAe7',
 }
 
 export const useHemiBtcToken = function () {
-  const hemi = useHemi()
+  const hemiChain = useHemi()
 
   return useMemo(
     function () {
       const address =
-        HEMI_BTC_ADDRESSES[hemi.id as keyof typeof HEMI_BTC_ADDRESSES]
+        HEMI_BTC_ADDRESSES[hemiChain.id as keyof typeof HEMI_BTC_ADDRESSES]
       if (!address) {
         throw new Error(
-          `HemiBTC token address not found for chain ID: ${hemi.id}`,
+          `HemiBTC token address not found for chain ID: ${hemiChain.id}`,
         )
       }
       // hemiBTC is always an EvmToken
-      return getTokenByAddress(address, hemi.id) as EvmToken
+      return getTokenByAddress(address, hemiChain.id) as EvmToken
     },
-    [hemi.id],
+    [hemiChain.id],
   )
 }


portal/app/[locale]/bitcoin-yield/_components/cardInfo.tsx
```diff
@@ -0,0 +1,47 @@
...
+}: Props<T>) {
+  const getValue = function () {
+    if (isError) {
+      return '-'
+    }
+    if (data === undefined) {
+      return '...'
+    }
+    return formatValue(data)
+  }
```
Comment (gabmontes) [inline portal/app/[locale]/bitcoin-yield/_components/cardInfo.tsx]: It seems unnecessary to pass a `formatValue` function to the component. The caller could send no data, so the component renders `...` or the formatted data so the component renders that instead. What is the current benefit of injecting that function?

portal/components/paragraph.tsx
```diff
@@ -0,0 +1,18 @@
...
+  /* eslint-disable sort-keys */
+  normal: 'font-normal',
+  medium: 'font-medium',
+  semibold: 'font-semibold',
+  /* eslint-enable sort-keys */
+} as const
+
+type Props = {
+  children: ReactNode
+  className?: string
```
ArturDolzan: Very minor, but since className is exposed as a prop and appended after the internal variant styles, it could potentially override typography settings (like font weight or size) defined in the component. Thoughts?

portal/hooks/useHemiBtcToken.ts
```diff
@@ -0,0 +1,30 @@
...
+import { type Address, type Chain } from 'viem'
+import { hemi as hemiMainnet, hemiSepolia } from 'viem/chains'
+
+const HEMI_BTC_ADDRESSES: Record<Chain['id'], Address> = {
+  [hemiMainnet.id]: '0xAA40c0c7644e0b2B224509571e10ad20d9C4ef28',
+  [hemiSepolia.id]: '0x36Ab5Dba83d5d470F670BC4c06d7Da685d9afAe7',
+}
+
+export const useHemiBtcToken = function () {
+  const hemi = useHemi()
```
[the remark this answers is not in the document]
gndelia: Renamed here ea31a3d6754556fffdedbc4254253db8f1afa390

portal/components/paragraph.tsx
ArturDolzan: Got it, that makes sense. No problem then.. it was just a small concern about potential overrides, but your reasoning is totally fine. Perhaps in the future we may want to consider using `twMerge` to deal with that.

PR #1642, opened 2025-11-17
Add format to voting power

### LINKED ISSUE #1633: Voting power formatting issue
### Description

When displaying the voting power in the HEMI stake card, the HEMI value is not properly formatted nor trimmed.

Image

### Proposed fix

Use the same number formatting as in every other place so the number would read "12,345.679", with the comma and the decimals rounded to prevent having too many on display. The whole line should read:

```
12,345.679 HEMI = 12,345.679 veHEMI
```

arielmergen:
### Description

These commits add a function to format the data sent to the component InfoRow 
Also adds a new util in formats utils

### Screenshots
Captura de pantalla 2025-11-17 a la(s) 7 12 19 p  m

https://github.com/user-attachments/assets/f855c07f-d8da-4b43-8c39-fdcaf4918627

### Related issue(s)

Closes #1633 
Fixes #
Related to #

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx
+++ b/portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx
@@ -96,10 +96,6 @@ const InfoRow = ({ label, value }: { label: string; value: ReactNode }) => (
   </p>
 )
 
-const SymbolContainer = ({ children }: { children?: ReactNode }) => (
-  <span>{`ve${children?.toString().trim()}`}</span>
-)
-
 const VotingPowerEquivalence = ({
   amount,
   token,
@@ -110,13 +106,11 @@ const VotingPowerEquivalence = ({
   votingPower: string
 }) => (
   <div className="flex items-center gap-x-1">
-    <DisplayAmount amount={amount} showSymbol={true} token={token} />
+    <DisplayAmount amount={amount} token={token} />
     <span>=</span>
     <DisplayAmount
       amount={votingPower}
-      showSymbol={true}
-      symbolContainer={SymbolContainer}
-      token={token}
+      token={{ ...token, symbol: `ve${token.symbol}` }}
     />
   </div>
 )


arielmergen:
> Isn't it better to have a component that renders the numbers with the proper locale and decimals, and use it everywhere? A custom solution for this item is ok but we may be losing an opportunity to improve. cc/ @gndelia

Something like this?

`<FormattedTokenAmount 
  amount={input} 
  token={token}
  locale={locale}
  maximumFractionDigits={3}
/>`

@gabmontes @gndelia  ?

gndelia:
@gabmontes  We already have functions for formatting in [utils/format.ts](https://github.com/hemilabs/ui-monorepo/blob/main/portal/utils/format.ts) - @arielmergen you should check those. We don't need a component as we don't want to couple the formatting with a specific html tag.

portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx
```diff
@@ -87,13 +89,30 @@ export function getNearestValidValues({
...
+}: {
+  amount: string
+  token: EvmToken
+  votingPower: string
+}) => (
+  <div className="flex items-center gap-x-1">
+    <DisplayAmount amount={amount} showSymbol={true} token={token} />
+    <span className="text-neutral-600">=</span>
+    <DisplayAmount amount={votingPower} showSymbol={false} token={token} />
+    <span className="text-neutral-600">ve{token.symbol}</span>
```
Comment (gabmontes) [inline portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx]: The voting power will always be "veHEMI". Let's the component display the token symbol.

```suggestion
    <DisplayAmount amount={votingPower} showSymbol={true} token={token} />
```

portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx
```diff
@@ -87,13 +89,38 @@ export function getNearestValidValues({
...
+  votingPower,
+}: {
+  amount: string
+  token: Token
+  votingPower: string
+}) => (
+  <div className="flex items-center gap-x-1">
+    <DisplayAmount amount={amount} showSymbol={true} token={token} />
+    <span>=</span>
+    <DisplayAmount
```
Comment (ArturDolzan) [inline portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx]: In this case, even when using the symbolContainer property of the DisplayAmount component, it still renders the wrong symbol in the tooltip. See the video below.


https://github.com/user-attachments/assets/e486c68a-12d3-404b-99c3-fbe8124c5aa9

Comment (arielmergen) [reply portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx]: Thank you, @ArturDolzan. Fixed

ArturDolzan (approved):

portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx
```diff
@@ -87,13 +89,32 @@ export function getNearestValidValues({
...
+  amount: string
+  token: Token
+  votingPower: string
+}) => (
+  <div className="flex items-center gap-x-1">
+    <DisplayAmount amount={amount} token={token} />
+    <span>=</span>
+    <DisplayAmount
+      amount={votingPower}
+      token={{ ...token, symbol: `ve${token.symbol}` }}
```
Comment (gabmontes) [inline portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx]: The symbol of veHemi is "veHemi", not "veHEMI". Should we read that from the veHemi contract instead?

Comment (arielmergen) [reply portal/app/[locale]/staking-dashboard/_components/lockup/index.tsx]: As we discussed, I will create a new issue with this task. And we move forward with this workaround.

PR #1654, opened 2025-11-20
Iterate navbar Get Started CTA to save space

### LINKED ISSUE #1653: Iterate navbar 'Get started' CTA
As we’re adding a new navbar item ‘Bitcoin Yield’ we need to iterate on the ‘Get started’ card CTA so it doesn’t take up so much space.
On mobile, we just need to update the CTA to the ‘Secondary-S’ variant from the design system.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=16261-1421&t=lKHSSl8oDmidSkOb-1)
Screenshot:

Image

ArturDolzan:
### Description

This PR implements a simplified version of the navbar 'Get Started' CTA. The previous card-style CTA was taking up too much vertical space in the navbar, making it difficult to add new items like Bitcoin Yield. The new button design maintains the same functionality and tracking while being more compact and fitting better within the navbar layout.

### Screenshots

https://github.com/user-attachments/assets/eb8daa4e-5eef-44e1-b437-bee1c02183d0

### Related issue(s)

Closes #1653 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/_components/navbar/_components/tvl/index.tsx
+++ b/portal/app/[locale]/_components/navbar/_components/tvl/index.tsx
@@ -42,7 +42,7 @@ const TvlImpl = function () {
     isNotTestnet && (
       <section
         className="h-22 bg-sky-450 relative mx-0.5 mb-4 flex flex-col gap-y-3 rounded-lg 
-        p-4 md:mb-0 md:mt-2"
+        p-4 md:mb-0 md:mt-2.5"
       >
         <div className="flex w-full items-center justify-between">
           <span className="text-sm font-medium text-white">{t('tvl')}:</span>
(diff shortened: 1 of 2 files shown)


dvnahuel:
> > > In the mobile views, I think we should close the nav bar once the user click the "Getting Started" button and the page navigates there. I understand this PR did not change the old behavior so in that case, let's discuss that and open an issue. cc/ @dvnahuel
> > 
> > 
> > Yep! We should definitely do that now that the interaction on that action changed. Issue [here](https://github.com/hemilabs/ui-monorepo/issues/1655) @ArturDolzan
> 
> @dvnahuel in mobile views (phone), the menu already closes when the user taps any option. Are you referring to tablet view? If so, I think this behavior should apply to all menu options, not only “Get Started”. Could you please clarify?

Now that we have more time before the release, let disregard the issue and the comment, so I can think about a better solution...

PR #1672, opened 2025-11-28
Fix voting power to use contract delegation

### LINKED ISSUE #1640: Fix Voting Power to Include Vote Delegation
### Description

For the $HEMI staking, currently, the voting power calculation only uses `balanceOfNFT` code, but the voting power should include delegated votes.

### Proposed fix

veHemi.voteDelegation().getVotes(walletAddress)

ArturDolzan:
### Description

This PR adds `getPositionVotingPower` function that reads voting power directly from the vote delegation contract instead of calculating locally. It returns 0 when the position is delegated to another wallet, so users see accurate voting power for their positions. Also adds usePositionVotingPower hook and updates the VotingPower component.

Side note: There's also possible to get the total voting power for a wallet, which includes delegations received from other users (different positions), so if @dvnahuel wants to show this somewhere in the UI we can easily add it.. 

### Screenshots

https://github.com/user-attachments/assets/9cd72978-87fd-433d-b18e-858bd074713f

### Related issue(s)

Closes #1640
Fixes #1640

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/staking-dashboard/_hooks/usePositionVotingPower.ts
+++ b/portal/app/[locale]/staking-dashboard/_hooks/usePositionVotingPower.ts
@@ -1,4 +1,5 @@
 import { useQuery } from '@tanstack/react-query'
+import { useHemi } from 'hooks/useHemi'
 import { useHemiWalletClient } from 'hooks/useHemiClient'
 import { getPositionVotingPower } from 've-hemi-actions/actions'
 import type { Chain } from 'viem'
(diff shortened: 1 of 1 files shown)


gabmontes (approved):

portal/app/[locale]/staking-dashboard/_hooks/usePositionVotingPower.ts
```diff
@@ -0,0 +1,30 @@
...
+    enabled: !!hemiWalletClient && !!address && tokenId > BigInt(0),
+    queryFn: () =>
+      getPositionVotingPower({
+        client: hemiWalletClient!,
+        ownerAddress: address!,
+        tokenId,
+      }),
+    queryKey: getPositionVotingPowerQueryKey({
+      tokenId,
+    }),
```
Comment (gndelia) [inline portal/app/[locale]/staking-dashboard/_hooks/usePositionVotingPower.ts]: I think the chainId should also be included. Note that tokenIds are just numbers, so the tokenId `1` will exists in both mainnet and testnet but they will be different NFTs. I know it's unlikely to cause issues as testnet is disabled for prod, but for the sake of correctness, I think we should include it

Comment (ArturDolzan) [reply portal/app/[locale]/staking-dashboard/_hooks/usePositionVotingPower.ts]: Agreed. Applied here cbfb0916fc1b4c57c7aa63c092e91354fc5c43f6. Thanks!

gndelia (changes requested):

Issue #1671, opened 2025-11-28
Implement new BTC wallet connector.


dvnahuel:
With the update to the wallet connector, we need to apply the new styles to the drawer cards. These cards will now function as dropdowns: when the user clicks on a card, it should expand to display the available wallet connection options.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=16480-12445&t=nvcHLeifOoqy9bxN-1)

PR #1682, opened 2025-12-02
Add OKX bitcoin wallet connectors

gabmontes:
### Description

This PR adds a new bitcoin connector for the OKX Chrome extension and mobile app. It also improves the UniSat connector. The new connector is now exposed in the btc-wallet package but is not used in the portal. Additional work is needed to create a wallet selector and allow the users to connect either the UniSat or the OKX wallet.

The Binance mobile wallet was also tested but their newest version (3.6.3) disconnects right after the connection is established. It is possibly a bug with the wallet as in previous versions the connection was successful. Due to this problem, the connector for the Binance wallet was removed from this PR.

### Summary of features and tests

| Support | UniSat extension | OKX extension | OKX mobile | Binance mobile |
| ------- | ---------------- | ------------- | ---------- | -------------- |
| Testnet | yes              | no            | no         | no             |
| P2PKH   | yes              | yes           | yes        | no             |

| Connector method  | UniSat extension | OKX extension | OKX mobile | Binance mobile |
| ----------------- | ---------------- | ------------- | ---------- | -------------- |
| connect           | ok               | ok            | ok         | no             |
| disconnect        | ok               | ok            | ok         | ok             |
| getAccounts       | ok               | ok            | ok         | no legacy      |
| getBalance        | ok               | ok            | ok         | no             |
| getNetwork        | ok               | ok            | ok         | ok             |
| onAccountsChanged | ok               | ok            | n/a        | n/a            |
| onChainChanged    | ok               | n/a           | n/a        | n/a            |
| sendBitcoin       | ok               | ok            | yes        | n/a            |
| switchNetwork     | ok               | not impl      | not impl   | not impl       |

- UniSat mobile app does not allow browsing websites/dapps. It was possible before but the feature is no longer present.
- Binance Chrome extension does not currently have Bitcoin support.
- Binance mobile v3.6.3 refuses to connect.

### Related issue(s)

Related to #1671 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/packages/btc-wallet/connectors/unisat.ts
+++ b/packages/btc-wallet/connectors/unisat.ts
@@ -42,7 +42,7 @@ const wallet = {
   },
   id: 'unisat',
   isInstalled,
-  name: 'UniSat',
+  name: 'Unisat',
   onAccountsChanged(handler) {
     assertInstalled()
     provider.on('accountsChanged', handler)
--- a/packages/btc-wallet/utils/psbt.ts
+++ b/packages/btc-wallet/utils/psbt.ts
@@ -14,8 +14,10 @@ export async function sendBitcoin(
   satoshis: number,
   options: { memo?: string } = {},
 ) {
-  const [address] = await provider.getAccounts()
-  const network = await provider.getNetwork()
+  const [[address], network] = await Promise.all([
+    provider.getAccounts(),
+    provider.getNetwork(),
+  ])
   const client = esploraClient({
     network: network === 'livenet' ? 'mainnet' : network,
   })


packages/btc-wallet/connectors/okx.ts
```diff
@@ -0,0 +1,72 @@
+import { Unisat } from '../unisat'
```
gndelia: Perhaps it's time to rename `Unisat` to something more generic. It feels a weird mix using Unisat here. What do you think?

packages/btc-wallet/connectors/okx.ts
```diff
@@ -0,0 +1,72 @@
...
+const provider = (typeof window !== 'undefined' &&
+  window.okxwallet?.bitcoin) as Unisat
+
+const isInstalled = () => !!provider
+
+const assertInstalled = function () {
+  if (!isInstalled()) {
+    throw new Error('OKX Wallet is not installed')
+  }
+}
```
gndelia: I think here, when we do the `as Unisat` we're hiding the fact that `provider` may not be defined. It's true that we're handling it in runtime by calling `assertInstalled`, but if we forget to call `assertInstalled`, this will fail at runtime (if `provider` happens to be undefined) with a null pointer exception. I think we can tweak the code a bit to solve this so Typescript warns us if we don't check `assertInstalled`.

We can use our own assertion function, and do something like this


```suggestion
// here adding the | undefined type
const provider = (typeof window !== 'undefined' &&
  window.okxwallet?.bitcoin) as Unisat | undefined

// If this function does not throw when called, then TS ensures "p" is defined
// Note that typescript forces for these kind of functions to use plain function declarations, without assignments :/ See https://github.com/microsoft/TypeScript/issues/34523
function assertInstalled(
  p: Unisat | undefined,
): asserts p is NonNullable<Unisat> {
  if (!p) {
    throw new Error('OKX Wallet is not installed')
  }
}
```

This way, you need to update every function to work like this

```ts
async connect() {
    // note here I'm passin provider
    assertInstalled(provider)
    await provider.requestAccounts()
  },
```

and this is inferred correctly as `provider` being defined

Image

and if we forget to call `assertInstalled`

```ts
// this does not compile
 getAccounts() {
    return provider.getAccounts()
  },
```

See 

Image


What do you think?

packages/btc-wallet/connectors/okx.ts
gndelia: "BaseConnector" and if changes are needed, either OKX or unisat extend it lol

packages/btc-wallet/connectors/okx.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+  sendBitcoin(toAddress, satoshis, options) {
+    assertInstalled()
+    // The method sendBitcoin in the provider does not support options.memo.
+    // See: https://web3.okx.com/es-la/build/dev-docs/sdks/chains/bitcoin/provider#sendbitcoin
+    return sendBitcoin(provider, toAddress, satoshis, options)
+  },
+  switchNetwork(network) {
+    assertInstalled()
+    return provider.switchNetwork(network) // NOT IMPLEMENTED BY OKX
```
gndelia: So.. what happens here when switchNetwork is called? Does this throw at "not implemented"?
Should we add a clarification comment?

packages/btc-wallet/connectors/unisat.ts
```diff
@@ -1,67 +1,75 @@
...
   },
   getNetwork() {
     assertInstalled()
-    return window.unisat.getNetwork()
+    return provider.getNetwork()
   },
   id: 'unisat',
   isInstalled,
-  name: 'Unisat',
+  name: 'UniSat',
```
gndelia: This change may mess up our analytics, as it will show as 2 different wallets.

packages/btc-wallet/connectors/unisat.ts
```diff
@@ -1,67 +1,75 @@
...
   switchNetwork(network) {
     assertInstalled()
-    return window.unisat.switchNetwork(network)
+    return provider.switchNetwork(network)
   },
 } satisfies WalletConnector
 
 export const unisat = {
   downloadUrls: {
-    android: 'https://play.google.com/store/apps/details?id=io.unisat',
```
gndelia: Was the Android one removed for any particular reason? I think we should also add the iOS one.

packages/btc-wallet/utils/psbt.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+// eslint-disable-next-line complexity
+export async function sendBitcoin(
+  provider: Unisat,
+  toAddress: string,
+  satoshis: number,
+  options: { memo?: string } = {},
+) {
+  const [address] = await provider.getAccounts()
+  const network = await provider.getNetwork()
```
gndelia: It seems we can run these 2 in parallel

```suggestion
  const [[address], network] = await Promise.all([
    provider.getAccounts(),
    provider.getNetwork()
  ])
 
```

packages/btc-wallet/connectors/okx.ts
gabmontes: Good suggestion. But too much noise for no actual benefit IMO.

packages/btc-wallet/connectors/okx.ts
gabmontes: The provider throws an exception. If they fix that in the future, by adding support back, we change nothing.

packages/btc-wallet/connectors/unisat.ts
```diff
@@ -1,67 +1,75 @@
...
+// Exclude them from UniSat detection to avoid false positives.
+const isInstalled = () =>
+  !!provider && !(provider.isBinance || provider.isOkxWallet)
 
 const assertInstalled = function () {
   if (!isInstalled()) {
-    throw new Error('Unisat is not installed')
+    throw new Error('UniSat Wallet is not installed')
   }
 }
```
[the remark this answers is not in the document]
gabmontes: In a future PR?

packages/btc-wallet/utils/psbt.ts
```diff
@@ -0,0 +1,67 @@
...
+    network: bitcoin.networks[network === 'livenet' ? 'bitcoin' : network],
+  })
+  for (const input of inputs) {
+    const txHex = await client.bitcoin.transactions.getTxHex({
+      txid: input.txid as string,
+    })
+    psbt.addInput({
+      hash: input.txid,
+      index: input.vout,
+      nonWitnessUtxo: Buffer.from(txHex, 'hex'),
```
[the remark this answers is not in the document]
gabmontes: Even if we remove it from our code, `bitcoinjs-lib` uses Buffer extensively so the polyfill, if any, will have to be included anyway.

PR #1686, opened 2025-12-02
Imporve display error l1

### LINKED ISSUE #1674: Improve UX by Displaying L1 Input Errors

This ticket is the second iteration of this [issue](https://github.com/hemilabs/ui-monorepo/issues/1601).
Currently, no error is surfaced to the user in the L1 input; the input is simply disabled.
In this iteration, we need to display a proper message to the user instead of only disabling the input.

"No matching token was found."
"No se encontró ningún token coincidente."

Direction to test: 0x17BA356d3DFc89fc0cE3d6ABdfE1dd29a69A55cf

The user doesn’t receive any information about what happened.

### Proposed implementation

Check the tokenSelection.tsx component and update it to display an error message when the L2 token doesn’t have a counterpart.

Image

arielmergen:
### Description

Improve the logic to show the error when the L1 address is not configured in the contract."

### Screenshots

https://github.com/user-attachments/assets/51f0990d-2d1c-4f91-adf4-9c70d3f3cb50

Fix here Loading.

https://github.com/user-attachments/assets/f91886d2-f832-4130-bec7-cbd3929481b9

### Related issue(s)

Closes #1674
Fixes #
Related to #

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/portal/components/customTokenDrawer/index.tsx
+++ b/portal/components/customTokenDrawer/index.tsx
@@ -49,8 +49,12 @@ const canSubmit = ({
 const getL1AddressValidity = function (
   l1CustomToken: Token | undefined,
   l2customToken: L2Token | undefined,
+  isL2PlaceholderData: boolean,
 ) {
   if (l2customToken && l2customToken.address && !l2customToken.l1Token) {
+    if (isL2PlaceholderData) {
+      return undefined
+    }
     return 'l1-address-not-configured'
   }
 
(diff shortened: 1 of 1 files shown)


gndelia (approved):

portal/components/customTokenDrawer/index.tsx
```diff
@@ -46,8 +46,20 @@ const canSubmit = ({
...
+  isL2PlaceholderData: boolean,
+) {
+  if (l2customToken && l2customToken.address && !l2customToken.l1Token) {
+    if (isL2PlaceholderData) {
+      return undefined
+    }
+    return 'l1-address-not-configured'
+  }
+
+  return l1CustomToken ? 'this-address-is-valid' : 'this-address-is-not-valid'
```
gabmontes: The mix if nested `if`'s and conditionals make this difficult to read. What about something like this?

```suggestion
  if (l2customToken && l2customToken.address && !l2customToken.l1Token && isL2PlaceholderData) {
    return undefined
  } else if (l2customToken && l2customToken.address && !l2customToken.l1Token && !isL2PlaceholderData) {
    return 'l1-address-not-configured'
  } else if (l1CustomToken) {
    return 'this-address-is-valid'
  }
  return 'this-address-is-not-valid'
```

gabmontes: As said, I think it makes the code much more clear. But I'm ok either way, as long as the logic is correct.

PR #1701, opened 2025-12-05
Remove const enums

### LINKED ISSUE #1503: Replace enums with const objects

Similar to https://github.com/hemilabs/ui-monorepo/pull/1488 - we should drop enums in favor of const objects. These are less buggy as they are plain js objects, and do not need translation from Typescript.

Once removed, we could enable `erasableSyntaxOnly` so they are not used in the future

gndelia:
### Description

Doing some small improvements on Typescript: This PR removes the usage of Typescript Enums (Recommended [reading](https://www.totaltypescript.com/why-i-dont-like-typescript-enums)), and enables the [erasableSyntaxOnly](https://www.totaltypescript.com/erasable-syntax-only) in the `tsconfig.json` so they are no longer used. Not using Enums makes us closer to running Typescript code with Node 24 using `node index.ts` 🙌🏽 

Enums acted both as runtime objects and types, so the replacement requires defining both the const object (which replaces the runtime version) and the type manually (which replaces the enum as a type)

Using const objects also brings another problem when creating inline arrays (as the types are stricter than anyone would think), so I'm adding [@ts-reset](https://www.totaltypescript.com/ts-reset-article#make-includes-on-as-const-arrays-less-strict), which has a rule that makes `includes` work as anyone would expect. This allows us to remove many `@ts-expect-error` and makes the code more JavaScript-ish while keeping type safety 🙌🏽 

I also added it in the `subgraph-api` but as there are no runtime changes I don't think we need to deploy a new version

### Screenshots

No changes to users

### Related issue(s)

Closes #1503

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/staking-dashboard/_components/stakeReview/reviewStake.tsx
+++ b/portal/app/[locale]/staking-dashboard/_components/stakeReview/reviewStake.tsx
@@ -13,7 +13,7 @@ import { useNeedsApproval } from 'hooks/useNeedsApproval'
 import { useTranslations } from 'next-intl'
 import {
   StakingDashboardStatus,
-  StakingDashboardStatusType,
+  type StakingDashboardStatusType,
 } from 'types/stakingDashboard'
 import { getNativeToken } from 'utils/nativeToken'
 import { parseTokenUnits } from 'utils/token'


subgraph-api/reset.d.ts
```diff
@@ -0,0 +1,4 @@
+// unsure why but eslint can't recognize this is installed
+// and these 2 rules fail. Not ideal but..
+// eslint-disable-next-line node/no-missing-import
+import '@total-typescript/ts-reset/array-includes' // eslint-disable-line node/no-unpublished-import
```
gabmontes: The linter considers dev dependencies as "unpublished". The idea is that even when you run "npm i --production", everything will work just fine.

In this case, it seems to be considering `ts-reset` is required at runtime, thus the error.

packages/btc-wallet/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
-    "baseUrl": "./"
+    "baseUrl": "./",
+    "erasableSyntaxOnly": true
```
gabmontes: Can we have this defined in a "root" config where every config automatically inherits from?

If we have to create a root tsconfig and then modify every other to explicitly inherit from that one, partially defeats the purpose but ensures we don't forget to add rules to specific configs and helps adding rules to all configs at once.

gndelia: Not really easy. `extends` only supports extending from 1 config (See https://github.com/microsoft/TypeScript/issues/37884), and we're already extending from `@tsconfig/bases`, using different bases depending on the code running on the browser (the portal,btc-wallet) or not-being ui based (the rest basically)

We _could_ get away with 2 root files (one for each type) if you feel like it, but I think it increases the complexity a bit. I'm looking forward to this option being added at `@tsconfig/bases` though.

Would you still prefer me to add 2 root config files?

ArturDolzan (commented):
I left a minor improvement otherwise LGTM.

Issue #1769, opened 2026-01-07
Throttle requests to Hemi RPC


gndelia:
### Feature or change description

New limits have been set in the Hemi RPC (both testnet and mainnet)

Requests are limited to 300 req/min

We should solve at the provider level so that we don't surpass this limit (as we get 429 requests)

gndelia:
AFAIK this hasn't been a problem. Closing

PR #1774, opened 2026-01-09
Simplify hemi public client

gndelia:
### Description

As part of #1769  I'm trying to simplify a bit how viem clients are created/managed and used, now that I've been using the library for a while and gaining experience with it.

This PR does 2 things:

First, it exports the actions from `hemi-viem-stake-actions` from the `/actions` folder, matching the structure of other viem packages (making it more consistent). That's on 71de73894356cd81edb4163f66886e478003f4ad

Then, more importantly, replaces usages of `hemiClient` from 

```typescript
await hemiClient.getVaultChildIndex()
```

to

```typescript
await getVaultChildIndex(hemiClient)
```

The benefit, besides being better for [tree shaking](https://viem.sh/docs/clients/custom.html#tree-shaking), is that these functions accept a regular `Client` object that's coming from `viem`. This means that when calling this functions, I no longer need a `HemiPublicClient` type that has defined those functions (this was typed in [here](https://github.com/hemilabs/ui-monorepo/blob/0e2b7a6e3edc3795ce94caba0f9f09505e892f41/portal/hooks/useHemiClient.ts#L75)) - I can just import the function I want to use and use a regular viem client. This also simplifies the creation of `publicClient` as I no longer need to [extend the hemi client with functions even though they may not be used](https://github.com/hemilabs/ui-monorepo/blob/0e2b7a6e3edc3795ce94caba0f9f09505e892f41/portal/hooks/useHemiClient.ts#L41).

The commit 250622c559366d5b0c961e289b329fe8d07ccfa6 does that, as well as updating the tests

### Screenshots

No changes to users

### Checklist

- [x] Manual testing passed.


gabmontes (approved):
LGTM

PR #1834, opened 2026-03-11
Replace hooks with the published

### LINKED ISSUE #1807: Replace hooks with the published ones

Some hooks have been published in [@hemilabs/react-hooks](https://www.npmjs.com/package/@hemilabs/react-hooks). All the implementations were taken from here. We should replace the local hooks with the published versions.

Some notes:

1- It's published but untested 😅  so test after changes!
2- Some implementations were slightly modified (like the `useEnsureConnectedTo` as it does not consider bitcoin), so you may want to review if the local copy should extend the published version

arielmergen:
### Description

Dependencies (portal): Add @hemilabs/react-hooks@1.3.0 and @hemilabs/wallet-watch-asset@1.0.2; remove wallet-watch-asset (GitHub) and delete patches/wallet-watch-asset+1.0.1.patch.

**Hooks:** Use or extend the published implementations: re-export from the package for useDebounce, useOnClickOutside, useOnKeyUp, useWindowSize, and useUpdateNativeBalanceAfterReceipt; use package useNativeBalance in useBalance.ts and keep local useTokenBalance; extend package useEnsureConnectedTo with Bitcoin support locally; switch useAddTokenToWallet to import from @hemilabs/wallet-watch-asset; update comment in useWatchedAsset.ts.

---

## Implementation changes (logic modified or rewritten)

| File | Change |
|------|--------|
| **portal/hooks/useEnsureConnectedTo.ts** | Rewritten: uses the package hook for EVM and adds Bitcoin support locally. |
| **portal/hooks/useBalance.ts** | `useNativeTokenBalance` now uses `@hemilabs/react-hooks/useNativeBalance`; local implementation removed; `useTokenBalance` remains local. |

---

## Re-exports only (no new local logic)

| File | Change |
|------|--------|
| **portal/hooks/useDebounce.ts** | Re-export from `@hemilabs/react-hooks/useDebounce`. |
| **portal/hooks/useOnClickOutside.ts** | Re-export from `@hemilabs/react-hooks/useOnClickOutside`. |
| **portal/hooks/useOnKeyUp.ts** | Re-export from `@hemilabs/react-hooks/useOnKeyUp`. |
| **portal/hooks/useWindowSize.ts** | Re-export from `@hemilabs/react-hooks/useWindowSize`. |
| **portal/hooks/useInvalidateNativeBalanceAfterReceipt.ts** | Re-export of `useUpdateNativeBalanceAfterReceipt` from `@hemilabs/react-hooks/useUpdateNativeBalanceAfterReceipt`. |

---

## Other changes (import, comment, deps, patch)

| File | Change |
|------|--------|
| **portal/hooks/useAddTokenToWallet.ts** | Import only: `wallet-watch-asset` → `@hemilabs/wallet-watch-asset`. |
| **portal/hooks/useWatchedAsset.ts** | Comment only: storage key note updated. |
| **portal/package.json** | Added `@hemilabs/react-hooks` and `@hemilabs/wallet-watch-asset`; removed `wallet-watch-asset`. |
| **patches/wallet-watch-asset+1.0.1.patch** | Removed (dependency no longer used). |
| **package-lock.json** | Updated for dependency changes. |

---

**Summary:** Two files have real implementation changes: **useEnsureConnectedTo.ts** and **useBalance.ts**. The rest are re-exports, import/comment updates, or dependency/config changes.

### Related issue(s)

Closes #1807 
Fixes #
Related to #

### Checklist

- [ ] Manual testing passed.


gndelia (commented):
I think the hooks should be consumed directly from the lib, and we should drop the files that only work as "intermediates" here if they don't add any value

portal/hooks/useInvalidateNativeBalanceAfterReceipt.ts
```diff
@@ -1,30 +1 @@
...
-        nativeTokenBalanceQueryKey,
-        (old: GetBalanceReturnType) => ({
-          ...old,
-          value: old.value - effectiveGasPrice * gasUsed - amount,
-        }),
-      ),
-    [queryClient, nativeTokenBalanceQueryKey],
-  )
-}
+export { useUpdateNativeBalanceAfterReceipt } from '@hemilabs/react-hooks/useUpdateNativeBalanceAfterReceipt'
```
gabmontes: Isn't it better to replace the reference to this file with the reference to the new hook instead? Same for use ClickOutside, useOnKeyUp and useWindowSize. I would have expected all these files to be removed.

gabmontes (changes requested):
The changes are unfortunately not trivial as initially expected. Even some hooks like `useErc20TotalSupply` were renamed adding up to the changes and complexity of the review.

Tests show something has broken within the app. I.e. the tunnel page is unable to enable the swap button, most probably due to an issue with the approval hooks.

Image

Either retest the application to verify nothing was broken or split the PR into trivial hook replacements on one and all other hooks that needed to be adapted in another.

gabmontes (changes requested):
Even when the PR seems to be ok now, merging it seems risky without extensive testing. There are too many changes and some were certainly not trivial.

I suggest splitting this PR into many smaller ones that only address one concern at a time:

- One PR that replaces the local wallet-watch-asset with the package and re-apply the patch.
- One PR that replaces all the hooks that are exact matches. It on changes references from local hooks to ones in the hooks package.
- One PR per hook that is not an exact match, including logic changes, adapting code, etc.

@gndelia do you think taking such an approach is an overkill and we should merge the PR as it is anyway? Any better approach?

gndelia:
Can this one be closed? I think you addressed all (or most) in other PRs

PR #1854, opened 2026-04-02
Add Hemi Earn tabbed table

ArturDolzan:
### Description

This PR adds a two-tab table (Pools / My positions) below the info cards on the /hemi-earn page. Pools tab shows each vault's token, total deposits, APY breakdown tooltip, exposure tokens with hover-spread animation, and a deposit CTA button. My positions tab shows user deposits, APY, yield earned, and a manage button, with an empty state when no positions exist. Hooks are returning mocked data for now. CTAs will be implemented in upcoming PRs.

### Screenshots

https://github.com/user-attachments/assets/13eee334-8511-4537-a8a8-e15fea215ab9

Captura de Tela 2026-04-02 às 14 50 06

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/types.ts
+++ b/portal/app/[locale]/hemi-earn/types.ts
@@ -14,17 +14,17 @@ export type EarnCardData = {
 }
 
 export type EarnPool = {
-  vaultAddress: Address
-  token: EvmToken
   apy: { base: number; incentivized: number; total: number }
-  totalDeposits: bigint
   exposureTokens: { address: Address; chainId: EvmToken['chainId'] }[]
+  token: EvmToken
+  totalDeposits: bigint
+  vaultAddress: Address
 }
 
 export type EarnPosition = {
-  vaultAddress: Address
-  token: EvmToken
   apy: { base: number; incentivized: number; total: number }
-  yourDeposit: bigint
+  token: EvmToken
+  vaultAddress: Address
   yieldEarned: string
+  yourDeposit: bigint
 }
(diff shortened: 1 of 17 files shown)


gabmontes (dismissed):
uACK

portal/app/[locale]/hemi-earn/types.ts
```diff
@@ -1,11 +1,30 @@
...
+  exposureTokens: { address: Address; chainId: EvmToken['chainId'] }[]
+}
+
+export type EarnPosition = {
+  vaultAddress: Address
+  token: EvmToken
+  apy: { base: number; incentivized: number; total: number }
+  yourDeposit: bigint
+  yieldEarned: string
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/types.ts]: properties in the types in this file are not sorted (unsure why Eslint is not catching this..)

gndelia (commented):
Are the tables from this PR different from the generic Table component? Can't we use them for any reason?

ArturDolzan:
> Are the tables from this PR different from the generic Table component? Can't we use them for any reason?

@gndelia I think I remember why we didn't use it to btc-yield as well..  the generic component requires a fixed-height parent container (like staking-dashboard's h-[56dvh]) because it uses TanStack Virtual under the hood.. so the earn tables (and the btc-yield) have at most 1–3 rows and live in a naturally-scrolling page, so forcing a fixed height would cause awkward internal scrolling for almost no content. 

Although we could keep it in mind and iterate the generic component to accept this kind of tables, which we naturally know its height. We could open an issue for it. Thoughts?

portal/app/[locale]/hemi-earn/types.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/types.ts]: Applied on 731bd2af394d3bfc1eaa761a0fce0aae9d355384. Thanks!

gndelia (approved):

arielmergen (approved):

gabmontes (approved):
Still uACK

PR #1875, opened 2026-04-21
Add vault history endpoint to subgraph-api

ArturDolzan:
### Description

Expose VaultHistory data from the hemi-earn-vaults-subgraph via a new REST route: 
GET /:chainId/earn/vaults/:address/history. Supports an optional "since" query param (unix timestamp) for filtering by time period.

### Screenshots

https://github.com/user-attachments/assets/f1ec3826-681f-4c38-a907-aff1c041b49c

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/subgraph-api/README.md
+++ b/subgraph-api/README.md
@@ -23,6 +23,7 @@ These environment variables control how the API works:
 - `GET /:chainId/staked`
 - `GET /:chainId/claim/:address/:claimGroup`
 - `GET /:chainId/earn/vaults/:address/history`
+  - Optional query param: `since=<unix-seconds>` to return history entries from that timestamp onward.
 
 ## Local development and testing
 
--- a/subgraph-api/README.md
+++ b/subgraph-api/README.md
@@ -22,8 +22,7 @@ These environment variables control how the API works:
 - `GET /:chainId/withdrawals/:address/(btc|evm)`
 - `GET /:chainId/staked`
 - `GET /:chainId/claim/:address/:claimGroup`
-- `GET /:chainId/earn/vaults/:address/history`
-  - Optional query param: `since=<unix-seconds>` to return history entries from that timestamp onward.
+- `GET /:chainId/earn/vaults/:address/history?period=<1w|1m|3m|1y>`
 
 ## Local development and testing
 
--- a/subgraph-api/README.md
+++ b/subgraph-api/README.md
@@ -22,6 +22,7 @@ These environment variables control how the API works:
 - `GET /:chainId/withdrawals/:address/(btc|evm)`
 - `GET /:chainId/staked`
 - `GET /:chainId/claim/:address/:claimGroup`
+- `GET /:chainId/earn/vaults/:address/history`
 
 ## Local development and testing
 
--- a/subgraph-api/src/route-handlers/get-vault-history.ts
+++ b/subgraph-api/src/route-handlers/get-vault-history.ts
@@ -34,13 +34,11 @@ export async function getVaultHistoryHandler(
 ) {
   const { chainId } = req.data
   const { address } = req.params
-  const { period } = req.query
+  const { period = '1w' } = req.query
 
-  if (!period || !validPeriods.includes(period as Period)) {
+  if (!validPeriods.includes(period as Period)) {
     sendJsonResponse(res, 400, {
-      error: `period is required and must be one of: ${validPeriods.join(
-        ', ',
-      )}`,
+      error: `period must be one of: ${validPeriods.join(', ')}`,
     })
     return
   }
--- a/subgraph-api/src/route-handlers/get-vault-history.ts
+++ b/subgraph-api/src/route-handlers/get-vault-history.ts
@@ -2,7 +2,7 @@ import type { Request, Response } from 'express'
 import { type Address } from 'viem'
 
 import { getVaultHistory } from '../subgraph.ts'
-import { isInteger, sendJsonResponse } from '../utils.ts'
+import { sendJsonResponse } from '../utils.ts'
 
 type PathParameters = {
   address: Address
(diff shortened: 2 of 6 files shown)


subgraph-api/src/subgraph.ts
```diff
@@ -620,6 +620,60 @@ export const getMerkleClaim = function ({
...
+        vault
+      }
+    }`,
+    variables: { since: since.toString(), vault: address.toLowerCase() },
+  }
+
+  return request<GetVaultHistoryQueryResponse>(subgraphUrl, schema).then(
+    function (response) {
+      checkGraphQLErrors(response)
+      return response.data.vaultHistories
```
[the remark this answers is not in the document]
ArturDolzan: Applied on e83642c89f909377fc52506bbc0d7ccecb923f08. Thanks!

subgraph-api/README.md
```diff
@@ -22,6 +22,7 @@ These environment variables control how the API works:
 - `GET /:chainId/withdrawals/:address/(btc|evm)`
 - `GET /:chainId/staked`
 - `GET /:chainId/claim/:address/:claimGroup`
+- `GET /:chainId/earn/vaults/:address/history`
```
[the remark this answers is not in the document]
ArturDolzan: Good point! Applied on e83642c89f909377fc52506bbc0d7ccecb923f08. Thanks!

subgraph-api/src/subgraph.ts
```diff
@@ -620,6 +620,64 @@ export const getMerkleClaim = function ({
...
+
+  const subgraphUrl = getSubgraphUrl({ chainId, subgraphIds })
+
+  const schema = {
+    query: `query GetVaultHistory($vault: Bytes!, $since: BigInt!) {
+      vaultHistories(
+        where: { vault: $vault, timestamp_gte: $since }
+        orderBy: timestamp
+        orderDirection: asc
+        first: 1000
```
gndelia: The max limit for subgraphs is 100 entities, so this won't work as expected

```suggestion
        first: 100
```

You may need pagination (depending on the usage of this endpoint, it may or may not be important)

If you implement pagination, note also that using `skip` and `limit` works up to 5000 entities. After that, you need to start filtering with `where`.

You can check [this](https://github.com/hemilabs/ui-monorepo/blob/50d0ac9a4e5a34df958bb38581c1915c55553e99/portal/utils/sync-history/common.ts#L14) that we use when syncing the tunnel history with subgraphs

ArturDolzan: The _worst_ scenario would be 365 entities, so added a pagination on e127b408e999e951cf9a2f90463876ce46a894a3. Thanks for pointing that out!

gndelia (commented):
Is the idea to return the data of the past year? (last 365 days)

With the current API implementation, a consumer could just pass `?since=0` and the api would try to return all entities (particularly, given the `while(true)` with no enforcement of the API)

Perhaps if that's the only intended usage, you could just remove the `since` and always return the past year.

What are the use-cases you thought for this endpoint?

ArturDolzan:
> Is the idea to return the data of the past year? (last 365 days)
> 
> With the current API implementation, a consumer could just pass `?since=0` and the api would try to return all entities (particularly, given the `while(true)` with no enforcement of the API)
> 
> Perhaps if that's the only intended usage, you could just remove the `since` and always return the past year.
> 
> What are the use-cases you thought for this endpoint?

The component has a period selector (1w, 1m, 3m, 1y).. `since` lets the frontend fetch only the data for the selected period.

gndelia:
Then I would suggest narrowing down to those 4 values. The current implementation allows for any timestamp, which means that there's the open door for overloading the API by scaning the whole history

ArturDolzan:
> Then I would suggest narrowing down to those 4 values. The current implementation allows for any timestamp, which means that there's the open door for overloading the API by scaning the whole history

Good point! Replaced `since` with a required `period` param that only accepts 1w, 1m, 3m, 1y. The server calculates the timestamp internally, so no unbounded queries are possible. I also updated the video in the description.

gndelia (dismissed):

subgraph-api/src/route-handlers/get-vault-history.ts
```diff
@@ -0,0 +1,58 @@
...
+  const { period } = req.query
+
+  if (!period || !validPeriods.includes(period as Period)) {
+    sendJsonResponse(res, 400, {
+      error: `period is required and must be one of: ${validPeriods.join(
+        ', ',
+      )}`,
+    })
+    return
+  }
```
gabmontes: Query string parameters are usually optional, not required. If not provided, let's assume a sane default (the most commonly used or expected value) instead of throwing an error.

ArturDolzan: Applied on ebba2389be4c4dfd40d81106dd01c41a3ee516d0. Thanks!

gabmontes (approved):

subgraph-api/src/route-handlers/get-vault-history.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export async function getVaultHistoryHandler(
+  req: Request<PathParameters, object, object, QueryParams> & { data: Data },
+  res: Response,
+) {
+  const { chainId } = req.data
+  const { address } = req.params
+  const { period = '1w' } = req.query
+
+  if (!validPeriods.includes(period as Period)) {
```
gndelia, line 39: Not a blocker, but just an FYI

In the portal we use `ts-reset` so this is no longer a problem

See [this file](https://github.com/hemilabs/ui-monorepo/blob/main/portal/reset.d.ts)

[this is the lib](https://www.npmjs.com/package/@total-typescript/ts-reset)
you may want to include it in the subgraph-api in a future PR

gndelia (approved):

PR #1884, opened 2026-04-30
Refactor Spinner color prop to variant

### LINKED ISSUE #1651: Use color variants for spinners

The [spinner component](https://github.com/hemilabs/ui-monorepo/blob/main/portal/components/spinner.tsx) has a `color` property (used for coloring the underlying SVG) but all in all, there are only 2 colors we've been using for it. We should use a variant mechanism (similar to buttons) to avoid hardcoding the same color everywhere

ArturDolzan:
### Description

This PR refactors the Spinner component to expose a typed `variant` prop instead of a free-form `color` string. The component now owns the small set of valid colors (`light`, `orange-600`, `orange-bright`, `orange-warm`) and call sites pick one by name, which keeps loading states visually consistent across the portal and avoids arbitrary hex values scattered around the codebase.

### Screenshots

No UI changes.

### Related issue(s)

Closes #1651 
Fixes #1651 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/components/spinner.tsx
+++ b/portal/components/spinner.tsx
@@ -1,15 +1,24 @@
 import React, { useId } from 'react'
 
+const variantColors = {
+  'light': '#FFF7F0',
+  'orange-600': '#FF4600',
+  'orange-bright': '#FF6A00',
+  'orange-warm': '#FF6C15',
+} as const
+
+type SpinnerVariant = keyof typeof variantColors
+
 type SpinnerProps = {
-  size?: 'xSmall' | 'small' | 'medium' | 'large' | number
-  color?: string
   className?: string
+  size?: 'xSmall' | 'small' | 'medium' | 'large' | number
+  variant?: SpinnerVariant
 }
 
 export const Spinner = function ({
   className = '',
-  color = '#FFF7F0',
   size = 'medium',
+  variant = 'light',
 }: SpinnerProps) {
   // IMPORTANT: We generate a unique ID for each spinner instance using React's useId() hook.
   // This ensures that when multiple spinners are used on the same page, each has its own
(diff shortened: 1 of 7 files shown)


gabmontes (dismissed):

portal/components/spinner.tsx
```diff
@@ -1,15 +1,24 @@
 import React, { useId } from 'react'
 
+const variantColors = {
+  'light': '#FFF7F0',
+  'orange-600': '#FF4600',
+  'orange-bright': '#FF6A00',
+  'orange-warm': '#FF6C15',
```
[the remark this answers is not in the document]
gndelia: This suggestions seem correct 🤔  was it invalidated for some reason?

gndelia: Maybe it's me, but I can't tell the difference between these 2

Image

Should we unify?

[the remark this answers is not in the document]
ArturDolzan: I decided against because out of the 4 colors in variantColors, only #FF4600 has a matching constant in styles/index.ts.. Importing orange600 for just one would leave the map half-inline / half-reference.. but I'm good with both. Thoughts?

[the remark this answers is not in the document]
gndelia: I think `FF4600 ` was relevant because it is (was?) the Hemi orange... I don't think it half inline- half reference is a problem imho 🤔

ArturDolzan: Indeed, they're pretty similar.. if @dvnahuel is not against it, I vote for unifying.

gndelia: Note that they may be part of the orange theme, so we can't change them there, but we can unify them in the spinner, using only 3 variants instead of 4

dvnahuel: We should be using `#FF4600` which is the right Hemi orange color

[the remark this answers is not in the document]
ArturDolzan: Applied on 9712cd49ff49d7a4c3e595a8ed094f903793c467. Thanks!

ArturDolzan: Applied on 9712cd49ff49d7a4c3e595a8ed094f903793c467. Thanks!

gabmontes (approved):

gndelia (approved):

PR #1901, opened 2026-05-08
Test Table safeData reference stability

ArturDolzan:
### Description

This PR adds a regression guard for the Table infinite render loop. The bug itself is already gone, but the failure mode is silent — type-check, lint, build and the existing test suite all pass when the empty-data branch in `useTableData` returns a fresh `[]` literal, even though useReactTable's `getCoreRowModel` memo invalidates every render and locks up the app via `_autoResetPageIndex` to `setPagination`.

To make it testable without dragging `@testing-library/react` + `jsdom` into the portal (the current convention is pure-function tests only), I extracted the array-selection logic out of `useTableData` into a new `getSafeData` helper in `components/table/_utils`. The hook now just calls `getSafeData(data, showSkeleton, skeletonArray)`. Behaviour is unchanged.

### Screenshots

https://github.com/user-attachments/assets/51e7503b-cfdb-4315-92ef-b9fe40f4d1d4

### Related issue(s)

Related to #1898 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/components/table/_hooks/useTableData.tsx
+++ b/portal/components/table/_hooks/useTableData.tsx
@@ -1,7 +1,7 @@
 import { useMemo } from 'react'
 import Skeleton from 'react-loading-skeleton'
 
-import { getNewColumnOrder } from '../_utils'
+import { getNewColumnOrder, getSafeData } from '../_utils'
 import { TableProps } from '../index'
 
 type Props<TData> = Omit<
@@ -16,9 +16,6 @@ type Props<TData> = Omit<
   > & {
     width: number
   }
-
-// using a stable reference
-const emptyArray: unknown[] = []
 
 export function useTableData<TData>({
   columns,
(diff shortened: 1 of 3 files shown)


gndelia (approved):
Nice 🎉 

This will prevent changes on the hook that could break this.

There's a more subtle case: `data` itself (passed to `getSafeData`) should be stable as well.

So far, it hasn't been a problem because all usages, when passed to the `Table` component, use `useMemo`. In this tests, you can't reproduce it because `data` itself is a parameter of `getSafeData`

I am unsure if this specific scenario can be covered with a test. But perhaps we can add it to the copilot-instructions / Claude to always pass a stable reference to Table, and see how it works. It's one of these things that's hard to caught with deterministic tools (?)

Give it some thought (not needed for this PR). Approving.

arielmergen (approved):

PR #1912, opened 2026-05-14
Fix veHemi logo valid tooltip

### LINKED ISSUE #1874: Hovering veHEMI shows no logo

In the Governance Staking page, when hovering on the Voting Power value, as the token value is "rounded", it tries to use DisplayAmount, which renders the token logo... but veHEMI is not a real token so it has no "logo". So it ends up looking like this

Image

We should either hide the logo or define one (or use the Hemi one?)

This also happens in the "Your Positions" section and the "Total Voting Power" section

arielmergen:
### Description

Fix the Governance Staking tooltip for **veHEMI** voting power amounts shown via `DisplayAmount`.

## Changes

1. **`portal/hooks/useVeHemiToken.ts`**  
   The veHEMI contract token from `useToken` often has no `logoURI` (and no `l1LogoURI`), so `TokenLogo` fell back to the generic badge and looked wrong next to the full-precision amount. We merge in HEMI’s `logoURI` and `l1LogoURI` when the ve token is missing them, since veHEMI is a vote-escrowed HEMI and should present the same visual identity.

2. **` portal/components/displayAmount.tsx`** (tooltip content)  
   Replaced the flex wrapper from `<span>` to `<div>` because `TokenLogo` renders block-level nodes (`<div>`). A `<span>` must only contain phrasing content; nesting a `<div>` inside it is invalid HTML and relies on browser error recovery.

3. **` portal/components/tooltip/base.tsx`** (overlay)  
   Replaced the wrappers around dynamic `{text}` from `<p>` to `<div>` for the `simple`, `info`, and rich body cases. A `<p>` cannot legally contain `<div>` (e.g., from `TokenLogo`), which again produced invalid markup and unpredictable DOM repair.

### Screenshots

Captura de pantalla 2026-05-14 a la(s) 4 25 51 p  m

### Related issue(s)

Closes #1874 
Fixes #1874 
Related to #

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/portal/components/displayAmount.tsx
+++ b/portal/components/displayAmount.tsx
@@ -45,7 +45,7 @@ export const DisplayAmount = function ({
       id={`amount-tooltip-${token.symbol}`}
       text={
         notZero ? (
-          <span className="flex items-center gap-x-1">
+          <div className="flex items-center gap-x-1">
             {showTokenLogo && (
               <TokenLogo size="small" token={token} version={logoVersion} />
             )}
@@ -56,7 +56,7 @@ export const DisplayAmount = function ({
               // @ts-expect-error NumberFormat.format accept strings, typings are wrong. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#parameters
               amount,
             )} ${token.symbol}`}</span>
-          </span>
+          </div>
         ) : null
       }
       variant="simple"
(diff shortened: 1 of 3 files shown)


gabmontes (changes requested):
Fix the PR description. It looks like some edit is missing.

arielmergen:
Done @gabmontes, thank you

gabmontes (approved):
LGTM

ArturDolzan (approved):

PR #1921, opened 2026-05-18
Add hemi-earn-router-subgraph for deposits

### LINKED ISSUE #1916: Create a subgraph for Router deposits (Hemi earn)
## Context

`Router.sol` is the user-facing entry point on Hemi for cross-chain deposits into Vetro vaults. A deposit is not atomic — it travels Hemi → remote chain via LayerZero, gets fulfilled (or undone) by an off-chain Agent, and then either auto-settles or waits for the user to claim/recover. We need a subgraph (The Graph) on the Hemi network that surfaces each deposit's full lifecycle so the UI can show in-flight, claimable, undone, and finalized deposits.

Redeem operations are out of scope for this first pass — only `requestRedeem` is in the contract, no UI flow is wired up yet — but the data model and handlers must coexist with redeem events safely (which share `RequestFulfilled`/`RequestClaimed`/`RequestUndone`/`RequestRecovered` with deposits) so redeem can be added later without rewriting the schema.

## Goal

Create a new workspace package `subgraphs/hemi-earn-router-subgraph` that indexes the Router contract on Hemi mainnet and exposes one `Deposit` entity per request, with its current lifecycle status and the transaction hashes that drove each transition.

## Lifecycle to model

```
DepositRequested → IN_TRANSIT ─┬─ RequestFulfilled → FULFILLED → RequestClaimed   → DEPOSITED  (terminal)
                               └─ RequestUndone    → UNDONE    → RequestRecovered → CANCELLED  (terminal)
```

All transition events are emitted by the Router. `RequestClaimed` / `RequestRecovered` fire whether the settlement was automatic (triggered inside `_handleRequestFulfillment` / `_handleRequestUndo` when `automatic = true`) or user-initiated (`claimDeposit` / `recoverDeposit`); the subgraph does not need to distinguish these — only capture the tx hash.

## Schema

```graphql
enum DepositStatus {
  IN_TRANSIT
  FULFILLED
  DEPOSITED
  UNDONE
  CANCELLED
}

type Deposit @entity(immutable: false) {
  id: ID!                  # requestId as string
  requestId: BigInt!
  asset: Bytes!            # Hemi-side underlying token
  amount: BigInt!          # `assets` from DepositRequested
  receiver: Bytes!
  automatic: Boolean!
  initiatedAt: BigInt!     # block.timestamp of DepositRequested (set once, never updated)
  status: DepositStatus!
  depositTxHash: Bytes!    # tx hash of DepositRequested
  claimTxHash: Bytes       # tx hash of RequestClaimed (null until claimed)
  recoverTxHash: Bytes     # tx hash of RequestRecovered (null until recovered)
}
```

Notes:
- `initiatedAt` is set once at request time and never updated (matches the "date initiated" requirement).
- `automatic` is stored so the UI can explain the expected flow to the user.
- `id` uses `requestId.toString()` because requestIds are sequential `uint256`s scoped to a single Router instance.
- `DepositStatus` is intentionally a **GraphQL enum** (string values on the wire), not an `Int`. The Graph's AssemblyScript codegen produces string constants for enum fields; values travel as JSON strings; downstream layers (API, frontend) should generate their TS types directly from this schema with `graphql-codegen` so all three layers share one source of truth and codegen catches drift at build time. Numeric encoding would mean every layer carries its own lookup table, and the contract's on-chain `uint8` `Status` enum is coarser than this lifecycle anyway (it folds claimed + recovered into `FINALIZED`), so there is no single-number-source-of-truth to mirror.

## Events to handle

From `Router.sol`:

| Event | Signature (for subgraph.yaml) | Handler behavior |
|---|---|---|
| `DepositRequested` | `DepositRequested(indexed uint256,indexed address,uint256,indexed address,bool)` | Create `Deposit` entity. `status = IN_TRANSIT`, fill all initial fields, set `depositTxHash`. |
| `RequestFulfilled` | `RequestFulfilled(indexed uint256,uint256)` | Load `Deposit` by `requestId`. If `null`, return (it's a redeem). Otherwise set `status = FULFILLED`. |
| `RequestClaimed` | `RequestClaimed(indexed uint256,indexed address)` | Load `Deposit` by `requestId`. If `null`, return. Otherwise set `status = DEPOSITED`, `claimTxHash = event.transaction.hash`. |
| `RequestUndone` | `RequestUndone(indexed uint256,uint256)` | Load `Deposit` by `requestId`. If `null`, return. Otherwise set `status = UNDONE`. |
| `RequestRecovered` | `RequestRecovered(indexed uint256,indexed address)` | Load `Deposit` by `requestId`. If `null`, return. Otherwise set `status = CANCELLED`, `recoverTxHash = event.transaction.hash`. |

The `null-return` pattern on the four shared events is what keeps redeem-related emissions from corrupting the deposit store. When redeem is added later, those handlers will load from a separate `Redeem` entity and the deposit handlers stay unchanged.

`RedeemRequested` is **not** subscribed to in this iteration (no Redeem entity yet), but the ABI file should still contain it so adding it later doesn't require an ABI rebump.

## Package structure

Model after `subgraphs/hemi-earn-vaults-subgraph` (newest layout, has matchstick tests) combined with the multi-event handler pattern from `subgraphs/hemi-tunnel-withdrawals-proof-claim-subgraph`:

```
subgraphs/hemi-earn-router-subgraph/
├── abis/
│   └── Router.json              # full Router ABI (so RedeemRequested is there for future use)
├── src/
│   └── mappings/
│       └── router.ts            # 5 handlers; load-then-update pattern; null-return on shared events
├── tests/
│   ├── router.test.ts           # matchstick-as suite
│   └── router-utils.ts          # newMock* helpers for each event
├── schema.graphql
├── subgraph.yaml                # single dataSource, network: hemi, eventHandlers (not blockHandlers)
├── networks.json                # only `hemi` key; address + startBlock will be filled in once Router is deployed
├── package.json                 # mirror hemi-earn-vaults-subgraph scripts, scoped to hemi only
└── tsconfig.json
```

### `package.json` shape

Copy from `subgraphs/hemi-earn-vaults-subgraph/package.json` and drop the `hemi-sepolia` variants. Keep:

- `@graphprotocol/graph-cli@0.97.1`, `@graphprotocol/graph-ts@0.38.1`
- devDeps: `assemblyscript@0.19.23`, `matchstick-as@0.6.0`
- scripts: `codegen`, `build`, `build:hemi`, `deploy`, `deploy-studio:hemi`, `deploy-local:hemi`, `create-local`, `remove-local`, `graph:auth`, `prepare`, `pretest`, `test`.
- `name: "hemi-earn-router-subgraph"`, `version: "1.0.0"`.

### `networks.json`

```json
{
  "hemi": {
    "Router": {
      "address": "0x0000000000000000000000000000000000000000",
      "startBlock": 0
    }
  }
}
```

Placeholders until the Router is deployed on Hemi. Bumping `package.json` version after filling these in is what triggers the deployment workflow.

### `subgraph.yaml`

Single dataSource, `name: Router`, `network: hemi`, `kind: ethereum`, `apiVersion: 0.0.7`, five `eventHandlers` (one per event above), `file: ./src/mappings/router.ts`.

### Mapping handlers

AssemblyScript pattern, follows `subgraphs/hemi-tunnel-withdrawals-proof-claim-subgraph/src/mappings/portal.ts`:

```ts
export function handleDepositRequested(event: DepositRequestedEvent): void {
  const id = event.params.requestId.toString()
  const deposit = new Deposit(id)
  deposit.requestId = event.params.requestId
  deposit.asset = event.params.asset
  deposit.amount = event.params.assets
  deposit.receiver = event.params.receiver
  deposit.automatic = event.params.automatic
  deposit.initiatedAt = event.block.timestamp
  deposit.status = 'IN_TRANSIT'
  deposit.depositTxHash = event.transaction.hash
  deposit.save()
}

export function handleRequestFulfilled(event: RequestFulfilledEvent): void {
  const deposit = Deposit.load(event.params.requestId.toString())
  if (deposit == null) return // redeem, not a deposit
  deposit.status = 'FULFILLED'
  deposit.save()
}

// …same pattern for RequestClaimed (DEPOSITED + claimTxHash),
// RequestUndone (UNDONE), RequestRecovered (CANCELLED + recoverTxHash).
```

### Tests (matchstick-as)

Mirror the structure of `subgraphs/hemi-earn-vaults-subgraph/tests/`. One `*-utils.ts` with factory helpers built on `newMockEvent()` to construct each of the five events, and one `router.test.ts` with `describe`/`beforeEach(clearStore)` blocks. Coverage:

1. `DepositRequested` creates a `Deposit` with status `IN_TRANSIT` and all fields populated.
2. Happy path: deposit → `RequestFulfilled` flips to `FULFILLED` → `RequestClaimed` flips to `DEPOSITED` and writes `claimTxHash`.
3. Undo path: deposit → `RequestUndone` flips to `UNDONE` → `RequestRecovered` flips to `CANCELLED` and writes `recoverTxHash`.
4. `RequestFulfilled` / `RequestClaimed` / `RequestUndone` / `RequestRecovered` for an unknown `requestId` (simulating a redeem) is a no-op: entity count stays 0, no errors.
5. `automatic = true` does not change handler behavior (entity still walks the same states; only the tx hashes differ in source).

## Other repo changes

- **`subgraphs/README.md`** — no changes; the existing instructions already cover any subgraph in this folder.
- **CI matrix in `.github/workflows/subgraph-deployment.yml`** — **NOT** modified in this iteration. The subgraph is incomplete (no real Router address, no real ABI) so it must not be picked up by the deployment workflow. Adding it is deferred until the pending items below are resolved.

## Acceptance criteria

- `pnpm --filter hemi-earn-router-subgraph codegen` and `pnpm --filter hemi-earn-router-subgraph build:hemi` both succeed (using whatever placeholder ABI / address is in place — see "Pending" below).
- `pnpm --filter hemi-earn-router-subgraph test` passes the five matchstick scenarios above.
- A deposit going through the happy path or the undo path exposes a single `Deposit` entity whose `status` and `*TxHash` fields reflect the latest event, and querying by `requestId` returns it.
- Events emitted by a *redeem* request do not create or mutate any `Deposit` entity.

## Pending before this subgraph can be deployed

The work above lands the package, the schema, the handlers and the matchstick tests, but the subgraph is **not** ready to ship until all of the following are done in follow-up work:

1. **Router address & startBlock on Hemi.** Replace the zero placeholders in `subgraphs/hemi-earn-router-subgraph/networks.json` with the deployed Router proxy address and the block where it was deployed.
2. **Final Router ABI.** Replace `subgraphs/hemi-earn-router-subgraph/abis/Router.json` with the ABI emitted by the contracts repo's build artifacts for the deployed version. Re-run `pnpm codegen` and `pnpm build:hemi` to confirm the bindings still compile.
3. **Add to the CI matrix.** Append `hemi-earn-router-subgraph` to the `strategy.matrix.subgraph` list in `.github/workflows/subgraph-deployment.yml` so The Graph Studio deployment fires on version bumps.
4. **Create the Subgraph Studio entry.** Per `subgraphs/README.md`, manually create `hemi-earn-router-subgraph-hemi` in The Subgraph Studio and authorize the CLI before the first deploy.
5. **(Optional, when redeem lands)** Add a `Redeem` entity, a `RedeemRequested` handler, and update the four shared-event handlers to also load/update the `Redeem` entity. The deposit handlers should not need to change.

## References

- Template for layout & tests: `subgraphs/hemi-earn-vaults-subgraph/` (package.json scripts, networks.json shape, matchstick test structure under `tests/`).
- Template for multi-event lifecycle handlers: `subgraphs/hemi-tunnel-withdrawals-proof-claim-subgraph/src/mappings/portal.ts` (load-or-create-then-update pattern, mutable entity).
- Local & studio deployment instructions: `subgraphs/README.md`.
- CI workflow to update: `.github/workflows/subgraph-deployment.yml`.

gndelia:
### Description

New workspace package `subgraphs/hemi-earn-router-subgraph/` that indexes the `Router.sol` contract on Hemi into a unified `Request` entity with a `kind` discriminator (`DEPOSIT` / `REDEEM`) and a five-state lifecycle:

```
PENDING → FULFILLED → CLAIMED   (success terminal)
PENDING → UNDONE    → RECOVERED (failure terminal)
```

The schema is deliberately unified so future redeem support lands additively — subscribing to `RedeemRequested` and writing `kind = 'REDEEM'` to the same entity, no migration needed.

**Caveats:**

- **Redeems are not included in this PR.** Only the five deposit-lifecycle events are subscribed to (`DepositRequested`, `RequestFulfilled`, `RequestClaimed`, `RequestUndone`, `RequestRecovered`). The four events shared with redeem null-return on unknown `requestId`s so the deposit store stays isolated until redeem handlers are wired up later.
- **The code is untested against a real contract** — there is no deployed `Router` yet. The ABI in `abis/Router.json`, the address and `startBlock` in `networks.json`, and the manifest entries are placeholders. The handler implementation may need adjustments once the contract is finalized and deployed.
- **This subgraph will not be deployed by CI.** It is intentionally absent from the `strategy.matrix.subgraph` list in `.github/workflows/subgraph-deployment.yml`. The follow-up work to flip on deployment (fill in the real address/startBlock/ABI, add to the matrix, create the Studio entry) is captured in the issue's "Pending" section.

Also adjusts `.knip.json`: adds a per-workspace `entry` override for the new package (which uniquely carries matchstick tests today) and adds `assemblyscript` to root `ignoreDependencies` (peer of `matchstick-as`, needed at IDE/runtime but never directly imported).

### Screenshots

N/A — subgraph package, no UI surface.

### Related issue(s)

Closes #1916

### Checklist

- [x] Manual testing passed.


subgraphs/hemi-earn-router-subgraph/subgraph.yaml
```diff
@@ -0,0 +1,35 @@
...
+    source:
+      abi: Router
+      address: "0x0000000000000000000000000000000000000000"
+      startBlock: 0
+    mapping:
+      kind: ethereum/events
+      apiVersion: 0.0.7
+      language: wasm/assemblyscript
+      entities:
+        - Deposit
```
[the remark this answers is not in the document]
gndelia: Fixed in [9b8475e6](https://github.com/hemilabs/ui-monorepo/pull/1921/commits/9b8475e6). The `entities` list was stale from the pre-refactor schema.

subgraphs/hemi-earn-router-subgraph/abis/Router.json
```diff
@@ -0,0 +1,115 @@
...
+        "indexed": true,
+        "internalType": "address",
+        "name": "receiver",
+        "type": "address"
+      }
+    ],
+    "name": "RequestRecovered",
+    "type": "event"
+  }
+]
```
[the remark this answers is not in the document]
gndelia: The ABI in this PR is an intentional minimal placeholder — five deposit-lifecycle events only — and will be replaced wholesale with the real contract ABI once `Router.sol` is deployed (called out under "Caveats" in the PR description). At that point `RedeemRequested` and every other event/function will come in via the deployed-contract ABI, not by hand-transcription. Adding the event signature in isolation today doesn't save a step later, so leaving it out keeps the placeholder honest about its scope.

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
```diff
@@ -0,0 +1,67 @@
...
+  request.save()
+}
+
+export function handleRequestFulfilled(event: RequestFulfilledEvent): void {
+  const request = Request.load(event.params.requestId.toString())
+  if (request == null) {
+    return
+  }
+  request.status = 'FULFILLED'
+  // This are the actual shares the user received
```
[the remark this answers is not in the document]
gndelia: Fixed in [165e695a](https://github.com/hemilabs/ui-monorepo/pull/1921/commits/165e695a) — both the line 31 "This are" and the line 53 "wil" / "does updates" issues.

.knip.json
```diff
@@ -1,13 +1,20 @@
...
   "workspaces": {
     "portal-backend/api": {
       "entry": "scripts/*.js"
     },
     "subgraphs/*": {
-      "entry": "src/mappings/*.ts"
+      "entry": ["src/mappings/*.ts"]
+    },
+    "subgraphs/hemi-earn-router-subgraph": {
+      "entry": ["src/mappings/*.ts", "tests/*.test.ts"]
```
gabmontes, line 17: Can we merge these two?

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function handleRequestFulfilled(event: RequestFulfilledEvent): void {
+  const request = Request.load(event.params.requestId.toString())
+  if (request == null) {
+    return
+  }
+  request.status = 'FULFILLED'
+  // These are the actual shares the user received
+  request.amountOut = event.params.amountIn
```
gabmontes, line 32: The variable rename here is weird.

.knip.json
gndelia: No, it gives a warning because not all the subgraphs have tests on them, so it gives a "hint" to review the entries because `test/*` does not exist

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
gndelia: This is the code that's executed prior to this event


```js
function _handleRequestFulfillment(uint256 requestId_, address tokenIn_, uint256 amountIn_) private {
        RouterStorage storage $ = _getRouterStorage();
        Request storage _r = $.requests[requestId_];

        if (_r.asset == address(0)) revert RequestNotFound(requestId_);
        if (_r.status != Status.PENDING) revert InvalidStatus(requestId_, _r.status);

        _r.status = Status.FULFILLED;

        if (_r.kind == Kind.DEPOSIT) {
            address _share = $.assetsData[_r.asset].share;
            if (tokenIn_ != _share) revert InvalidTokenIn(tokenIn_, _share);
            // HERE
            _r.shares = amountIn_;
        }
        if (_r.kind == Kind.REDEEM) {
            if (tokenIn_ != _r.asset) revert InvalidTokenIn(tokenIn_, _r.asset);
            // HERE
            _r.assets = amountIn_;
        }

        emit RequestFulfilled(requestId_, amountIn_);

        if (_r.automatic) {
            _claim(_r.receiver, requestId_);
        }
    }
```

so basically the contract names it `amountIn_`, but the real meaning is the amount of shares (or assets) that will be sent to the user once they claim (in another tx / event)

As I unified them under one entity, I got `amountOut` suggestion, but I'm open to a different name. I'd like to avoid using 2 fields, making one `null` depending on the entity though. Any suggestions?

gabmontes: Why not keeping it as `amountIn`? How is `amountIn` that is set from `assets` on entity creation different? Do we ned both? At the end of the day, this is mostly driven by curiosity, not an actual concern or request for change.

gndelia: There's already an amountIn in the Request - it's the amount the user entered. Say for Deposit:

The user says "I want to deposit 1 hemiBTC". We record this as `amountIn`. The router gets that, does some async process. Eventually, the fulfillment says "the user gets 0.99 vetBTC). That is what's in the snippet above in the line `_r.shares = amountIn_;`. So the contract event names it `amountIn_;`, but to me that looks a bit incorrect.

Now, as I'm using a `Request` entity, I can't use `shares`, because for redeems the process is the other way around: A user initiates a `redeem` by entering the amount of _shares_, and then, after some async processes, the other line in the snippet above executes: `_r.assets = amountIn_;`. So again `amountIn_` comes from the event, but it's the amount of assets the user will get of the redeem process.

`amountOut` seems the safest note for both. We should also note that the Redeem process may change as the exit cooldown period needs to be added here

Thoughts?

gndelia: And yes, we need both (IMHO) because the user sends some tokens, but as there are async processes, rates may change, so you are not guaranteed to get the exact conversion

gabmontes (approved):

ArturDolzan (approved):
🔥

PR #1942, opened 2026-05-25
Update hemi-earn for new contract interface

ArturDolzan:
### Description

This PR updates the `hemi-earn-actions` package and its portal consumers to track the latest Router/Agent contract iteration, and along the way closes a couple of gaps that were going to bite us
  soon.

  The bulk of the change is a **contract-API migration**: `Status.UNDONE` is now `CANCELLED` (with the matching event, handler and error renames), `requestDeposit`/`requestRedeem` thread the new
  `operator` field end-to-end (encoders, action wrappers and the six portal call sites), and `quoteDepositFulfilment` is renamed to `Fulfillment` to match the contract spelling fix. Without these
  the package's ABI no longer matches what's on chain.

  While in there I added **claim/recover action wrappers** for the Router (claimDeposit, claimRedeem, recoverDeposit, recoverRedeem) with the same event-emitter + receipt orchestration pattern as
  `requestDeposit`, plus matching encoders for gas estimation. The ABIs were already in the package but had no wrappers, so any Activity-page work would have had to reimplement the plumbing inline.

  The third piece is a **bug fix for the staked-balance card** on the Earn home, which wasn't refreshing after a deposit or withdraw until a manual page reload. Root cause was
  `queryClient.ensureQueryData` inside `fetchEarnPositions` returning cached data even after `onSettled` invalidated the prefix (`revalidateIfStale: true` only schedules a background prefetch).
  Switched the inner balance reads to `fetchQuery`, which actually refetches synchronously when stale. The fetcher's tests were updated to mock `balanceOf` directly since the seeded-cache trick no
  longer covers the new behavior.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.


gabmontes (dismissed):
LGTM

portal/app/[locale]/hemi-earn/_fetchers/fetchEarnPositions.ts
```diff
@@ -50,10 +50,18 @@ export const fetchEarnPositions = async function ({
...
+  // marks these reads as stale but doesn't evict them. `ensureQueryData`
+  // returns stale cache even with `revalidateIfStale: true` (the latter only
+  // schedules a background prefetch), so the staked-balance card stayed on
+  // the pre-deposit value until a hard refresh. `fetchQuery` actually waits
+  // for fresh data when the query is stale.
   const settled = await Promise.allSettled(
     shares.map(share =>
       queryClient
-        .ensureQueryData(
+        .fetchQuery(
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_fetchers/fetchEarnPositions.ts]: I think here we should keep `ensureQueryData`

If any validation is needed from another operation, that operation should take care of firing the invalidation. I think fetchers should read from cache, and we should avoid leaving stale data if we know from mutation data is stale

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchEarnPositions.ts]: Makes sense. Fetcher as cache reader is the right separation. fixed by reverting the fetcher back to `ensureQueryData` and switching the mutations' `onSettled` to use `removeQueries` for the `earnPositionsKeyPrefix` instead of `invalidateQueries`. the nested share-balance entries are inactive queries, so the prefix invalidation wasn't refetching them; evicting forces a clean re-read on the next access. comment in the fetcher updated to explain the contract. Applied on b489308d2bb05032e3bbe19d712c250c7bdbaf5f. Thanks!

gndelia (approved):

arielmergen (approved):

PR #1949, opened 2026-05-26
Make Router subgraph mapping idempotent

ArturDolzan:
### Description

This PR fixes a bug where Hemi Earn deposits and redeems showed up as PENDING in the subgraph even after the on-chain fulfillment landed.

  The lifecycle handlers (Fulfilled / Claimed / Cancelled / Recovered) used to early-return when the originating DepositRequested or RedeemRequested hadn't been processed yet. That implicit
  assumption breaks in two real scenarios: the local anvil mock collapses the cross-chain round-trip into a single tx and emits RequestFulfilled inside requestDeposit before the final emit
  DepositRequested, and a subgraph (re)indexed with a startBlock past a request's creation block never sees the original Deposit event. In both cases the lifecycle event got silently dropped and
  the request was stranded in PENDING forever.

  The mapping is now order-agnostic. Lifecycle handlers create a placeholder Request on demand with sentinel values for the required fields, and the request-creation handlers
  (DepositRequested/RedeemRequested) populate the remaining fields when they arrive. The status field is owned exclusively by the lifecycle handlers, so the request-creation path never clobbers a
  more advanced state.

### Screenshots

Captura de Tela 2026-05-27 às 00 18 11

### Related issue(s)

Related to #1848

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
+++ b/subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
@@ -1,3 +1,5 @@
+import { BigInt, Bytes, ethereum } from '@graphprotocol/graph-ts'
+
 import {
   DepositRequested as DepositRequestedEvent,
   RedeemRequested as RedeemRequestedEvent,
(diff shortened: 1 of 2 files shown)


subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
```diff
@@ -8,62 +10,114 @@ import {
...
+  if (request == null) {
+    request = new Request(id)
+    request.requestId = requestId
+    // Sentinels for required fields — overwritten by Deposit/Redeem when the
+    // request-creation event arrives (later in same tx, or never if missed).
+    request.kind = 'DEPOSIT'
+    request.asset = Bytes.empty()
+    request.amountIn = BigInt.zero()
+    request.receiver = Bytes.empty()
+    request.automatic = false
```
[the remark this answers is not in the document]
ArturDolzan: Applied on 5d36bdcb11d986b088288db8c93959526f826733. Thanks!

gabmontes (dismissed):
LGTM

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
```diff
@@ -8,62 +10,118 @@ import {
...
 }
 
 export function handleRedeemRequested(event: RedeemRequestedEvent): void {
   const id = event.params.requestId.toString()
-  const request = new Request(id)
+  let request = Request.load(id)
+  if (request == null) {
+    request = new Request(id)
+    request.status = 'PENDING'
+  }
```
gndelia: shouldn't we use everywhere this 

```suggestion
const request = loadOrInitRequest(id)
```

?

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
```diff
@@ -8,62 +10,118 @@ import {
...
+  if (request == null) {
+    request = new Request(id)
+    request.requestId = requestId
+    // Sentinels for required fields — overwritten by Deposit/Redeem when the
+    // request-creation event arrives (later in same tx, or never if missed).
+    // `asset` and `receiver` are address-shaped in practice (the schema types
+    // them as `Bytes!` for storage); use the 20-byte zero address so that
+    // downstream consumers (e.g. portal calls to `getAddress(asset)`) keep
+    // working on placeholder rows instead of choking on `0x`.
+    request.kind = 'DEPOSIT'
```
gndelia: if we use this for both deposit / redeems, you may want to pass it as a parameter?

Otherwise, it should be `loadOrInitDepositRequest` right?

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
ArturDolzan: You're correct. Applied on aed657f701281297ccefb91d7956758494b46268. Thanks!

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
ArturDolzan: Applied on aed657f701281297ccefb91d7956758494b46268. Thanks!

gabmontes (approved):

gndelia (approved):

Issue #1425, opened 2025-08-05
Unify portal-api and subgraph-api into one


gndelia:
### Feature or change description

There are 2 APIs: The portal-api has all of the endpoints, except the Subgraph method calls, that are in the subgraph-api. I think we should merge both into `portal-api`, so we don't need to maintain a duplicated structure. One of the issues is that they run in different code (JavaScript vs Typescript)

### Problem or use case addressed

There are 2 APIs, duplicating the infra, CI processes, etc

### Proposed implementation

A possible plan that may be done in different PRs could be:

- [ ] Configure Typescript in the `portal-api`. Here, it is key to use the option `"allowJs": true` in the `tsconfig.json` so no actual code needs to be updated. The `Dockerfile` will need to run with `tsx` (_*_)
- [ ] Move code from `subgraph-api` to `portal-api`. Here, all the endpoints may be below `/subgraphs` pathname (or another defined name). Secrets and vars may need to be moved.
- [ ] At this point, the consumers of the subgraphs API (portal, metrics, etc) need to be updated to call the `portal-api`. Watch out for CORS issues
- [ ] Once the consumers are using the `portal-api`, we can remove and tear down the subgraph-api infrastructure

(_*_) From this point onwards, at any moment, the JavaScript code from the original `portal-api` can be migrated to Typescript if desired. Ideally, we can use a local package sharing types with the portal...

gndelia:
This is completed. Issues to tear down are created in the appropriate repo, and do not depend on us. Nothing to do here. Closing

PR #1950, opened 2026-05-27
Migrate portal-backend/api to ESM + TypeScript

gndelia:
### Description

Migrates `portal-backend/api` from CommonJS JavaScript to ESM + TypeScript, run directly via Node's native type-stripping (no bundler — `tsc` is typecheck-only). Source files become `.ts`, modules switch to named `import`/`export`, factory functions are renamed to `create*`, and Sentry now initializes via `--import` before Express loads, following the [Sentry ESM installation guide](https://docs.sentry.io/platforms/javascript/guides/node/install/esm/) (instrumentation must run before the instrumented modules are imported, which the old inline `require` couldn't guarantee under ESM).

Because the api imports the `ve-hemi-actions` / `ve-hemi-rewards` workspace packages, and Node can't type-strip `.ts` under `node_modules`, each package gains a `build` script and a `compiled` export condition pointing at its `dist`; the runtime opts in with `--conditions=compiled`. The api image therefore builds from the repo root as a multi-stage build that compiles those packages and ships only the slim `pnpm deploy` output. `docker-compose` and both CI workflows (`build-push-images`, `docker-checks`) pass the repo root as context with an explicit dockerfile/versionFile, and a `.dockerignore` plus a `packages/*/dist` gitignore entry are added.

**Commits:**

- 39b7d922 Migrate portal-backend/api to ESM + TypeScript
- c0daf894 Build portal-backend/api image from repo root

### Screenshots

N/A — backend-only change, no UI.

### Related issue(s)

Related to #1425

### Checklist

- [ ] Manual testing passed.


portal-backend/api/server.ts
```diff
@@ -0,0 +1,100 @@
...
+  res.status(500).send({ error: 'Internal Server Error' })
+}
+app.use(errorHandler)
+
+const port = config.get<number>('port')
+app.listen(port, function () {
+  const version = config.get<string>('version')
+  console.log(`Portal backend v${version} running on port ${port}`)
+  console.debug(`Config: ${JSON.stringify(config.util.toObject())}`)
+})
```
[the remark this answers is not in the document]
gabmontes: @gndelia I think we can remove that debug line. It has a purpose back then but given the deployment workflow works fine, we don't need it any longer.

.github/workflows/build-push-images.yml
gabmontes, line 50: What about building these other images from the root?

.github/workflows/docker-checks.yml
```diff
@@ -5,10 +5,18 @@ on:
...
+      # would be dependency-aware via `pnpm --filter "...[ref]"`, but that needs
+      # a dynamic-matrix rework — tracked for a future PR/issue.
+      - 'packages/ve-hemi-actions/**'
+      - 'packages/ve-hemi-rewards/**'
   push:
     paths:
       - 'portal-backend/**'
       - 'subgraph-api/**'
+      - 'packages/ve-hemi-actions/**'
+      - 'packages/ve-hemi-rewards/**'
```
gabmontes, line 17: Sort alphabetically on top of the other two?

.github/workflows/docker-checks.yml
```diff
@@ -24,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - run: docker build --tag ${{ matrix.image }}:local ${{ matrix.context }}
+      - run: docker build --tag ${{ matrix.image }}:local ${{ matrix.dockerfile && format('-f {0} ', matrix.dockerfile) || '' }}${{ matrix.context }}
```
gabmontes, line 35: A space may be missing in the expression. When `matrix.dockerfile` and `matrix.context` are both defined, the two expressions will result in a single string.

portal-backend/api/src/ve-hemi/index.ts
```diff
@@ -105,11 +105,8 @@ module.exports = function ({ cache }) {
    * As there would be no rewards for future epochs, the function averages the
    * rewards per day for the past 60 days and assumes that reward level will
    * continue for the 60 epochs.
-   *
-   * @param {"43111"|"743111"} chainId
-   * @returns {Promise<number[]>}
    */
-  async function getVeHemiRewards(chainId) {
+  async function getVeHemiRewards(chainId: string) {
```
gabmontes: Should we keep the stricter type `"43111"|"743111"` here?

portal-backend/api/server.ts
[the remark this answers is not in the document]
gndelia: Fixed in [b9c039d3](https://github.com/hemilabs/ui-monorepo/pull/1950/commits/b9c039d3ad5ddd77c028386004af83779ddb5f58). The startup `console.debug` that serialized the full resolved config (incl. `SENTRY_DSN`, `TVL_DUNE_API_KEY`, Redis URL) has been removed.

.github/workflows/build-push-images.yml
gndelia: Only `portal-backend/api` needs the repo root as context — it bundles the `ve-hemi-actions` / `ve-hemi-rewards` workspace packages via `pnpm deploy`. The cron images and `subgraph-api` are self-contained (their Dockerfiles only `COPY` their own folder), so keeping each scoped to its own context avoids sending the whole monorepo as build context. Opened #1951 to decide whether to standardize on root context for consistency.

.github/workflows/docker-checks.yml
gndelia: Fixed in [a52139a7](https://github.com/hemilabs/ui-monorepo/pull/1950/commits/a52139a75f612cb5e6cf784f83137c55d7c282a6). Sorted the `paths` entries alphabetically (`packages/` ahead of `portal-backend/` and `subgraph-api/`) in both `docker-checks.yml` and `build-push-images.yml`.

.github/workflows/docker-checks.yml
gndelia: The separator is already there — it's the trailing space inside `format('-f {0} ', ...)`. When `dockerfile` is set: `…:local ` + `-f <dockerfile> ` + `<context>` → `…:local -f <dockerfile> <context>` (single spaces). When it's unset, the expression is `''` and the literal space after `:local` separates the tag from the context. Both branches produce a valid command, so no space is missing.

portal-backend/api/src/ve-hemi/index.ts
gndelia: Fixed in [f52fbf39](https://github.com/hemilabs/ui-monorepo/pull/1950/commits/f52fbf392d1335cf308f05f246ad83c7b594111f). Restored `chainId: '43111' | '743111'` — it matches the route regex `/\/ve-hemi-rewards\/(7?43111)/` and stays assignable through the `any`-typed middleware boundary.

gabmontes (approved):

ArturDolzan (approved):

PR #1953, opened 2026-05-28
Add My earn transactions table on hemi-earn

ArturDolzan:
### Description

This PR implements the "My earn transactions" table on /hemi-earn (deposit MVP). It merges subgraph-indexed deposits with in-flight entries we track in localStorage so the user gets feedback the moment they sign the tx, well before the indexer catches up. Rows open a drawer with the deposit details, and failed entries get a Retry CTA that reuses the existing deposit flow.

  The hook layer is split three ways. `useEarnTransactionsQuery` owns the raw subgraph query with conditional polling (every 10s while anything's in flight, off otherwise). `useEarnTransactions` builds on it to expose the merged subgraph + local view for the table and drawer. `useEarnDeliveryWatcher` mounts exactly once via the hemi-earn layout and drives two side effects off the shared subscription... marking local entries settled once the subgraph indexes them, and invalidating Vetro-side balance caches (vault TVL, user pool position, staked-balance card) at the moment a deposit transitions in-flight => CLAIMED/RECOVERED in the subgraph. Single mount point keeps the invalidation from firing N times per consumer.

  This also fixes a timing bug in `useDeposit`. The hook used to fire optimistic updates on pool TVL, user position, and staked-balance card at `deposit-tx-succeeded`... but those values only actually move after LayerZero delivers and the share is minted minutes later. Wallet balance and gas still update at request-deposit mined (real on-chain changes); everything else now waits for the watcher.

  Small enum addition: `TX_PENDING` (portal-internal) to distinguish a local deposit whose request-deposit tx is still in flight from a subgraph `PENDING` (mined and indexed). The two were conflated and the drawer rendered "completed" before mining.

  After talking to @dvnahuel, we'll be adding a second "Cross-chain delivery/ claim CTA" step to both deposit drawers (and the withdraw drawers when that flow lands) to represent the bridging phase explicitly. When `automatic === false`, that step turns into a Claim CTA the user signs themselves, with its own gas fee disclosure. TODOs are placed in `historicalDepositReview.tsx` and `reviewDeposit.tsx`; not in this PR.

### Screenshots

**Main flow**

https://github.com/user-attachments/assets/da57ba5c-2416-4c8b-96ed-c005c749d3ab

**TX Fail**

https://github.com/user-attachments/assets/f891d0b8-119f-4860-adaa-c5a51eac3934

**Mobile**

https://github.com/user-attachments/assets/175a8b34-8ac8-441a-8547-8976753341ba

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const fetchEarnTransactions = async function ({
+  account,
+}: {
+  account: Address
+  networkType: string
+}): Promise<EarnTransaction[]> {
+  await new Promise(resolve => setTimeout(resolve, MOCK_DELAY_MS))
+  return buildMockTransactions(account)
+}
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts:60]: It will be replaced soon. No worries.

portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts
```diff
@@ -0,0 +1,60 @@
...
+      automatic: true,
+      claimTxHash: sampleTxHash,
+      initiatedAt: String(now - 60 * 60 * 6),
+      initiateTxHash: sampleTxHash,
+      kind: 'DEPOSIT',
+      receiver,
+      recoverTxHash: null,
+      requestId: '6',
+      status: 'CLAIMED',
+    },
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts:48]: It will be replaced soon. No worries.

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  const resolved = resolvePoolAndAsset(pools, transaction.asset)
+  if (!resolved) {
+    return (
+      <Drawer onClose={close}>
+        <div className="drawer-content" />
+      </Drawer>
+    )
+  }
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx:53]: Tracked in #1954 — will be addressed in a follow-up PR.

portal/app/[locale]/hemi-earn/_components/transactionsSection/depositSuccessToast.tsx
```diff
@@ -0,0 +1,85 @@
...
+      if (latchedKey === undefined) return undefined
+      const timer = setTimeout(() => setLatched(null), TOAST_MS)
+      return () => clearTimeout(timer)
+    },
+    [latchedKey],
+  )
+
+  if (!latched) {
+    return null
+  }
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/depositSuccessToast.tsx:95]: Good catch! Fixed on d2097b8cb4fee57c69fb142fd43d7ff425434bcf. Thanks.

gndelia:
@dvnahuel @ArturDolzan 

After the user makes the deposit, the operation appears as "In progress" in the table. There's a view button that opens the following drawer:

image


I think there are a few possible improvements

1. Deposit is not _complete_. It's in _progress_ as the table says! Perhaps we could tweak the wording, like "hemiBTC sent" or something like that
2. A follow-up step is missing here. Even though the user may not need to do anything (due to the `automatic` flag), it's doing something: _waiting_. So perhaps there should be a second step "waiting for X", which is everything that needs to happen in the background. I think here we could do something similar to Deposits in the tunnel where the user waits for the funds in the destination chain

image

Of course, we may want to be careful with the wording here as I think users don't really know (or care) about the bridging, but they are _Waiting_ for something.

Let me know your thoughts


EDIT: Oh, is that the 2nd step you mentioned to add? Lol I did not get it at first reading

portal/app/[locale]/hemi-earn/_components/earnStatusUpdaters/index.tsx
```diff
@@ -0,0 +1,9 @@
+import { EarnDepositsStatusUpdater } from './earnDepositsStatusUpdater'
+
+// Mounted in the hemi-earn layout. Mirrors the `<TunnelStatusUpdaters>`
+// composition pattern in `portal/components/tunnelStatusUpdaters/index.tsx`:
+// invisible watcher components that drive cross-route polling and
+// query-cache invalidations. Today there's only a deposit updater; withdraw
+// and any claim/recover-specific watchers will compose here without
+// restructuring.
+export const EarnStatusUpdaters = () => <EarnDepositsStatusUpdater />
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/earnStatusUpdaters/index.tsx]: Why can't we do here:

```suggestion
export const EarnStatusUpdaters = function () {
  useEarnDeliveryWatcher()
  return null
}
```

instead of adding a level of mis-direction? Am I missing something? Or is it because a `withdraw` updated (as you mention) will be added? In that case, then why can't we do

```suggestion
export const EarnStatusUpdaters = function () {
  useEarnDeliveryWatcher()
  useWithdarwWatcher()
  return null
}
```

?

portal/app/[locale]/hemi-earn/_components/icons/depositedCheckIcon.tsx
```diff
@@ -0,0 +1,10 @@
+export const DepositedCheckIcon = () => (
+  <svg fill="none" height="16" width="16" xmlns="http://www.w3.org/2000/svg">
+    <path
+      clipRule="evenodd"
+      d="M8 15A7 7 0 1 0 8 1a7 7 0 0 0 0 14Zm3.844-8.791a.75.75 0 0 0-1.188-.918l-3.7 4.79-1.649-1.833a.75.75 0 1 0-1.114 1.004l2.25 2.5a.75.75 0 0 0 1.15-.043l4.251-5.5Z"
+      fill="#A3A3A3"
+      fillRule="evenodd"
+    />
+  </svg>
+)
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/icons/depositedCheckIcon.tsx:20]: Is this icon

Image

?

I think we have this icon in other parts of the app. Maybe we can find it and reuse it? (Switching colors can be done with tailwind..)

gndelia:
While loading the (I pressume) allowance, the submit button shows as disabled

image


That's ok. But in the tunnel, we render the Circular spinner while doing so

image


In general, when the submit button is disabled due to some async check (for example, loading allowance or balances to check), we should show this instead of the text IMHO

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
```diff
@@ -0,0 +1,143 @@
...
+> = {
+  CANCELLED: (tx, t) => ({
+    description: <span>{t('step.deposit-cancelled')}</span>,
+    status: ProgressStatus.FAILED,
+    txHash: tx.initiateTxHash,
+  }),
+  CLAIMED: (tx, t) => ({
+    description: <span>{t('step.deposit-completed')}</span>,
+    status: ProgressStatus.COMPLETED,
+    txHash: tx.claimTxHash ?? tx.initiateTxHash,
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx]: wouldn't this link to an incorrect transaction?


```suggestion
    txHash: tx.claimTxHash,
```

`txHash` is optional in `StepPropsWithoutPosition`

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }),
+  PENDING: (tx, t) => ({
+    description: <span>{t('step.deposit-completed')}</span>,
+    status: ProgressStatus.COMPLETED,
+    txHash: tx.initiateTxHash,
+  }),
+  RECOVERED: (tx, t) => ({
+    description: <span>{t('step.deposit-recovered')}</span>,
+    status: ProgressStatus.COMPLETED,
+    txHash: tx.recoverTxHash ?? tx.initiateTxHash,
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx]: Similar to https://github.com/hemilabs/ui-monorepo/pull/1953/changes#r3320168233


```suggestion
    txHash: tx.recoverTxHash,
```

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx
```diff
@@ -0,0 +1,100 @@
...
+      emitter.on('unexpected-error', () => setOperationRunning('failed'))
+      emitter.on('user-signed-deposit', function (newInitiateTxHash) {
+        setTxDrawerQueryString(newInitiateTxHash)
+      })
+    },
+    pool,
+    selectedAsset: asset,
+    // Hide the specific failed row from the table once this retry is signed.
+    supersedesInitiateTxHash: transaction.initiateTxHash,
+    updateDepositOperation() {},
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx]: 👀

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/useTxDrawerQueryString.ts
```diff
@@ -0,0 +1,18 @@
...
+    'earnTxId',
+    parseAsString.withOptions({ clearOnDefault: true }),
+  )
+
+  const setTxDrawerQueryString = function (next: string | null) {
+    setTxId(next)
+  }
+
+  return { setTxDrawerQueryString, txId }
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/useTxDrawerQueryString.ts]: if you ask me, it's simpler to just...


```suggestion
export const useTxDrawerQueryString = function () {
  return useQueryState(
    'earnTxId',
    parseAsString.withOptions({ clearOnDefault: true }),
  )
```

but it's as you prefer. If you think this reads better (using an object), then it's ok for me

portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
```diff
@@ -0,0 +1,104 @@
...
+      </span>
+    ),
+    header: () => <Header text={t('column.date')} />,
+    id: 'date',
+    meta: { className: 'justify-start flex-grow-0', width: 140 },
+  },
+  {
+    cell: ({ row }) => (
+      <span className="text-neutral-900">
+        {row.original.kind === 'DEPOSIT' ? t('deposit') : ''}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx]: I must assume "withdrawals" go here ?


```suggestion
        {row.original.kind === 'DEPOSIT' ? t('deposit') : t('withdrawal')}
```

perhaps it's easier to add them now so this is complete already 😅

portal/app/[locale]/hemi-earn/_components/transactionsSection/depositSuccessToast.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  return (
+    <PoolToast
+      chainId={hemi.id}
+      key={latched.startedAt}
+      title={t('deposit-successful')}
+      transactionHash={latched.initiateTxHash}
+    />
+  )
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/depositSuccessToast.tsx:105]: I think I understand why you implemented this like this, but I wonder if an imperative call from an event (like `on('deposit-transaction-succeeded', () => showToast(deposit))`) would make this simpler. Food for thought, we can revisit later

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+      upsertLocalOperation,
+    ],
+  )
+
+  return (
+    <LocalEarnOperationsContext.Provider value={value}>
+      {children}
+    </LocalEarnOperationsContext.Provider>
+  )
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:187]: [use-local-storage-state](https://github.com/hemilabs/ui-monorepo/blob/30618c1639380d042e02dc5062a9fd841f247216/portal/package.json#L63) is a dependency already installed - it may facilitate you syncing of state and local storage. Was not used for any particular reason?

portal/app/[locale]/hemi-earn/_components/transactionsSection/compactCell.tsx
```diff
@@ -0,0 +1,15 @@
+import { type ComponentProps } from 'react'
+
+// Mirrors `components/table/_components/column.tsx` exactly except for the
+// row min-height: `min-h-12` instead of `min-h-16` so the row honours the
+// `rowSize={48}` passed to `<Table>`. Keeping the rest (incl. `flex-grow`)
+// identical guarantees body cells align vertically with the header columns
+// (which use `ColumnHeader` with its own `flex-grow`).
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/compactCell.tsx]: What if you use the original Colum with `style` property?

```tsx
<Column style={{ minHeight: <whatever-value-min-h-12-is> } />
```

we'd save duplicating the entire component. The original `Column` already accepts `style` so..

portal/app/[locale]/hemi-earn/layout.tsx
```diff
@@ -14,6 +17,11 @@ const Layout = function ({ children }: Props) {
     return <NotFound />
   }
 
-  return <>{children}</>
+  return (
+    <LocalEarnOperationsProvider>
+      <EarnStatusUpdaters />
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/layout.tsx:22]: Note that this won't update any earn status if the user is in a different page (No updates in the background if the user is, say, in the tunnel page).

If you want that, you may want to move this higher in the chain, or use web workers as we do for the tx history of the tunnel..

portal/app/[locale]/hemi-earn/types.ts
```diff
@@ -1,10 +1,60 @@
...
+  // `TX_PENDING` flags a local deposit whose request-deposit tx is still
+  // in flight (signed, not yet mined). It's distinct from subgraph
+  // `PENDING` so the drawer can render "in progress" before mining and
+  // "completed" once mined, without overloading a single status.
+  //
+  // `FAILED` flags a local deposit whose approve/deposit tx reverted (or
+  // whose wallet prompt the user rejected). The retry CTA in the drawer
+  // is what flips this back to an in-flight status.
+  | 'TX_PENDING'
+  | 'FAILED'
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/types.ts:20]: the coments are kind of disordered against the enum list. It's a bit confusing to read. I'd also either add comments to all of the states or none.

portal/app/[locale]/hemi-earn/types.ts
```diff
@@ -44,3 +94,36 @@ export type EarnPosition = {
...
+  kind: 'DEPOSIT' | 'REDEEM'
+  operation: DepositOperation | WithdrawOperation
+  operator?: Address
+  settled?: boolean
+  shareAddress: Address
+  // Unix seconds. Matches the unit of `TTL_SECONDS` in
+  // `localEarnOperationsContext.tsx` — if this ever changes to ms, the GC
+  // comparison there must change too.
+  startedAt: number
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/types.ts]: Some typing improvement to disallow invalid states is:

```ts

```
```suggestion
type LocalEarnOperation = {
  account: Address
  amountIn: string
  asset: Address
  chainId: Chain['id']
  initiateTxHash?: Hash
  kind: 'DEPOSIT' | 'REDEEM'
  operation: DepositOperation | WithdrawOperation
  operator?: Address
  settled?: boolean
  shareAddress: Address
  // Unix seconds. Matches the unit of `TTL_SECONDS` in
  // `localEarnOperationsContext.tsx` — if this ever changes to ms, the GC
  // comparison there must change too.
  startedAt: number
}

export type EarnDepositOperation = LocalEarnOperation & {
   kind: 'DEPOSIT'
   operation: DepositOperation
}

export type EarnRedeemOperation = LocalEarnOperation & {
   kind: 'REDEEM'
   operation: WithdrawOperation
}

```

This way you can ensure at Typescript level invalid states do not exists, like `kind: DEPOSIT` and `operation: WithdrawOperation` at the same time.

A guard type function may help you, like

```ts
const isDepositOperation = (operation: EarnDepositOperation | EarnRedeemOperation): operation is EarnDepositOperation => operation.kind === 'DEPOSIT'

// then..

if (isDepositOperation(operation)) {
   // here Typescript ensures you kind is 'Deposit' and operation is DepositOperation
} else { 
  // here is Redeem
}
```

USeful for doing some `ifs` in the table and call the appropiate component with the appropiate types.

Follow-up questions:

1.  Why `WithdrawOperation` and not `RedeemOperation`
2. Why _Local_EarnOperation ?

portal/messages/en.json
```diff
@@ -24,7 +24,7 @@
         "title": "Composition",
         "tokens-count": "{count, plural, one {# token} other {# tokens}}"
       },
-      "deposit-successful": "Deposit successful",
+      "deposit-successful": "Deposit submitted",
```
gndelia: key and value mismatch

```suggestion
      "deposit-submitted": "Deposit submitted",
```

gndelia (commented):
This PR was large to review 😅 

Sounds like some features could've been split into a different PR - for example, watches/udpaters are not needed to render the table.

Let's keep an eye on that for incoming PRs 😃

portal/app/[locale]/hemi-earn/types.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/types.ts]: Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd — split into `LocalEarnDepositOperation` + `LocalEarnWithdrawOperation` with a `LocalEarnOperation` union and exported the `isLocalEarnDeposit` guard. Used it to drop the `as DepositOperation` casts in `useEarnTransactions.localStatus` and `depositSuccessToast.isDepositConfirmedAndUnsettled`.

On the follow-ups:

1. **`WithdrawOperation` vs `RedeemOperation`** — they live at different layers. `WithdrawOperation` is the local UI-state type (with `WithdrawStatus` enum: APPROVAL_TX_PENDING / WITHDRAW_TX_* / etc.), naming aligned with the user-facing "Withdraw" tab. `kind: 'REDEEM'` mirrors the subgraph schema and the on-chain function name (ERC-4626 `redeem`). Each layer keeps its own terminology.

2. **Why "Local" prefix** — disambiguates from `EarnTransaction` in the same file. They model different things: `EarnTransaction` is a row coming back from the subgraph (post-indexing), `LocalEarnOperation` is the optimistic in-flight mirror this browser persists in localStorage. Without the `Local`, the two would visually conflate.

ArturDolzan:
> This PR was large to review 😅
> 
> Sounds like some features could've been split into a different PR - for example, watches/udpaters are not needed to render the table.
> 
> Let's keep an eye on that for incoming PRs 😃

Yes, I agree 😅
I initially thought about splitting some of it, especially the watchers/updaters part, but ended up bundling everything together to keep the flow connected.

Will definitely keep that in mind for upcoming PRs

portal/app/[locale]/hemi-earn/_components/earnStatusUpdaters/index.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/earnStatusUpdaters/index.tsx]: Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/_components/icons/depositedCheckIcon.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/icons/depositedCheckIcon.tsx:20]: Good point. Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

ArturDolzan:
> While loading the (I pressume) allowance, the submit button shows as disabled
> 
> image
> That's ok. But in the tunnel, we render the Circular spinner while doing so
> 
> image
> In general, when the submit button is disabled due to some async check (for example, loading allowance or balances to check), we should show this instead of the text IMHO

Good point! Applied on Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. See the video below:


https://github.com/user-attachments/assets/4d124057-8513-4f88-a85a-5d021dceb51c

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx]: Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx]: Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx]: I missed that. Fixed on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/useTxDrawerQueryString.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/useTxDrawerQueryString.ts]: I agree.. applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx]: Indeed! Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/_components/transactionsSection/depositSuccessToast.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/depositSuccessToast.tsx:105]: Yep, this is something to think about.. but state-driven was deliberate so the toast survives mid-window navigation (watcher mounts at the layout level, not the route) and fires on a second tab where localStorage syncs the entry without an emitter firing on that tab. Imperative would be cleaner per call site but would lose both. Happy to revisit once we see if those scenarios actually pull their weight.

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:187]: I missed that lib and it simplify a lot. Nice! Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/_components/transactionsSection/compactCell.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/compactCell.tsx]: It works! Applied on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/layout.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/layout.tsx:22]: Indeed.. deliberate scoping for the deposit flow..  cross-chain delivery is ~1–3 min and users typically wait inside /hemi-earn during that window. If they navigate away, the row reconciles to whatever the subgraph reports on next mount, no data loss; we just lose the live toast.

Withdraw will likely force the conversation since cooldown is 7 days and users definitely won't sit on the page. Plan A is bumping the watcher up to `app/[locale]/layout.tsx` alongside `<Workers>` when that flow lands; the watcher hook internals don't change. Plan B is the worker pattern you're using for the tunnel tx history if cross-tab/offline coverage becomes a requirement. We can definitely revisite this.

portal/app/[locale]/hemi-earn/types.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/types.ts:20]: Simplified on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/messages/en.json
ArturDolzan: Fixed on 4bbb34716b250296c7ce6d3a01a778ac4269f9bd. Thanks!

portal/app/[locale]/hemi-earn/_components/transactionsSection/compactCell.tsx
```diff
@@ -0,0 +1,6 @@
+import { Column } from 'components/table/_components/column'
+import { type ComponentProps } from 'react'
+
+export const CompactCell = (props: ComponentProps<'td'>) => (
+  <Column {...props} style={{ minHeight: '3rem', ...props.style }} />
+)
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/compactCell.tsx]: Is it cell or column?


```suggestion
export const CompactColumn = (props: ComponentProps<'td'>) => (
  <Column {...props} style={{ minHeight: '3rem', ...props.style }} />
)
```

portal/app/[locale]/hemi-earn/types.ts
Comment (gndelia) [reply portal/app/[locale]/hemi-earn/types.ts]: Do we need the local? In general, indexing is fast, and we haven't had this issue of "the subgraph hasn't indexed yet". Are you deleting from local storage the entries that already appear on the subgraph? Are they needed once they come from the subgraph?

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
```diff
@@ -0,0 +1,199 @@
...
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:187]: I'll ask again. Do we know if this is a problem that needs to be solved? Perhaps I'd start, as we do with other tables, just reading from the subgraph + optimistic updates, and see from there how it goes. My experience is that in 3/4 seconds at most, the subgraph indexes data so... 

To me, this seems to add complex code for a problem we may not even face. We may want to revisit it once the subgraph and contracts are deployed and see from there

Note that local fork may work different from real RPCs with the subgraph..

Comment (gndelia) [reply portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:187]: It's different if you want to use the cache as for faster readings from the subgraph. We did that for the tx history in the tunnel.

But there, we wouldn't have to drop data to read it from the package, so I'm a bit confused on what's the flow expected here 🤔 

Even there, we also used that because we had to reconcile data from multipe chains (L1 and L2, or bitcoin and hemi).

If we abstract away most of what the Agent is doing on Ethereum, and we just show "Pending", perhaps we can make it simpler here... The tx history has its own complexities for example. I am unsure if I would implement it as I did then 😅 


Thoughts?

dvnahuel:
> While loading the (I pressume) allowance, the submit button shows as disabled
> 
> image
> That's ok. But in the tunnel, we render the Circular spinner while doing so
> 
> image
> In general, when the submit button is disabled due to some async check (for example, loading allowance or balances to check), we should show this instead of the text IMHO

Agreed!

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:187]: Fair on the success path.. agree that if that were the only thing the local store solved, in-memory + optimistic would beat the complexity. The load-bearing case is FAILED: the subgraph indexes `RequestDeposit` events from successful txs only, so a reverted deposit never reaches it. Without local persistence there's no row to show in the table and no way to drive the retry CTA from the drawer opened by the table.. so the user would have to navigate back to the pool page to retry, losing the asset/amount/share context.

Thoughts?

portal/app/[locale]/hemi-earn/_components/transactionsSection/compactCell.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/compactCell.tsx]: Done on 6d1abb33ba6a18278fdc637a809cb643ac2c3658. Thanks!

portal/app/[locale]/hemi-earn/types.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/types.ts]: The reconcile flips entries to `settled: true` rather than deleting outright.. soft delete. The merge filters those out (`!op.settled`), so they aren't visible once the subgraph row takes over. You're right that we don't need them after that; they sit in localStorage until the 90-day TTL (or the 100-entry per-account cap) collects them. Easy to flip to hard delete though. Thoughts?

dvnahuel:
A couple of points of improvements:

- The spinner should be animated
Screenshot 2026-05-29 at 11 21 17 AM

- On responsive, when the layouts is left-aligned like we can see in the screenshot, let's make the gray container with the token logos full width and the items inside of it center-align
Screenshot 2026-05-29 at 11 22 31 AM

- Let's make the label order in which we showcase the table on mobile views to the following [ Action - Status - Type - Amount - Date ] 
Screenshot 2026-05-29 at 11 22 48 AM

ArturDolzan:
> A couple of points of improvements:
> 
> * The spinner should be animated
> 
> Screenshot 2026-05-29 at 11 21 17 AM
> * On responsive, when the layouts is left-aligned like we can see in the screenshot, let's make the gray container with the token logos full width and the items inside of it center-align
> 
> Screenshot 2026-05-29 at 11 22 31 AM
> * Let's make the label order in which we showcase the table on mobile views to the following [ Action - Status - Type - Amount - Date ]
> 
> Screenshot 2026-05-29 at 11 22 48 AM

Suggestions applied on bd6cfe872f6aaf8d5b530859df2d7597655ac9a6. Thanks @dvnahuel! I also added some videos below:

Spinner:

https://github.com/user-attachments/assets/d973b921-8d2a-45b7-8032-635b3d8c3931



Mobile:


https://github.com/user-attachments/assets/542ba299-dbd6-4caa-a8c6-186ed1f6cde1

gndelia (dismissed):

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
Comment (gabmontes) [reply portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:187]: Following up on @gndelia's question about whether settled entries need to be kept at all —

They don't. The soft delete (`settled: true`) doesn't add value over hard delete: once the flag is set the entry is filtered out everywhere, so keeping it in localStorage just grows the store. Hard deleting on reconcile gives the same table behavior with simpler code.

Concretely:

- `reconcileLocals` removes the entry instead of flipping `settled: true`
- `markSettledByInitiateTxHash` does the same — rename to `removeLocalOperation` to reflect intent
- Drop the `settled` field from `LocalEarnOperationBase` and remove the `!op.settled` filters in `useEarnTransactions`, `useEarnTransactionsQuery`, and `depositSuccessToast` — the toast's `seenOnMountRef` guard still works since a hard-deleted entry won't be in the store on next mount
- Reduce TTL from 90 to 7 days — the only entries that outlive reconciliation are `FAILED` ones (reverted txs that never reach the subgraph), and 7 days is a generous retry window for those

Net result: simpler types, smaller store, same correctness.

Comment (gabmontes) [reply portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:187]: Thinking about reorgs, hard deleting on reconcile behaves identically to soft delete if the subgraph rolls back — the row disappears either way. On Hemi PoS, reorgs are 1–2 blocks and the subgraph re-indexes within seconds, so the row reappears on its own after the next poll. If you happen to catch the brief window where it's gone, a page reload recovers it. Deep reorgs that permanently exclude a tx are not a realistic scenario on PoS.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts
```diff
@@ -222,6 +214,10 @@ export const useDeposit = function ({
         updateDepositOperation({
           status: DepositStatus.DEPOSIT_TX_FAILED,
         })
+        upsertLocalOperation({
+          ...baseLocalPayload,
+          operation: { status: DepositStatus.DEPOSIT_TX_FAILED },
+        })
```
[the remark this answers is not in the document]
Comment (gabmontes) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts:224]: The `deposit-transaction-succeeded` and `deposit-transaction-reverted` event handlers upsert without `initiateTxHash`, so the match falls back to `startedAt`. In the normal single-deposit path this works — the `user-signed-deposit` handler already stored the hash, and the merge preserves it. But two deposits started within the same second share the same `account + kind + startedAt` key, so the second handler's upsert could update the wrong entry.

The fix is one line per handler — `transactionHash` is on the receipt:

```ts
upsertLocalOperation({
  ...baseLocalPayload,
  initiateTxHash: receipt.transactionHash, // add this
  operation: { status: DepositStatus.DEPOSIT_TX_CONFIRMED },
})
```

Same for the `deposit-transaction-reverted` handler. Makes the canonical-key match unambiguous regardless of timing.

portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts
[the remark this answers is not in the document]
Comment (gabmontes) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts:60]: Can you open an issue to track the removal of the mock delay and the placeholder data before this goes to production? "Will be replaced soon" is easy to forget.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts:1]: `useDeposit` now calls `useLocalEarnOperations()` unconditionally, which throws if the hook is used outside `<LocalEarnOperationsProvider>`. Both current call sites (the pool page and the home retry drawer) are under the `hemi-earn` layout where the provider lives, so this is fine today. But nothing at the type level prevents future reuse outside that boundary — it will throw at runtime, not at compile time.

Can you add a comment next to the `useLocalEarnOperations()` call noting the context dependency, so future callers are aware? Something like:

```ts
// Requires <LocalEarnOperationsProvider> in the tree (mounted at the hemi-earn layout).
const { markSettledByInitiateTxHash, upsertLocalOperation } =
  useLocalEarnOperations()
```

Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts:1]: `updateDepositOperation` is now optional, which makes `useDeposit` usable outside the pool form context (the retry flow). That's intentional. But it means that when called without it — e.g. from `<RetryFailedDeposit>` — approval failures are silently dropped: no `updateDepositOperation` call, and no `upsertLocalOperation` either (since there's no `initiateTxHash` at the approval stage).

Is that the intended behavior? If an approval reverts during a retry, the user gets no feedback in the table and the original FAILED row has already been soft-deleted. Should the retry flow handle approval failures separately, or is the assumption that the wallet/toast layer covers it sufficiently?

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts:1]: Applied on 774f9cf63a71224ee37021a95f1724d5e9dd8f57. Thanks!

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts:1]: Original FAILED row isn't soft-deleted (actually now it was changed to a hard-delete) on approval revert... The supersede / `markSettledByInitiateTxHash` only fires inside the `user-signed-deposit` handler (`useDeposit.ts:158`). If approval reverts before the user signs the deposit tx, that handler never runs and the original FAILED row stays in the table. The user can retry again.

No table row for approval reverts is intentional... Local entries are only created at `user-signed-deposit`, that's the first moment we have an `initiateTxHash` to key on. Before that (any approval phase), we don't touch the local store. So approval reverts produce no row, both in the first-time flow and in retry. Feedback during approval comes from the wallet plus the local component state (`operationRunning: 'failed'` flips the button back from "Depositing" to "Try again").

You're right that the retry drawer doesn't reinforce the approval revert with an inline message.. that's a UX gap worth closing eventually. Happy to track it as a follow-up issue. Thoughts?

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:187]: Makes sense.. applied. Soft-delete (`settled: true`) is gone; reconcile and retry supersede both hard-delete via `deleteLocalOperationByInitiateTxHash`, the `settled` field is removed from the type, and the `reconciledRef` dedup in the watcher fell out as dead weight (entries gone from the store don't reappear in the loop). See 598ffacd.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useDeposit.ts:224]: Agreed... `receipt.transactionHash` is now passed on both `deposit-transaction-succeeded` and `deposit-transaction-reverted` upserts. `matchesExisting` keys on the canonical hash, so the same-second collision is closed. See 8529a6d1. Thanks for pointing this out!

portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts:60]: Absolutely! Issue created https://github.com/hemilabs/ui-monorepo/issues/1963

gndelia (dismissed):
uACK

arielmergen (dismissed):

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
```diff
@@ -0,0 +1,187 @@
...
+  useMemo,
+} from 'react'
+import useLocalStorageState from 'use-local-storage-state'
+import { type Address, type Hash } from 'viem'
+import { useAccount } from 'wagmi'
+
+import { type LocalEarnOperation } from '../types'
+
+const STORAGE_KEY = 'hemi-earn:local-operations'
+const TTL_SECONDS = 90 * 24 * 60 * 60
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx]: `TTL_SECONDS` is still `90 * 24 * 60 * 60`. With hard delete in place, the only entries that survive past reconciliation are `FAILED` ones (reverted txs). 90 days is a long accumulation window for dead entries — 7 days is a more reasonable retry ceiling.

— Auto Gabi

portal/app/[locale]/hemi-earn/_components/earnStatusUpdaters.tsx
```diff
@@ -0,0 +1,17 @@
+'use client'
+
+import { useEarnDeliveryWatcher } from '../_hooks/useEarnDeliveryWatcher'
+
+// Invisible watcher mounted in the hemi-earn layout. Drives cross-route
+// polling and Vetro-side cache invalidation off a single subscription —
+// `useEarnDeliveryWatcher` flips local entries to `settled` once the
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_components/earnStatusUpdaters.tsx]: This file-level comment still says `useEarnDeliveryWatcher` "flips local entries to `settled`" — that was the soft-delete behaviour, removed in 598ffacd. Should now read something like "removes local entries once the subgraph indexes them".

Same stale wording in `useEarnTransactions.ts`: `// Side effects (marking local entries 'settled', ...)` — both need updating now that it's a hard delete.

— Gabi 🤖

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx]: Good point. Applied on 80ac0e8bda7001947121b5eca46ca7e44190e5be. Thanks!

portal/app/[locale]/hemi-earn/_components/earnStatusUpdaters.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/earnStatusUpdaters.tsx]: Good catch! Fixed on 80ac0e8bda7001947121b5eca46ca7e44190e5be. Thanks!

gabmontes (approved):

gndelia (approved):

PR #1971, opened 2026-06-01
Update get-started page layout

### LINKED ISSUE #1709: Update the 'Get Started' page designs to better match the new portal look and content

arielmergen:
### Description

Reworks the Portal **Get Started** page into three focused steps and introduces a new Hemi tokens table, to better match the new portal look and content.

**Get Started page**
- Restructures the page into three steps: add the Hemi network to your wallet, add Hemi tokens, and learn how to use Hemi.
- Removes the old "Fund your wallets" and "Start using Hemi" sections and their assets/icons (faucets, MoonPay/Oku/Sushi, deploy-contract, etc.).
- Trims the tutorials down to the relevant ones (set up EVM/BTC wallet, tunnel assets to Hemi) and refreshes the wallet/tunnel illustrations.
- Improves spacing, alignment, and responsive layout across mobile, tablet, and desktop (full-width network cards and tabs on small screens).

**Add Hemi network**
- Makes the RPC and block explorer URLs clickable links.
- Guards the wallet client before adding the chain and refactors the automatic/manual add-network components.

**Add Hemi tokens (new)**
- New searchable table listing each Hemi token with its L1 and L2 addresses and an "Add token" action.
- Adds ERC-20 tokens to the wallet with the correct decimals/symbol and surfaces success/error feedback via toasts (success toast auto-closes).
- Aligns the table with the search input and lets the body shadow render correctly on mobile.

**Shared components & i18n**
- Small tweaks to the Table (header/column/body) and Toast components and related icons to support the new flows.
- Removes stale get-started translation keys and adds the new ones (en/es/pt).

### Screenshots
getstarted-desktop

get-started-mobile

### Related issue(s)

Closes #1709

### Checklist

- [ ] Manual testing passed.


portal/hooks/useHemiTokens.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  return useMemo(
+    () =>
+      tokenList.tokens.filter(
+        (token): token is EvmToken =>
+          isEvmToken(token) && token.chainId === hemi.id,
+      ),
+    [hemi.id],
+  )
+}
```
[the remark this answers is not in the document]
arielmergen: Applied.

portal/messages/en.json
```diff
@@ -340,30 +347,38 @@
...
       "learn-to-setup-evm-wallet": "Learn how to set up and get started with an EVM (MetaMask) wallet.",
-      "learn-to-tunnel-to-hemi": "Learn how to tunnel (bridge) your ETH assets to and from Hemi.",
+      "learn-to-tunnel-to-hemi": "Learn how to tunnel assets from Ethereum and Bitcoin to Hemi.",
       "pop-miner": "PoP Miner",
       "run-cli-pop-miner": "Run the CLI PoP Miner",
-      "set-up-btc-wallet": "Set Up a BTC Wallet",
+      "set-up-btc-wallet": "Set up a BTC wallet",
       "set-up-cli-pop-miner": "Set up and run your PoP miner, either by downloading pre-built binaries or by building from source.",
-      "set-up-evm-wallet": "Set Up an EVM Wallet",
+      "set-up-evm-wallet": "Set up a EVM wallet",
```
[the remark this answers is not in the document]
arielmergen: Applied.

portal/app/[locale]/get-started/_utils/getL1BridgePair.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (!network) {
+    return null
+  }
+
+  const bridgeAddress = bridgeInfo[l1ChainIdStr]?.tokenAddress
+  const address =
+    !bridgeAddress ||
+    isAddressEqual(bridgeAddress, NativeTokenSpecialAddressOnL2)
+      ? getNativeToken(l1ChainId).address
+      : bridgeAddress
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_utils/getL1BridgePair.ts]: Applied.

portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx
```diff
@@ -32,21 +34,16 @@ export const AddChainButton = function ({ chain }: Props) {
...
-    return t('add-to-wallet')
-  }
+  const label = isMutating ? t('adding-chain') : t('add-to-wallet')
 
   return (
-    <button className="hoverable-text cursor-pointer" type="button">
-      {getText()}
-    </button>
+    <div className="shrink-0">
+      <Button size="xSmall" variant="secondary">
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx]: Applied.

portal/app/[locale]/get-started/_components/addChainManually.tsx
```diff
@@ -18,50 +19,41 @@ const DataSection = ({
...
+      <ChainHeaderRow chain={chain} layer={layer} />
+      <DataSection
+        clickableLink={false}
+        label={`${t('rpc-url')}:`}
+        value={chain.rpcUrls.default.http[0]}
+      />
+      <DataSection label={`${t('chain-id')}:`} value={chain.id.toString()} />
+      <DataSection
+        label={`${t('currency-symbol')}:`}
+        value={chain.nativeCurrency.symbol}
```
Comment (gabmontes) [inline portal/app/[locale]/get-started/_components/addChainManually.tsx:50]: The block-explorer URL `DataSection` now has `clickableLink={false}`, making it render as plain text instead of a clickable `ExternalLink`. In the original code the default was `true`. The only other field with `clickableLink={false}` was the RPC URL, which has a known reason to be non-clickable. Worth confirming this is intentional and matches the Figma spec before merging.

— Gabi 🤖

portal/components/table/_components/virtualRows.tsx
```diff
@@ -18,6 +19,7 @@ export function VirtualRows<TData>({
   loading,
   onRowClick,
   onRowHover,
+  rowClassName = 'group/row absolute flex w-full items-center',
```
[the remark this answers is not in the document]
arielmergen: Applied

portal/app/[locale]/get-started/_components/addChainManually.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChainManually.tsx:50]: > The block-explorer URL `DataSection` now has `clickableLink={false}`, making it render as plain text instead of a clickable `ExternalLink`. In the original code the default was `true`. The only other field with `clickableLink={false}` was the RPC URL, which has a known reason to be non-clickable. Worth confirming this is intentional and matches the Figma spec before merging.
> 
> — Gabi 🤖

@gabmontes confirmed against Figma (16981:30961, Manual tab): both RPC URL and Block Explorer URL are plain neutral-500 text with a copy icon, not clickable links. `clickableLink={false}` for the block explorer is intentional and matches the spec. The previous default (`true`) was pre-redesign behavior on main.

@dvnahuel, do you have any comments or concerns about this approach?

portal/app/[locale]/get-started/_components/addChain/networkConfigCard.tsx
```diff
@@ -0,0 +1,11 @@
+import { ComponentProps } from 'react'
+
+export const NetworkConfigCard = ({
+  className = '',
+  ...props
+}: ComponentProps<'div'>) => (
+  <div
+    className={`rounded-lg bg-white p-4 shadow-[0px_0px_0px_1px_rgba(0,0,0,0.05)] ${className}`}
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/addChain/networkConfigCard.tsx]: I think that shadow should come from the standard shadows defined in the tailwind config. Check figma as it usually tells you which is by name

portal/app/[locale]/get-started/_components/tokenTable/addressCell.tsx
```diff
@@ -0,0 +1,40 @@
+import { RemoteChain } from 'types/chain'
+import { EvmToken } from 'types/token'
+
+import { formatTokenAddress } from '../../_utils/formatTokenAddress'
+import { getL1BridgePair } from '../../_utils/getL1BridgePair'
+
+type Props = {
+  address: string
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/tokenTable/addressCell.tsx]: To better keep the types in sync

```suggestion
  address: EvmToken['address']
```

portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx
```diff
@@ -32,21 +34,16 @@ export const AddChainButton = function ({ chain }: Props) {
...
-      return t('adding-chain')
-    }
-
-    if (accountStatus === 'disconnected') {
-      return tCommon('connect-wallet')
-    }
-
-    return t('add-to-wallet')
-  }
+  const label = isMutating ? t('adding-chain') : t('add-to-wallet')
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx]: Now, when the user is disconnected, it reads "Add to Wallet". Does it connect + adds to wallet automatically? What happens if the user is disconnected and clicks this button?

portal/app/[locale]/get-started/_utils/formatTokenAddress.ts
```diff
@@ -0,0 +1,8 @@
+import { formatEvmAddress } from 'utils/format'
+import { isNativeAddress } from 'utils/nativeToken'
+import { type Address, isAddress } from 'viem'
+
+export const formatTokenAddress = (address: string) =>
+  isNativeAddress(address) || !isAddress(address)
+    ? address
+    : formatEvmAddress(address as Address)
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_utils/formatTokenAddress.ts]: This is redundant and can be removed

```suggestion
    : formatEvmAddress(address)
```

portal/app/[locale]/get-started/_components/section.tsx
```diff
@@ -6,28 +6,32 @@ import { Step } from './step'
...
+            {children}
           </div>
-          {children}
-        </div>
-      </Card>
+        </Card>
+      </div>
     ) : (
-      children
+      <div className="pt-[68px]">{children}</div>
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/section.tsx]: I think there's a class already for this


```suggestion
      <div className="pt-18">{children}</div>
```

portal/app/[locale]/get-started/_components/step.tsx
```diff
@@ -1,22 +1,28 @@
...
+  <div className="flex size-full flex-col justify-center text-center text-[11px] font-semibold lining-nums tabular-nums leading-[0] tracking-[0.11px] text-white">
+    <span className="leading-[11px]">{position}</span>
+  </div>
+)
+
 export const Step = ({ description, position }: Props) => (
-  <div className="mt-1">
-    <div className="left-2.25 relative -translate-y-0.5">
-      <ShortVerticalLine stroke="stroke-neutral-300" />
+  <div className="relative h-[68px] w-5 shrink-0">
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/step.tsx]:
```suggestion
  <div className="relative h-18 w-5 shrink-0">
```

Comment (gndelia) [reply portal/app/[locale]/get-started/_components/step.tsx]: (or `h-17`, double check)

portal/app/[locale]/get-started/_components/section.tsx
Comment (gndelia) [reply portal/app/[locale]/get-started/_components/section.tsx]: (or `pt-17`, double check)

portal/app/[locale]/get-started/_components/step.tsx
```diff
@@ -1,22 +1,28 @@
...
+import { LongVerticalLine } from 'components/verticalLines'
 
 type Props = {
-  description: string
+  description?: string
   position: number
 }
 
+const StepNumber = ({ position }: { position: number }) => (
+  <div className="flex size-full flex-col justify-center text-center text-[11px] font-semibold lining-nums tabular-nums leading-[0] tracking-[0.11px] text-white">
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/step.tsx]: I think a few classes here can be dropped in favor of `text-xxs`

gndelia (changes requested):
Let's review the translation files as many files were dropped from the old implementation, and texts were updated, but I don't see those keys being removed (For example, the ones from `startUsingHemi.tsx`)

portal/app/[locale]/get-started/_components/tokenTable/column.tsx
```diff
@@ -0,0 +1,13 @@
...
+  className,
+  ...props
+}: ComponentProps<'td'>) => (
+  <td
+    className={`flex h-full min-h-16 w-full min-w-0 flex-grow items-center border-b border-solid border-neutral-100 py-2.5 first:[&>*]:pl-4 last:[&>*]:pr-4 ${
+      className ?? ''
+    }`}
+    {...props}
+  />
+)
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/tokenTable/column.tsx]: Why can't we use the `Column` component that exists already?

portal/app/[locale]/get-started/_components/addHemiToken.tsx
```diff
@@ -0,0 +1,160 @@
...
+    id: 'l1-address',
+    meta: { width: 147 },
+  },
+  {
+    cell: ({ row }) => (
+      <TokenTableAddressCell
+        address={row.original.address}
+        networkName={hemi.name}
+      />
+    ),
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/addHemiToken.tsx]:
```suggestion
    cell: function ({ row }) {
    const hemi = useHemi()
    return (
      <TokenTableAddressCell
        address={row.original.address}
        networkName={hemi.name}
      />
    )
    },
```

This way, you no longer need the `useHemi` as part of the `useMemo` nor the `tokenTableColumns` fn

portal/app/[locale]/get-started/_components/addHemiToken.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  const columns = useMemo(
+    () => tokenTableColumns({ hemi, remoteNetworks, t }),
+    [hemi, remoteNetworks, t],
+  )
+
+  const tokens = useMemo(
+    () => filterTokensBySearch(hemiTokens, debouncedSearchText),
+    [debouncedSearchText, hemiTokens],
+  )
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/addHemiToken.tsx:113]: debouncing makes sense if, after typing, we're requesting some information asynchronously, to avoid firing lots of requests - here I would just filter directly as we type, as all information is in the client side

portal/app/[locale]/get-started/_components/learnMore.tsx
```diff
@@ -2,135 +2,88 @@ import { AnalyticsEvent } from 'app/analyticsEvents'
...
-        <Box
-          event="tut - add hemi"
-          heading={t('add-themi-wallet')}
-          href="https://docs.hemi.xyz/how-to-tutorials/tutorials/add-themi-to-metamask"
-          subheading={t('learn-to-add-themi-wallet')}
-        />
-      )}
-    </div>
+    <ExternalLink
+      className="flex h-28 overflow-hidden rounded-lg bg-white shadow-[0px_0px_0px_1px_rgba(0,0,0,0.05)] transition-colors hover:bg-neutral-50"
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/learnMore.tsx]: the shadow here should come from the tailwind config theme. Review it - there's already shadows defined

portal/components/table/index.tsx
```diff
@@ -40,14 +42,16 @@ type TableHeaderProps<TData> = {
 
 const TableHeader = <TData,>({
   hasVerticalBodyScrollbar,
+  headerClassName,
+  headerRowClassName,
```
gndelia: Can we double-check why these were added? I don't see them used in the get started's table

portal/components/table/index.tsx
```diff
@@ -102,13 +111,15 @@ type TableBodyProps<TData> = {
...
   CellComponent,
+  bodyContainerClassName,
   fetchMoreOnBottomReached,
   headerScrollRef,
   isFetching,
   loading,
   onRowClick,
   onRowHover,
   placeholder,
+  rowClassName,
```
gndelia, line 115: Can we double-check why these were added? I don't see them used in the get started's table

portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx
```diff
@@ -32,21 +34,16 @@ export const AddChainButton = function ({ chain }: Props) {
...
 
   return (
-    <button className="hoverable-text cursor-pointer" type="button">
-      {getText()}
-    </button>
+    <div className="shrink-0">
+      <Button disabled={isMutating} size="xSmall" variant="secondary">
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx]:
```suggestion
      <Button disabled={isMutating} size="xSmall" type="button" variant="secondary">
```

This was removed, but it shouldn't

portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx]: Disconnected
openDrawer() → opens the connect-wallet drawer. addChain is not called.

Connected, network not added
walletClient.addChain({ chain }) → wallet prompt to add the network. Label changes to "Adding...".

Connected, network already added or already on that chain
Shows "Added ✓" (no meaningful click action).

The label is isMutating ? "Adding..." : "Add network". We kept it that way because Figma shows a single label ("Add network") for this button, so we didn't branch on isConnected.

When the wallet isn't connected, clicking the row opens the connect drawer (openDrawer() in AddChain) rather than calling addChain — the user would need to click again after connecting to add the network. If we think that's confusing in practice, we could switch the disconnected label to something like "Connect Wallet". 

@dvnahuel, what do you think?

portal/app/[locale]/get-started/_components/addChain/networkConfigCard.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChain/networkConfigCard.tsx]: Applied in a4b6df1.

portal/app/[locale]/get-started/_components/tokenTable/addressCell.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/tokenTable/addressCell.tsx]: Applied in 2773163.

portal/app/[locale]/get-started/_utils/formatTokenAddress.ts
Comment (arielmergen) [reply portal/app/[locale]/get-started/_utils/formatTokenAddress.ts]: Applied in 473cba5.

portal/app/[locale]/get-started/_components/section.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/section.tsx]: Applied as `pt-17` in 473cba5.

portal/app/[locale]/get-started/_components/step.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/step.tsx]: Applied as `h-17` in 473cba5.

portal/app/[locale]/get-started/_components/step.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/step.tsx]: The step number already uses `text-xxs`. The remaining classes (`font-semibold`, `lining-nums`, `tabular-nums`, `text-white`) are orthogonal to the font size token — happy to drop something specific if you can point it out.

portal/app/[locale]/get-started/_components/tokenTable/column.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/tokenTable/column.tsx]: Refactored `Column` with a `variant` prop (`default` | `plain`) and removed `GetStartedTableColumn`. Get Started's table now uses `Column` with `variant="plain"`. See 228f686 and 473cba5.

portal/app/[locale]/get-started/_components/addHemiToken.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addHemiToken.tsx]: Applied in 473cba5 — extracted `HemiAddressCell` so `useHemi` lives inside the cell and `tokenTableColumns` no longer needs the `hemi` param.

portal/app/[locale]/get-started/_components/addHemiToken.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addHemiToken.tsx:113]: Applied in 473cba5 — removed `useDebounce` and filter directly on `searchText`.

portal/app/[locale]/get-started/_components/learnMore.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/learnMore.tsx]: Applied as `shadow-bs` in 473cba5.

portal/components/table/index.tsx
arielmergen: Good catch — removed `headerClassName`/`headerRowClassName` from `Table` since nothing consumes them. See 473cba5.

portal/components/table/index.tsx
arielmergen: Removed in 473cba5 together with the other unused header props.

portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx]: Restored in b874673.

portal/app/[locale]/get-started/_components/addHemiToken.tsx
```diff
@@ -0,0 +1,160 @@
...
+  return query
+    ? tokens.filter(
+        token =>
+          token.name.toLowerCase().includes(query) ||
+          token.symbol.toLowerCase().includes(query) ||
+          (userTypedAddress &&
+            isAddress(token.address) &&
+            isAddressEqual(token.address, trimmed)),
+      )
+    : tokens
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addHemiToken.tsx:96]: Fixed in 709c563 — store the narrowed `Address` value in `typedAddress` so `isAddressEqual` gets the proper type.

portal/app/[locale]/get-started/_components/addChain/networkConfigCard.tsx
```diff
@@ -0,0 +1,11 @@
+import { ComponentProps } from 'react'
+
+export const NetworkConfigCard = ({
+  className,
+  ...props
+}: ComponentProps<'div'>) => (
+  <div
+    className={`shadow-bs rounded-lg bg-white p-4 ${className}`}
+    {...props}
+  />
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChain/networkConfigCard.tsx:10]: Fixed in a3f53cc — added `?? ''` fallback.

portal/app/[locale]/get-started/_components/addChainManually.tsx
```diff
@@ -18,50 +19,41 @@ const DataSection = ({
...
+  <div className="flex items-center justify-between gap-4 border-b border-neutral-100 py-4 last:border-b-0">
+    <span className="body-text-semibold shrink-0 text-neutral-950">
       {label}
     </span>
     {value.startsWith('https') ? (
-      <div className="flex w-full items-center">
+      <div className="flex min-w-0 items-center justify-end gap-1">
         <ConfigurationUrl clickableLink={clickableLink} href={value} />
       </div>
     ) : (
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChainManually.tsx:30]: This check predates this PR and matches Hemi chains' actual RPC/explorer URLs (all `https`). Not changing it as part of this PR.

portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx
Comment (gndelia) [reply portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx]: > Disconnected
> openDrawer() → opens the connect-wallet drawer. addChain is not called.

ok, but the text says "Add token to wallet", not reflecting the call to action. If the user is disconnected, it should read "Connect wallet" as it did before

portal/app/[locale]/get-started/_components/step.tsx
Comment (gndelia) [reply portal/app/[locale]/get-started/_components/step.tsx]: `text-[11px]`, `leading-[0]` and  `tracking-[0.11px]`, all these 3 are part of `text-xss` which is the 11px size

IF there's a small discrepancy in figma (like tracking or line-height are slightly different), I'd prefer consistency and just use `text-xxs` directly instead of adding custom classes which for the regular eye bear no differece

arielmergen:
@gndelia good catch — cleaned up 41 unused keys (23 from `get-started` and 18 from `get-started.learn-more-tutorials`) across en/es/pt in fe83abe. Restored `add-networks`, `add-networks-mainnet` and `add-networks-testnet` in 7537e80 + df8beaa since they're consumed via dynamic interpolation `add-networks-${networkType}` and as a nested namespace.

portal/app/[locale]/get-started/_components/addChainManually.tsx
```diff
@@ -18,50 +19,41 @@ const DataSection = ({
...
+      <DataSection label={`${t('chain-id')}:`} value={chain.id.toString()} />
+      <DataSection
+        label={`${t('currency-symbol')}:`}
+        value={chain.nativeCurrency.symbol}
+      />
+      <DataSection
+        clickableLink={false}
+        label={`${t('block-explorer-url')}:`}
+        value={chain.blockExplorers!.default.url}
+      />
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChainManually.tsx:55]: Intentional and matches the Figma spec — both RPC URL and Block Explorer URL are plain neutral-500 text with a copy icon in the Manual tab. Same context as @gabmontes' earlier thread.

portal/messages/en.json
```diff
@@ -299,71 +299,45 @@
...
     "rpc-url": "RPC URL",
-    "start-using-hemi": "Start using Hemi",
-    "subheading": "Set up the Hemi network in your wallet to start using the protocol.",
-    "sushi-swap": "Sushi Swap",
-    "swap-tokens": "Swap tokens",
-    "swap-tokens-mainnet": "Swap tokens on Sushi",
-    "swap-tokens-testnet": "Swap tokens on Hemi",
+    "search-tokens-placeholder": "Search tokens...",
+    "token": "Token",
+    "subheading": "Activate compliant Bitcoin yield without leaving custody",
```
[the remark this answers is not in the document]
arielmergen: Fixed in f2867a9.

portal/messages/es.json
```diff
@@ -299,71 +299,45 @@
...
     "rpc-url": "URL de RPC",
-    "start-using-hemi": "Comience a usar Hemi",
-    "subheading": "Configure la red de Hemi en su wallet y comience a interactuar con el protocolo.",
-    "sushi-swap": "Sushi Swap",
-    "swap-tokens": "Intercambie tokens",
-    "swap-tokens-mainnet": "Intercambie tokens en Sushi",
-    "swap-tokens-testnet": "Intercambie tokens en Hemi",
+    "search-tokens-placeholder": "Buscar tokens...",
+    "token": "Token",
+    "subheading": "Active rendimiento de Bitcoin compatible sin renunciar a la custodia",
```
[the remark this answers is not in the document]
arielmergen: Fixed in f2867a9.

portal/messages/pt.json
```diff
@@ -299,71 +299,45 @@
...
     "rpc-url": "URL RPC",
-    "start-using-hemi": "Começar a usar o Hemi",
-    "subheading": "Configure a rede da Hemi na sua wallet e comece a interagir com o protocolo.",
-    "sushi-swap": "Sushi Swap",
-    "swap-tokens": "Trocar tokens",
-    "swap-tokens-mainnet": "Trocar tokens na Sushi",
-    "swap-tokens-testnet": "Trocar tokens no Hemi",
+    "search-tokens-placeholder": "Pesquisar tokens...",
+    "token": "Token",
+    "subheading": "Ative rendimento Bitcoin compatível sem sair da custódia",
```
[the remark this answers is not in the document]
arielmergen: Fixed in f2867a9.

portal/app/[locale]/get-started/_components/addChainManually.tsx
Comment (gabmontes) [reply portal/app/[locale]/get-started/_components/addChainManually.tsx:50]: Let's wait for @dvnahuel to land in BA and provide feedback. If he is ok with these URLs not being links, mark the comment as resolved.

portal/messages/en.json
```diff
@@ -333,71 +299,45 @@
     "your-rewards-are-on-their-way": "Your rewards are on their way!"
   },
   "get-started": {
-    "add-funds-to-your-wallet": "Add funds to your wallet",
+    "add-hemi-network-to-your-wallet": "Add Hemi network to your wallet",
+    "add-hemi-tokens": "Add Hemi tokens",
+    "add-hemi-tokens-description": "Hemi tokens are paired with their ERC-20 versions on Ethereum. To see and use an asset on Hemi, your wallet needs the Hemi (L2) token address.",
     "add-hemi-to-your-wallet": "Add Hemi to your wallet",
```
gndelia: It seems this key was removed from `addHemiWallet.tsx` but it's still kept here. Can we double check?

portal/messages/en.json
```diff
@@ -333,71 +299,45 @@
...
-    "start-using-hemi": "Start using Hemi",
-    "subheading": "Set up the Hemi network in your wallet to start using the protocol.",
-    "sushi-swap": "Sushi Swap",
-    "swap-tokens": "Swap tokens",
-    "swap-tokens-mainnet": "Swap tokens on Sushi",
-    "swap-tokens-testnet": "Swap tokens on Hemi",
+    "search-tokens-placeholder": "Search tokens...",
+    "token": "Token",
+    "subheading": "Activate compliant Bitcoin yield without leaving custody",
     "tunnel-assets": "Tunnel assets",
```
gndelia: This one was removed from `startUsingHemi.tsx` (as the file was dropped), but the translation key is still here. Can we double check?

portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChain/addChainButton.tsx]: Gotcha, Added when the wallet is disconected.

portal/messages/en.json
gabmontes, line 1: The subtitle now ends with a period (`"Activate compliant Bitcoin yield without leaving custody."`). Is this consistent with the Figma design, @dvnahuel?

— Gabi 🤖

dvnahuel:
Let's change the description for each 'step' to the following:

1. **Add Hemi network to your wallet** - This lets your wallet interact with Hemi for transactions, bridges, and dApps.
2. **Add Hemi tokens** - Hemi tokens are paired with their ERC-20 versions on Ethereum. To see and use an asset on Hemi, your wallet needs the Hemi (L2) token address.
3. **Learn how to use Hemi** - New to Hemi? Our step-by-step tutorials walk you through everything you need to know, from setting up your wallet to bridging tokens and using dApps.

@arielmergen

portal/messages/en.json
arielmergen: Good catch — removed in a0daa71. The matching key in `genesis-drop` namespace remains since it's still used there.

portal/messages/en.json
arielmergen: Good catch — removed in a0daa71. The matching key in `tunnel-page.transaction-history` namespace remains since it's still used there.

dvnahuel:
Let's add a higher bottom margin (current one is too tight):

- For mobile: 48px
- For Tablet: 48px
- For desktop: 64px

Current implementation (for reference)
Screenshot 2026-06-03 at 12 41 58 PM

@arielmergen

dvnahuel:
On the first step, switching from manual to automatic generates a layout swap, we should iterate that


https://github.com/user-attachments/assets/96952c69-d7ee-4d9f-84e5-1d2831cf7732

portal/app/[locale]/get-started/_components/addChainManually.tsx
Comment (dvnahuel) [reply portal/app/[locale]/get-started/_components/addChainManually.tsx:50]: Good catch, let's add the link for both. In terms of interactions, let's keep the same as in prod as well.

portal/messages/en.json
dvnahuel: Yes, now all subtitles are going to have a period.

arielmergen:
Done — step 1's description now uses the copy you provided (in 31c1950 + i18n keys in 05d2513). Steps 2 and 3 already matched the text in the existing keys, so no changes needed there.

image

arielmergen:
Added bottom spacing in 793e6f7 — `pb-12 lg:pb-16` (48px on mobile/tablet, 64px on desktop).

arielmergen:
Addressed in 37f3e4e — the chain row in automatic mode now uses `items-start` with the button vertically centered separately, so "Hemi Layer 2" sits at the same top offset as the manual card (~17px instead of ~21px). Let me know if you'd still like to iterate further on the transition between modes.

https://github.com/user-attachments/assets/80de70c1-7254-4c86-af23-ca6f0b84baeb

gndelia:
image

This is not vertically centered

dvnahuel:
For the second step, on the token addresses table the 'add token' derived states should not be render inside the CTA, they should be render on the toast component.

States

1. Default [ + Add token ] —> Token not yet added to wallet.
2. Success or already added [ ✓ Added ] -> Toast: (success icon) Token added to your wallet successfully.
3. Error (any of them) -> Toast: (warning icon) Couldn't add token to your wallet. Please try again.

When the wallet is not connected, the [ + Add token ] CTA should open the wallet drawer.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=18937-23222&t=Z8dyhTUMLHqVHkMa-1)

Screenshot of the proposal:
Screenshot 2026-06-03 at 3 51 01 PM

Current implementation:
Screenshot 2026-06-03 at 12 37 59 PM

dvnahuel:
On tablet views, let's make the 'Hemi l2' card full-width

Screenshot 2026-06-03 at 12 40 06 PM

dvnahuel:
On mobile the segmented control component should be full-width

Screenshot 2026-06-03 at 12 40 39 PM

dvnahuel:
On both tablet and mobile views these cards should be full-width

Screenshot 2026-06-03 at 12 40 16 PM

dvnahuel:
On the second step, in the token addresses table, there's some L1 addresses missing, do you know why that's happening? @gndelia @gabmontes @arielmergen 

Screenshot 2026-06-03 at 12 37 53 PM

dvnahuel:
We should revise the overall table layouts on mobile views:

- The label should be left-align (to be align with the CTA's) and also we should add some more left padding (16px) so that the CTA don't feel cramped with the border of the table.
Screenshot 2026-06-03 at 12 41 15 PM


- There's a lot of white space if you scroll to the right in the table
Screenshot 2026-06-03 at 12 41 27 PM

arielmergen:
> image
> This is not vertically centered

Good catch centered.

Commit [63ece85](https://github.com/hemilabs/ui-monorepo/commit/63ece85574037bee8efd7212a24fd8ec37f343ce)

arielmergen:
> For the second step, on the token addresses table the 'add token' derived states should not be render inside the CTA, they should be render on the toast component.
> 
> States
> 
> 1. Default [ + Add token ] —> Token not yet added to wallet.
> 2. Success or already added [ ✓ Added ] -> Toast: (success icon) Token added to your wallet successfully.
> 3. Error (any of them) -> Toast: (warning icon) Couldn't add token to your wallet. Please try again.
> 
> When the wallet is not connected, the [ + Add token ] CTA should open the wallet drawer.
> 
> Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=18937-23222&t=Z8dyhTUMLHqVHkMa-1)
> 
> Screenshot of the proposal: Screenshot 2026-06-03 at 3 51 01 PM
> 
> Current implementation: Screenshot 2026-06-03 at 12 37 59 PM

Done, added toast to show the state. 

[6a37611beac](https://github.com/hemilabs/ui-monorepo/tree/6a37611beac7ad15f6ad479a352043d2f89232c0)
[6a37611beac](https://github.com/hemilabs/ui-monorepo/tree/6a37611beac7ad15f6ad479a352043d2f89232c0)

https://github.com/user-attachments/assets/84861471-ff81-4bc2-8a20-347e9d2afaa5

arielmergen:
> On the second step, in the token addresses table, there's some L1 addresses missing, do you know why that's happening? @gndelia @gabmontes @arielmergen
> 
> Screenshot 2026-06-03 at 12 37 53 PM

2 reasons:
- Some tokens are pure hemi tokens (no token on Ethereum side)
- Some tokens can be bridged outside of the tunnel (Say Stargate for example). The L1 address is only relevant for our tunnel

portal/app/[locale]/get-started/_components/addChainManually.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChainManually.tsx:50]: Done in 567915dd — removed `clickableLink={false}` from both the RPC URL and Block Explorer URL fields, so both now render as clickable `ExternalLink`s (default `clickableLink={true}`), matching prod interactions. Thanks @dvnahuel.

portal/messages/en.json
arielmergen: Confirmed intentional per @dvnahuel: all subtitles now end with a period. No change needed.

arielmergen:
> On tablet views, let's make the 'Hemi l2' card full-width
> 
> Screenshot 2026-06-03 at 12 40 06 PM

Added 
Captura de pantalla 2026-06-08 a la(s) 12 23 25 p  m

[496c5b04](https://github.com/hemilabs/ui-monorepo/tree/496c5b0459f4b305b02e8d7678362e2f4e9d325b)

arielmergen:
> On mobile the segmented control component should be full-width
> 
> Screenshot 2026-06-03 at 12 40 39 PM

[a2e60e2c](https://github.com/hemilabs/ui-monorepo/commit/a2e60e2c8fa2173c2e57ff94a4fde3d3174d1505)

arielmergen:
> On both tablet and mobile views these cards should be full-width
> 
> Screenshot 2026-06-03 at 12 40 16 PM

Added 

Captura de pantalla 2026-06-08 a la(s) 12 53 12 p  m

Captura de pantalla 2026-06-08 a la(s) 12 53 29 p  m


commit [424d0e3bb](https://github.com/hemilabs/ui-monorepo/tree/424d0e3bb97737226a761dc4351a91ea98ce0cc0)

arielmergen:
> We should revise the overall table layouts on mobile views:
> 
> * The label should be left-align (to be align with the CTA's) and also we should add some more left padding (16px) so that the CTA don't feel cramped with the border of the table.
There's a lot of white space if you scroll to the right in the table
> 

Added

Captura de pantalla 2026-06-08 a la(s) 1 31 55 p  m

Captura de pantalla 2026-06-08 a la(s) 1 33 38 p  m



[3f8ec7a2bbd](https://github.com/hemilabs/ui-monorepo/tree/3f8ec7a2bbd2f778daf839bbf2b1ceed379501b6)

arielmergen:
> Let's review the translation files as many files were dropped from the old implementation, and texts were updated, but I don't see those keys being removed (For example, the ones from `startUsingHemi.tsx`)

 - [fe83abeb](https://github.com/hemilabs/ui-monorepo/tree/fe83abeb9e36f97ca24b6699ff74776dee12b00a) "Remove unused i18n keys" (06-02 16:11)
 - [a0daa711 ](https://github.com/hemilabs/ui-monorepo/tree/a0daa711892321c29e6e43b896449bdaef78aa05)"Remove leftover get-started i18n keys" (06-03 11:37)

portal/components/addTokenToWallet.tsx
```diff
@@ -10,6 +12,7 @@ import { isNativeToken } from 'utils/nativeToken'
 type Props = {
   labels: Record<'error' | 'idle' | 'pending' | 'success', string>
   token: EvmToken
+  variant?: 'button' | 'link'
 }
```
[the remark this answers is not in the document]
arielmergen: Fixed in 5d4e17f. Removed the unused `variant` prop and its button branch — restoring the component to its single-purpose (link) form.

portal/app/[locale]/get-started/_components/tokenTable/addTokenButton.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const isTokenAdded = useWatchedAsset(token.address)
+
+  const [toastKey, setToastKey] = useState(0)
+  const [toastVariant, setToastVariant] = useState<'error' | 'success' | null>(
+    null,
+  )
+
+  const { isPending, mutate, reset, status } = useAddTokenToWallet({
+    token,
+  })
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/tokenTable/addTokenButton.tsx:76]: Addressed in 5d4e17f by removing the premature `variant="button"` from the shared component (it had no call sites). We kept `AddTokenTableButton` as a separate component on purpose: it adds behavior the shared `AddTokenToWallet` doesn't have and that its 3 review-screen call sites don't want — connect-wallet drawer handling, success/error toasts, and an "Added" state. Consolidating would mean expanding the shared component in ways that affect those screens, so a dedicated table action is the cleaner split for now.

gabmontes (dismissed):

portal/app/[locale]/get-started/_components/addChain/index.tsx
```diff
@@ -23,7 +23,7 @@ export const AddChain = function ({ chain, children }: Props) {
   const { enabled, track } = useUmami()
   const { data: walletClient } = useWalletClient()
 
-  const { mutate: addChain } = useMutation({
+  const { isPending, mutate: addChain } = useMutation({
     mutationFn: (c: Chain) => walletClient!.addChain({ chain: c }),
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChain/index.tsx]: Fixed in b5d0bbd. Replaced the non-null assertion with a guard inside `mutationFn`: it now throws if `walletClient` isn't ready, so react-query moves the mutation to an error state instead of hard-crashing (and `onSuccess` won't run, so the chain isn't falsely marked as added).

gabmontes (dismissed):

.gitignore
```diff
@@ -7,6 +7,7 @@
 .vscode/
 lerna-debug.log
 node_modules
+.pnpm-store
```
gabmontes: The `.pnpm-store` directory is not created in standard pnpm setups — it only appears when `--store-dir .pnpm-store` is passed explicitly (or configured via a local `.npmrc`). Neither of those is the case here, so adding this to `.gitignore` would mask the root cause rather than fix it.

Could you check how this directory ended up inside the repo? It might be a one-off manual `pnpm install --store-dir .pnpm-store`, a script, or an IDE integration that passed that flag. The right fix is to remove that flag so the global store is used, not to ignore the symptom.

— Gabi 🤖

arielmergen: You're right. I checked: there's no `store-dir` in `.npmrc` and nothing in the repo (scripts, lerna/turbo configs, CI, IDE settings) passes `--store-dir .pnpm-store`; the global pnpm store is correctly configured (`~/Library/pnpm/store`). The local `.pnpm-store` was a one-off local artifact, never tracked by git. Reverted the .gitignore entry in 1a99137b and removed the stray local directory, so the global store is used as expected.

gabmontes (dismissed):

gndelia (changes requested):
When testing this, I'm unable to add any token

I always get the following toast:

Image

By debugging, I get the following error:

> addTokenButton.tsx:79 InvalidParamsRpcError: Invalid parameters were provided to the RPC method. Double check you have provided the correct parameters.
> 
> Details: Invalid ERC-20 decimals: undefined.
> Version: viem@2.28.4
> (anonymous)	@	addTokenButton.tsx:79


🤔

dvnahuel:
> When testing this, I'm unable to add any token
> 
> I always get the following toast:
> 
> Image
> By debugging, I get the following error:
> 
> > addTokenButton.tsx:79 InvalidParamsRpcError: Invalid parameters were provided to the RPC method. Double check you have provided the correct parameters.
> > Details: Invalid ERC-20 decimals: undefined.
> > Version: viem@2.28.4
> > (anonymous)	@	addTokenButton.tsx:79
> 
> 🤔

The same happens to me

dvnahuel:
On mobile views the table seems to break.
IMG_0074

dvnahuel:
The margin top and bottom padding on this table should be 4px not 16px 
Screenshot 2026-06-09 at 3 46 23 PM

dvnahuel:
This link text should have the same styles as the other texts on the same column (13px) but with the font-weight on medium

Screenshot 2026-06-09 at 3 44 01 PM

arielmergen:
> When testing this, I'm unable to add any token
> 
> I always get the following toast:
> 
> Image
> By debugging, I get the following error:
> 
> > addTokenButton.tsx:79 InvalidParamsRpcError: Invalid parameters were provided to the RPC method. Double check you have provided the correct parameters.
> > Details: Invalid ERC-20 decimals: undefined.
> > Version: viem@2.28.4
> > (anonymous)	@	addTokenButton.tsx:79
> 
> 🤔

Thank you! I've already sent the fix.

dvnahuel:
The table text labels should be in 13px size.

Screenshot 2026-06-10 at 3 05 10 PM

arielmergen:
> On mobile views the table seems to break. IMG_0074

**I've tested it, and I don't seem to be able to reproduce it.**
Captura de pantalla 2026-06-10 a la(s) 3 06 44 p  m

arielmergen:
Fix the margin top and bottom padding on this table; it should be 4px, not 16px
Captura de pantalla 2026-06-10 a la(s) 4 29 29 p  m

Table text labels
Captura de pantalla 2026-06-10 a la(s) 4 27 30 p  m

portal/hooks/useAddTokenToWallet.ts
```diff
@@ -33,10 +33,12 @@ export const useAddTokenToWallet = function (options: Options) {
         {
           address: token.address as Address,
           chainId: token.chainId,
+          decimals: token.decimals,
           // token logos include the small Hemi logo, but wallets crop it.
```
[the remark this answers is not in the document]
arielmergen: This isn't a high-severity issue, and the described failure mode doesn't
  actually occur:
  
  The mutationFn is async and runs inside useMutation. If walletClient is
  undefined and watchAsset throws, React Query catches it and transitions the
  mutation to status === 'error' — which is exactly the "clean transition to
  error state" the comment says wouldn't happen. The caller already handles
  this:

  // addTokenButton.tsx
  } else if (status === 'error') {
    setToastVariant('error')
    ...
  }

  So there's no unhandled runtime crash — the user gets an error toast.

  On top of that, the scenario is a very narrow race: the button only calls
  mutate() when isConnected is true (otherwise it opens the connect drawer and
  returns), and by then useWalletClient has resolved in practically all real
  cases.

  Finally, this is pre-existing code — the only change in this PR on this line
  is adding decimals: token.decimals. The walletClient/address usage is
  untouched.
  
  Adding an explicit guard for a clearer error message would be a minor
  nice-to-have, but it's not blocking and not high severity.

gndelia:
image

The icons and the text are not properly aligned

arielmergen:
> image
> The icons and the text are not properly aligned

Fixed

Captura de pantalla 2026-06-12 a la(s) 11 52 00 a  m

arielmergen:
@copilot resolve the merge conflicts in this pull request

gndelia (dismissed):

portal/components/toast/index.tsx
```diff
@@ -78,7 +105,7 @@ export const Toast = function ({
           </Link>
         )}
       </div>
-      <button className="size-5" onClick={() => setClosedToast(true)}>
+      <button className="size-5 shrink-0" onClick={() => setClosedToast(true)}>
         <CloseIcon className="size-full [&>path]:hover:stroke-white" />
       </button>
```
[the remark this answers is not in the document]
arielmergen: Addressed in `f82c4e98`: the close button now uses `[&>path]:hover:fill-white` (the `CloseIcon` paths are fill-based) and has `aria-label="Close notification"`.

portal/app/[locale]/get-started/_components/learnMore.tsx
```diff
@@ -2,135 +2,88 @@ import { AnalyticsEvent } from 'app/analyticsEvents'
 import { ButtonLink } from 'components/button'
 import { Card } from 'components/card'
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/learnMore.tsx:3]: This doesn't apply here: the parent `get-started/page.tsx` is already marked `'use client'`, so `LearnMore` (and `TutorialCard`) are part of the client module graph. It compiles fine and both `useUmami` and the `onClick` handlers work — verified running locally with no error. The repo also doesn't follow a convention of marking every hook-using subcomponent with its own `'use client'`; several siblings (e.g. `configurationUrl.tsx`) rely on the page-level boundary too. Resolving as not applicable.

gndelia (dismissed):

portal/app/[locale]/get-started/_components/addChainManually.tsx
```diff
@@ -18,50 +19,39 @@ const DataSection = ({
...
-          label={t('currency-symbol')}
-          value={chain.nativeCurrency.symbol}
-        />
-        <DataSection
-          label={t('block-explorer-url')}
-          value={chain.blockExplorers!.default.url}
-        />
-      </div>
-    </div>
+    <NetworkConfigCard>
```
Comment (gabmontes) [inline portal/app/[locale]/get-started/_components/addChainManually.tsx:40]: The old code passed `clickableLink={false}` explicitly to this row to keep the RPC URL as plain text. That argument was dropped in this refactor, so `ConfigurationUrl` now falls back to its `clickableLink={true}` default — making the RPC URL a clickable external link.

Was that intentional? The block-explorer URL becoming clickable was a deliberate fix, but the RPC URL changing silently is a different case. The `DataSection` `clickableLink` prop is still wired through to `ConfigurationUrl`, so passing it back would restore the previous behaviour.

— Gabi 🤖

dvnahuel:
LGTM!

portal/app/[locale]/get-started/_components/addChainManually.tsx
Comment (arielmergen) [reply portal/app/[locale]/get-started/_components/addChainManually.tsx:40]: Good catch — making the RPC URL clickable wasn't intentional. Restored the previous behaviour by passing `clickableLink={false}` to the RPC `DataSection`, so it's plain text + copy again, while the block-explorer URL stays a clickable link as intended. Fixed in 4fe2192c.

gndelia (approved):

gabmontes (approved):

PR #1972, opened 2026-06-02
Add cross-chain step to deposit drawers

ArturDolzan:
### Description

This PR adds a "Get share tokens" step after the existing Stake step in both deposit drawers (pool page + transactions table), surfacing the cross-chain delivery window between the
  request-deposit mining on Hemi and the share OFTs landing on the user's wallet. The step renders NOT_READY while approval and deposit signing are in flight, PROGRESS after the deposit mines, and
  COMPLETED once the subgraph reports CLAIMED/RECOVERED. There's no CTA on it yet — manual claim (FULFILLED with automatic=false) and recovery (CANCELLED) are scoped for a follow-up.

  The approval step is rendered in both drawers whenever an approval hash exists. On the pool drawer it's the live signing flow — the drawer opens on user-signed-approval and the row goes from
  PROGRESS to COMPLETED as the approval tx mines (same shape as the tunnel pattern). On the historical drawer (opened from the transactions table) it's rendered as already-COMPLETED whenever the
  row carries an approvalTxHash. Since the indexer can't link an approval tx to a request, the hash lives in the per-browser local store; useEarnTransactions joins it onto the subgraph row in the
  merge, so the same drawer enriches historical entries originated on this device.

  To keep the approval hash addressable after the subgraph indexes a row, the local store moves from hard-delete to soft-delete: useEarnDeliveryWatcher now flips a settled: true flag on entries it
  used to remove. The merge filters settled entries out of the in-flight bucket but still uses them to enrich the corresponding subgraph row. TTL + per-account cap continue to bound storage. Retry
  callers carry the prior approval forward via a new priorApprovalTxHash prop on useDeposit — when allowance is still on-chain and user-signed-approval doesn't fire on the retry, the new local
  entry inherits the original hash so the approval step keeps showing across retries.

  The success toast moves to the hemi-earn layout (a single mount point that survives navigation between the pool page and the home). It fires when a deposit transitions into subgraph CLAIMED —
  i.e., the full cross-chain flow completing — instead of at request-deposit mined. The title reverts to "Deposit successful". The toast also resets its snapshot and already-shown refs on wallet
  swap so historical CLAIMED rows on a freshly-connected account don't fire false toasts.

  Two smaller fixes ride along: the total amount in the drawer is preserved across resetStateAfterOperation() by capturing amountIn in DepositOperation on the first signing event (drawer reads from
   there instead of the live form input), and the "Add token to wallet" CTA is gated on subgraph CLAIMED so the wallet isn't pointed at a token before the shares actually land.

### Screenshots

https://github.com/user-attachments/assets/b9c2edd8-efb1-415a-bef1-900c951d6480

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx
```diff
@@ -163,63 +130,59 @@ export const ReviewDeposit = function ({ onClose }: Props) {
...
+        return ProgressStatus.PROGRESS
+      }
+      return ProgressStatus.NOT_READY
+    }
+
+    return {
+      description: <span>{t('get-share-tokens')}</span>,
+      status: getStatus(),
+      txHash: terminalHash,
     }
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx:207]: Intentional — the cross-chain delivery isn't a user-signed tx (the Agent on Ethereum performs the claim/recover).

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
```diff
@@ -26,85 +31,67 @@ type Translator = ReturnType<
...
   t: Translator,
-): StepPropsWithoutPosition => ({
-  explorerChainId: hemi.id,
-  ...stepConfigByStatus[tx.status](tx, t),
+  status: ProgressStatusType,
+) => ({
+  description: <span>{t('step.get-share-tokens')}</span>,
+  status,
+  txHash: getTerminalDeliveryTxHash(tx),
 })
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx:93]: Same.

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
```diff
@@ -26,85 +31,67 @@ type Translator = ReturnType<
...
+    waitingForShares: ProgressStatus.NOT_READY,
+  },
+}
+
+function buildStakeStep(tx: EarnTransaction, token: EvmToken, t: Translator) {
+  const status = stepStatesByStatus[tx.status].stake
+  return {
+    description: (
+      <div className="flex items-center gap-x-2">
+        <SparkleIcon />
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx:75]: Was omitting `explorerChainId` here intentional? `claimTxHash` and `recoverTxHash` both originate from `RequestClaimed`/`RequestRecovered` events on the Router contract, which is deployed on Hemi — the same chain as `buildStakeStep`'s `explorerChainId: hemi.id`. Without it, the `Completed` sub-component in `step.tsx` renders the "Confirmed" label but suppresses the "See on explorer" link (the guard is `explorerChainId && txHash`), so users see that cross-chain delivery completed but can't click through to verify it on-chain.

*— Gabi 🤖*

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx:75]: Intentional — this step represents the protocol's background delivery (the Agent's auto-claim emitting `RequestClaimed`/`RequestRecovered`), not a user-signed transaction

gndelia (commented):
> The approval step is no longer rendered. Approval becomes a pre-drawer wallet modal, mirroring the tunnel pattern where the approval row only surfaces when its tx hash is available (historical/re-sync).

This is incorrect. In the tunnel, as soon as the user signs on their wallet, the drawer appears, and the approve step is listed

The approve step may not be listed after a resync though, because there's no way to link an approval to an operation

you can also check on other pages (such as stake)

ArturDolzan:
> > The approval step is no longer rendered. Approval becomes a pre-drawer wallet modal, mirroring the tunnel pattern where the approval row only surfaces when its tx hash is available (historical/re-sync).
> 
> This is incorrect. In the tunnel, as soon as the user signs on their wallet, the drawer appears, and the approve step is listed
> 
> The approve step may not be listed after a resync though, because there's no way to link an approval to an operation
> 
> you can also check on other pages (such as stake)

@dvnahuel should we add the approval step on the drawer opened by the form? cc @gndelia. Please see the video below:

https://github.com/user-attachments/assets/a18b5f53-4a90-4897-a0ca-72efbd00265b

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx:75]: Fair enough on the intent, but it leads to an internal inconsistency: `buildWaitingForSharesStep` already passes `txHash` (the Agent's `claimTxHash`/`recoverTxHash`) — but in `step.tsx`'s `Completed` component, `txHash` is only consumed inside the `explorerChainId && txHash` guard. Without `explorerChainId`, the hash is passed to the component and silently ignored.

So which is the intended design: hide delivery tx details entirely (→ remove `txHash` too), or surface the hash with an explorer link for verifiability (→ add `explorerChainId: hemi.id`)? As it stands the two choices are in conflict.

*— Gabi 🤖*

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx:75]: Deliberate. The `Completed` step gates two separate things on `txHash`: the "Confirmed" caption (`txHash ? ...`) and the explorer link (`explorerChainId && txHash`). We want the caption — it's the user-facing signal that the cross-chain phase actually landed — but not the link (since the user didn't sign the claim/recover tx). Dropping `txHash` to keep things symmetric would lose the "Confirmed" caption entirely (`bottom` returns `null`), which is a worse outcome than the current asymmetry.

gndelia:
> > > The approval step is no longer rendered. Approval becomes a pre-drawer wallet modal, mirroring the tunnel pattern where the approval row only surfaces when its tx hash is available (historical/re-sync).
> > 
> > 
> > This is incorrect. In the tunnel, as soon as the user signs on their wallet, the drawer appears, and the approve step is listed
> > The approve step may not be listed after a resync though, because there's no way to link an approval to an operation
> > you can also check on other pages (such as stake)
> 
> @dvnahuel should we add the approval step on the drawer opened by the form? cc @gndelia. Please see the video below:
> 
>  Gravacao.de.Tela.2026-06-02.as.21.41.03.mov

To me, this looks better

Are we saving the deposits on local storage? I remember you're saving some info there. IF that's the case, you could save the approveTxHash just like we do in the tunnel, and the step would be shown when reopening.

The tradeoff is that the only scenario where we can show it (and imho there's no way to do so with 100% certainty) is when data is resynced, or read from scratch from the subgraph. There, we wouldn't show it, and imho it's not ideal but ok

If you're not saving anything of this on local storage then ignore my comment

ArturDolzan:
> > > > The approval step is no longer rendered. Approval becomes a pre-drawer wallet modal, mirroring the tunnel pattern where the approval row only surfaces when its tx hash is available (historical/re-sync).
> > > 
> > > 
> > > This is incorrect. In the tunnel, as soon as the user signs on their wallet, the drawer appears, and the approve step is listed
> > > The approve step may not be listed after a resync though, because there's no way to link an approval to an operation
> > > you can also check on other pages (such as stake)
> > 
> > 
> > @dvnahuel should we add the approval step on the drawer opened by the form? cc @gndelia. Please see the video below:
> > Gravacao.de.Tela.2026-06-02.as.21.41.03.mov
> 
> To me, this looks better
> 
> Are we saving the deposits on local storage? I remember you're saving some info there. IF that's the case, you could save the approveTxHash just like we do in the tunnel, and the step would be shown when reopening.
> 
> The tradeoff is that the only scenario where we can show it (and imho there's no way to do so with 100% certainty) is when data is resynced, or read from scratch from the subgraph. There, we wouldn't show it, and imho it's not ideal but ok
> 
> If you're not saving anything of this on local storage then ignore my comment

Sounds good! Applied on 8aa9746e68cd7e173638fbbc6c28f742656d7ac1. So `useDeposit` now saves the `approvalTxHash` to localStorage on `user-signed-approval`, and the historical drawer enriches the subgraph row with it so the approval step renders whenever the entry was originated on this browser. Resyncs / fresh subgraph reads won't show the step, like you mentioned.

Updated the video and the PR description to reflect all this.

gabmontes (dismissed):
LGTM. The soft-delete pattern, approval-hash carry-through on retries, and the two-step drawer structure are all sound. The `txHash`-without-`explorerChainId` on the delivery step is intentional and the reasoning holds.

*— Gabi 🤖*

portal/app/[locale]/hemi-earn/_utils/index.ts
```diff
@@ -1,8 +1,21 @@
...
+// The terminal on-chain hash for a delivery — claim for CLAIMED, recover
+// for RECOVERED. Used by the drawer's waiting-for-shares step to render
+// the "Confirmed" bottom row.
+export const getTerminalDeliveryTxHash = function (
+  tx: EarnTransaction | undefined,
+) {
+  if (tx?.status === 'CLAIMED') return tx.claimTxHash ?? undefined
+  if (tx?.status === 'RECOVERED') return tx.recoverTxHash ?? undefined
+  return undefined
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_utils/index.ts:31]: sounds like a few tests would come good here

dvnahuel:
> > > The approval step is no longer rendered. Approval becomes a pre-drawer wallet modal, mirroring the tunnel pattern where the approval row only surfaces when its tx hash is available (historical/re-sync).
> > 
> > 
> > This is incorrect. In the tunnel, as soon as the user signs on their wallet, the drawer appears, and the approve step is listed
> > The approve step may not be listed after a resync though, because there's no way to link an approval to an operation
> > you can also check on other pages (such as stake)
> 
> @dvnahuel should we add the approval step on the drawer opened by the form? cc @gndelia. Please see the video below:
> 
>  Gravacao.de.Tela.2026-06-02.as.21.41.03.mov

Let's move forward with the implementation proposed by Gonza.

portal/app/[locale]/hemi-earn/_utils/index.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/index.ts:31]: Added on d3581abec05f20a49d8c3a0a79cad5d9fa660090. Thanks!

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx
```diff
@@ -37,6 +42,17 @@ export const ReviewDeposit = function ({ onClose }: Props) {
...
   const { address } = useAccount()
 
+  // Shared subscription with the layout-mounted watcher; lets the new
+  // get-share-tokens step flip to COMPLETED off the subgraph status.
+  const { data: subgraphRows = [] } = useEarnTransactionsQuery()
+  const subgraphRow = depositOperation?.transactionHash
+    ? subgraphRows.find(
+        r =>
+          r.initiateTxHash.toLowerCase() ===
+          depositOperation.transactionHash!.toLowerCase(),
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx]: Any reason not to use `isAddressEqual` here?

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx
```diff
@@ -104,7 +120,7 @@ export const ReviewDeposit = function ({ onClose }: Props) {
         }
       : undefined
 
-  const addApprovalStep = function (): StepPropsWithoutPosition {
+  const addApprovalStep = function () {
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx:119]: I think this improves typings

```suggestion
  const addApprovalStep = function (): StepPropsWithoutPosition {
```

Comment (gndelia) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx:119]: ah actually I see the array where the output of this function is called is already typed so yes, I think this is correct to remove

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx]: `initiateTxHash` is a tx hash (32 bytes), not an address, so `isAddressEqual` wouldn't fit type-wise — `viem` doesn't ship a hash-equality helper either. There was already a tiny `hashesMatch` helper living inside `useEarnTransactions.ts` doing exactly this, so I promoted it to `_utils/index.ts` and used it here and in the other places we were doing the same compare. Tests in `_utils/index.test.ts`.

Comment (gndelia) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx]: oops you're correct!

gndelia (approved):

gabmontes (approved):
v3 looks good. `hashesMatch` extracted and tested, toast correctly fires on subgraph CLAIMED, amount preserved across reset. LGTM.

*— Gabi 🤖*

Issue #1973, opened 2026-06-02
Hemi Earn: derive asset registry from Router.assetsData on-chain (drop hardcoded constants + filter disabled assets)


ArturDolzan:
## Context

The new `Router.assetsData(asset)` (introduced in `hemi-earn` `master` after the `b071540 Stateless Agent` refactor) returns the full registry struct:

```solidity
struct AssetData {
  address share;        // Hemi-side share OFT
  address remoteAsset;  // Ethereum-side asset (Vetro gateway input)
  address remoteShare;  // Ethereum-side staking vault (sVetToken)
  bool enabled;         // accepts new requests
}
```

Most of `packages/hemi-earn-actions/src/constants.ts` (`HEMI_EARN_SHARES_REGISTRY` + derived helpers) hardcodes data that is now the contract's source of truth. The portal also doesn't honor the `enabled` flag — disabled pools currently still show up.

### 1. Replace hardcoded registry with on-chain reads

Remove from `packages/hemi-earn-actions/src/constants.ts`:
- `HEMI_EARN_SHARES_REGISTRY` and `HemiEarnShareEntry` type
- `SVETBTC_OFT_ADDRESS` (placeholder)
- `HEMI_EARN_SHARES` derived list
- `getStakingVaultForShare`
- `getGatewayForShare`
- `getPeggedTokenForShare`
- `findShareEntry`

Keep:
- `HEMI_EARN_SUPPORTED_ASSETS` (Hemi-side asset list — Router has no enumeration getter, so this stays as the catalog entrypoint)
- `HEMI_EARN_ROUTER_ADDRESS`, `HEMI_EARN_AGENT_ADDRESS`

Add:
- `useAssetData(asset: Address)` React-Query hook wrapping `getAssetData()` (`packages/hemi-earn-actions/src/actions/public/getAssetData.ts`). `staleTime: Infinity` — the struct is immutable post-registration.

Migrate call sites:
- `portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useQuoteDeposit.ts` (resolves `share` for `quoteDepositFulfillment`)
- `portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts` (resolves `stakingVault` for `useAssetsToShares`)
- Any other consumer of `getStakingVaultForShare` / `getGatewayForShare` / `getPeggedTokenForShare`

### 2. Honor the `enabled` flag

[...]

PR #1974, opened 2026-06-02
Realign hemi-earn-actions and subgraph with new contracts

ArturDolzan:
### Description

This PR realigns `hemi-earn-actions`, the portal deposit/redeem flows, and the `hemi-earn-router-subgraph` with the current `hemi-earn` `master` contracts. Several shape changes had landed on the contract side.

  On the contract side, `Router.quoteRedeem` and `Router.requestRedeem` now take a new `bool isInstant_` parameter so the Router can reserve the right remote gas budget at quote time. The portal resolves this on-chain through a new `resolveIsInstant` action that wraps `getCooldownEnabled` + `getInstantWithdrawWhitelist` from `@vetro-protocol/earn`, mirroring the exact check `Agent.handleRedeemRequest` performs. The resolved value is threaded from `useQuoteRedeem` (which exposes it on the quote tuple) through `useWithdraw`, `useWithdrawFees`, the review/retry
  components, and the `encodeRequestRedeem` encoder, so the request body and the gas reservation can never disagree.

  `Router.assetsData(asset)` expanded to `{share, remoteAsset, remoteShare, enabled}` and the `requests` struct gained an `operator` field — the `AssetData` and `Request` types in the package reflect the new tuple shape so decoding works again. The `Stateless Agent` refactor (commit `b071540`) also stopped resolving share-from-asset internally on the Agent, so `quoteDepositFulfillment` now expects the **remote staking vault** address directly; `useQuoteDeposit` was passing the Hemi-side asset and getting the wrong (or zero) fee back — fixed by passing `getStakingVaultForShare(shareAddress)`.

  New `Router.cancel(requestId)` is wrapped by a `cancelRedeem` wallet action and indexed by the subgraph through a new `CancellationRequested` handler that records `cancellationRequestedAt` without moving the request out of `PENDING` — the keeper-driven `Agent.cancel` callback is what eventually transitions the Router status. UI for the cancel flow lands in a follow-up PR, together with the `constants.ts` cleanup (tracked in #1973).

  The subgraph also picks up the new event shapes (`amountOutMin` in deposit/redeem requested events, `receiver` removed from `RequestClaimed`/`RequestRecovered`, `amountIn` renamed to `amount` in fulfilled/cancelled).

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useQuoteRedeem.ts
```diff
@@ -2,43 +2,74 @@ import { useQuery } from '@tanstack/react-query'
...
+        caller: account!,
+        client: ethereumClient,
+        stakingVault: getStakingVaultForShare(shareAddress),
+      })
       const fulfillmentFee = await quoteRedeemFulfillment({
         agentAddress: getHemiEarnAgentAddress(),
         asset,
-        client: getEvmL1PublicClient(mainnet.id),
+        client: ethereumClient,
       })
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useQuoteRedeem.ts:70]: Nice! Applied on f4d8dafd3303fa6e5dc68ea807285287fe7bf124. Thanks!

packages/hemi-earn-actions/src/actions/public/getAssetData.ts
```diff
@@ -10,6 +10,15 @@ export type AssetData = {
...
   remoteAsset: Address
+  // Ethereum-side staking vault (sVetToken) the Agent stakes pegged tokens
+  // into. Carried in the compose message on deposit so the stateless Agent
+  // knows which vault to use without an on-chain registry lookup.
+  remoteShare: Address
+  // Whether the asset accepts new deposit/redeem requests. The Router reverts
+  // `requestDeposit`/`requestRedeem` for disabled assets. Pool enumeration in
+  // the portal is still static (see `fetchHemiEarnShares`); when that path is
+  // moved to read `getAssetData` on-chain, filter on `enabled !== false`.
+  enabled: boolean
```
gndelia: can we sorth these alphabetically? `enabled` should go at the top

packages/hemi-earn-actions/src/actions/public/resolveIsInstant.ts
```diff
@@ -0,0 +1,41 @@
...
+  stakingVault: Address
+}): Promise<boolean> {
+  if (isAddressEqual(stakingVault, zeroAddress)) {
+    throw new Error(
+      'resolveIsInstant: `stakingVault` cannot be the zero address',
+    )
+  }
+  if (isAddressEqual(caller, zeroAddress)) {
+    throw new Error('resolveIsInstant: `caller` cannot be the zero address')
+  }
```
gndelia: These 2 checks are already being made inside vetro's function, so you can discard them

packages/hemi-earn-actions/src/routerAbi.ts
```diff
@@ -3,7 +3,7 @@ export const routerAbi = [
     inputs: [
       { internalType: 'address', name: 'asset_', type: 'address' },
       { internalType: 'uint256', name: 'assets_', type: 'uint256' },
-      { internalType: 'uint128', name: 'fulfillmentFee_', type: 'uint128' },
+      { internalType: 'uint128', name: 'callbackFee_', type: 'uint128' },
```
gndelia, line 6: I see this change, but the rest of the code keeps using `fulfillmentFee` for variables and all that. Should we update them as well?

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useQuoteRedeem.ts
```diff
@@ -2,43 +2,76 @@ import { useQuery } from '@tanstack/react-query'
...
+      const ethereumClient = getEvmL1PublicClient(mainnet.id)
+      const [isInstant, fulfillmentFee] = await Promise.all([
+        resolveIsInstant({
+          caller: account!,
+          client: ethereumClient,
+          stakingVault: getStakingVaultForShare(shareAddress),
+        }),
+        quoteRedeemFulfillment({
+          agentAddress: getHemiEarnAgentAddress(),
+          asset,
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useQuoteRedeem.ts:61]: On the contract side, `quoteRedeemFulfillment` is named to be 

> The underlying asset token on the remote chain.

That seems to be needing an Ethereum address (though is not totally clear for me). Claude is telling me that here `asset` seems to be the OFT address that lives on Hemi, so that means it's incorrect. I can't follow the code to understand where this `asset` is coming from. Since you wrote it, can you double check we're feeding the appropiate info here?

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
```diff
@@ -145,3 +150,19 @@ export function handleRequestRecovered(event: RequestRecoveredEvent): void {
...
+  event: CancellationRequestedEvent,
+): void {
+  const request = loadOrInitRequest(
+    event.params.requestId.toString(),
+    FALLBACK_KIND,
+    event,
+  )
+  request.cancellationRequestedAt = event.block.timestamp
+  request.save()
+}
```
gndelia, line 168: FYI it seems now on "cancellation" the funds are not released. So it means that there's an extra step from Cancelled -> Finalized where users will get the funds. In `automatic=true` that happens at the same time, but we may need to consider this case. Take note so we don't forget about this - we'll need the scenario to release funds manually if for some reason fails to do so automatically

packages/hemi-earn-actions/src/actions/public/getAssetData.ts
ArturDolzan: Done — sorted alphabetically in 43618998. Kept the on-chain struct order in `routerAbi.ts` because that order is load-bearing for viem decoding.

packages/hemi-earn-actions/src/actions/public/resolveIsInstant.ts
ArturDolzan: Good catch — confirmed both checks are already done inside `@vetro-protocol/earn` via `isAddressValid`. Dropped the local guards (and the matching tests) in fdfacd40.

packages/hemi-earn-actions/src/routerAbi.ts
ArturDolzan: You're right — the previous PR only renamed it in the ABI and left the rest of the code on `fulfillmentFee`. Renamed everything (types, variables, error strings, portal hooks) to `callbackFee` in 81e7d141.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useQuoteRedeem.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useQuoteRedeem.ts:61]: Confirmed the bug — `quoteRedeemFulfillment` runs on Ethereum and expects the Ethereum-side asset, but the portal was passing the Hemi-side OFT. The reason it didn't blow up in the sandbox is that AgentMock._quote returns 0 unconditionally (ignoring the address arg), so the wrong address was never validated. Resolved via `Router.assetsData(asset).remoteAsset` in 49b688f3 — same fix applied to the analogous `previewDeposit` call in useQuoteDeposit.

subgraphs/hemi-earn-router-subgraph/src/mappings/router.ts
ArturDolzan: Great point — and it applies symmetrically to the FULFILLED side too: if `automatic=true` and the auto-finalize call reverts (`AutoFinalizationFailed`), the user is stuck without a CTA. Opened #1977 to cover both manual claim (FULFILLED → CLAIMED) and manual recover (CANCELLED → RECOVERED). Same UI pattern, same trigger condition (`automatic=false` OR `autoFinalizationFailed=true`). The subgraph `AutoFinalizationFailed` indexing needed to drive that UI will land with the issue.

gndelia (dismissed):

packages/hemi-earn-actions/src/actions/public/resolveIsInstant.ts
gabmontes, line 1: The rest of the public actions guard against zero-address inputs (e.g. `quoteDepositFulfillment` throws on `zeroAddress` share, `quoteRedeem` throws on `zeroAddress` asset). `resolveIsInstant` makes two external contract calls with `stakingVault` but never validates it. A caller that passes `zeroAddress` will get an unexpected RPC error instead of the descriptive throw the package convention implies.

Suggest adding before the `Promise.all`:
```ts
if (isAddressEqual(stakingVault, zeroAddress)) {
  throw new Error('resolveIsInstant: `stakingVault` cannot be the zero address')
}
```

— Auto Gabi

packages/hemi-earn-actions/test/actions/public/resolveIsInstant.test.ts
gabmontes, line 1: Three of the four possible input combinations are tested. The fourth — `cooldownEnabled=false` AND `whitelisted=true` — is missing. The short-circuit means it resolves to `true` regardless of whitelist, but a dedicated test case makes the intent explicit and catches any future refactor that breaks it.

— Auto Gabi

packages/hemi-earn-actions/src/actions/wallet/cancelRedeem.ts
gabmontes, line 1: The validation block only guards on `requestId` being zero. Unlike `requestRedeem` / `requestDeposit`, there's no guard on `account` being the zero address. The contract will revert (`msg.sender != operator`) but the client gets a generic `user-signing-tx-error` rather than the descriptive `tx-failed-validation` event. Since the comment says "Caller must be the request's operator", adding a client-side pre-check for `account !== zeroAddress` would be consistent with the rest of the wallet actions.

— Gabi 🤖

gabmontes (commented):
Solid realignment — the `isInstant` threading and the `callbackFee`/`share` fixes are clean. Three minor findings below.

— Gabi 🤖

packages/hemi-earn-actions/src/actions/public/resolveIsInstant.ts
ArturDolzan: This conflicts with an earlier review (https://github.com/hemilabs/ui-monorepo/pull/1974#discussion_r3341908137) that explicitly asked the opposite, because @vetro-protocol/earn already validates the address.

gndelia: Bad robot that doesn't read discussions :whip: !

packages/hemi-earn-actions/test/actions/public/resolveIsInstant.test.ts
ArturDolzan: Good catch — added the missing `cooldownEnabled=false` + `whitelisted=true` case in a9f31dc1 so the truth table is fully covered.

packages/hemi-earn-actions/src/actions/wallet/cancelRedeem.ts
ArturDolzan: Fair point — added the guard in 4b639311. Emits `tx-failed-validation` with `'invalid account'` if `account = zeroAddress`, mirroring the existing `requestId` guard.

gabmontes (approved):
LGTM

— Gabi 🤖

gndelia (approved):

Issue #1981, opened 2026-06-03
Implement Slippage Control and Advanced Settings on Hemi Earn.


dvnahuel:
## Summary

Implement a slippage tolerance control and the Advanced Settings panel Hemi Earn. The control provides an `Auto` default with manual override and progressive warnings at higher thresholds.

## Context

Users need the ability to adjust slippage tolerance before submitting a transaction. High slippage increases exposure to price impact, so the UI should clearly communicate risk at elevated levels while still allowing users to proceed if they choose.

### Advanced Settings Panel

- **Max slippage control** with two modes:
  - `Auto` (default, set to 0.5%) 
  - Manual input (user types a custom %)
- **"Approve 10x deposit"** toggle (below slippage control)

### States & Thresholds

| Slippage | State | Behavior |
|---|---|---|
| Auto (1%) | Default | No warnings |
| ≤5% | Custom, safe | No warnings |
| >5% and ≤10% | High slippage | Inline warning label (yellow): "High slippage" |
| >10% | Very high slippage | Inline warning label (red): "Very high slippage" |

### Warning Modal

Triggered when the user attempts to deposit/stake with slippage **above 10%**.

- **Title:** High Slippage Warning
- **Description:** `Your slippage is set above [x]%. This increases your exposure to price impact. You may receive significantly fewer tokens than quoted.`
  - The `[x]%` value is highlighted in the corresponding warning color (yellow for >10%, red for >20%).
- **Checkbox:** "I understand and accept the risk" (must be checked to enable the action button)
- **CTA:** `Deposit` button, disabled until checkbox is checked.

### Auto Button Behavior

- Clicking `Auto` resets slippage to the default value (1%).
- `Auto` appears as a toggle/chip next to the manual input field.
- When `Auto` is active, the manual input shows the auto value but is not editable until the user clicks on it.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=18815-8300&t=Z8dyhTUMLHqVHkMa-1)

PR #1984, opened 2026-06-05
Wire hemi-earn slippage and preview errors

ArturDolzan:
### Description

This PR implements the UX_SPEC slippage defaults for Hemi Earn (0.5% deposit, 1% redeem) and plumbs `sharesOutMin`/`assetsOutMin` through every deposit and withdraw entry point via a new `applySlippage` helper.

  Along the way it fixes a few related bugs in the forms. The spinner was getting stuck because `isLoading` was being derived from data values (`hasFees`) instead of the actual query state — switching to react-query's `isLoading` flags makes the spinner stop correctly on both success and error. The "Amount too small" safety net handles the ERC4626 floor-rounding case where `convertToShares` returns zero for tiny deposits, and the same path now discriminates between three distinct causes: floor-rounding ("Amount too small"), gateway returning `peggedAmount=0n` ("Asset unavailable"), and query errors ("Preview failed — try again"). Before, all three rendered the same misleading message.

  Also drops the hand-rolled `inversePreviewRedeem` in favor of `Gateway.previewWithdraw`, which is the canonical inverse and now ships natively in `@vetro-protocol/gateway` 1.1.0. Same on-chain math, fewer assumptions about the fee model.

  Finally extracts `resolveErrorKey`, `computeIsLoading`, `resolveValidationError`, and `resolvePreviewIssue` into a shared `_utils/formState.ts` so `deposit.tsx` and `withdraw.tsx` use one implementation. This also drops the cyclomatic complexity in both forms below the lint threshold.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #1848 
Related to #1981

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-earn/_constants/slippage.ts
```diff
@@ -0,0 +1,13 @@
...
+// otherwise revert `claimUnstake`.
+// User-selectable slippage control tracked in
+// https://github.com/hemilabs/ui-monorepo/issues/1981
+export const DEPOSIT_SLIPPAGE_BPS = BigInt(50)
+export const REDEEM_SLIPPAGE_BPS = BigInt(100)
+
+const BPS_DENOMINATOR = BigInt(10000)
+
+export const applySlippage = (amount: bigint, bps: bigint) =>
+  (amount * (BPS_DENOMINATOR - bps)) / BPS_DENOMINATOR
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_constants/slippage.ts]: Good catch — fixed in [`529fc908`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/529fc908). Also added a `RangeError` for `bps > BPS_DENOMINATOR` so a misconfigured value can't slip past the clamp into the same zero-out failure mode (relevant for #1981 down the line).

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/deposit.tsx
```diff
@@ -122,14 +139,45 @@ export const Deposit = function ({ onSwitchToWithdraw }: Props) {
...
+
+  const {
+    data: quote,
+    isError: isQuoteError,
+    isLoading: isQuoteLoading,
+  } = useQuoteDeposit({
+    amount,
+    asset: selectedAsset.address,
+    shareAddress: pool.shareAddress,
+  })
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/deposit.tsx:154]: Yep, refactored in [`054b4e2c`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/054b4e2c). `useDepositFees` and `useWithdrawFees` now take `quote` + `shares` from the caller and derive `canSubmit` + slippage minimum internally. Form owns the single preview subscription per query, and both hooks share a mirror shape now.

portal/app/[locale]/hemi-earn/_constants/slippage.ts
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_constants/slippage.ts]: Added in [`3bb25599`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/3bb25599) — covers typical amounts, the new clamp branch, zero/negative input, edge `bps` values (0, `BPS_DENOMINATOR`, out-of-range), and large bigints. Thanks for flagging!

packages/hemi-earn-actions/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "dependencies": {
     "@vetro-protocol/earn": "1.0.0",
-    "@vetro-protocol/gateway": "1.0.0",
+    "@vetro-protocol/gateway": "1.1.0",
```
gndelia: Can we bump to `1.2.0` ? I published it yesterday

```suggestion
    "@vetro-protocol/gateway": "1.2.0",
```

gndelia:
It seems we should relate to https://github.com/hemilabs/ui-monorepo/issues/1981 as well

portal/app/[locale]/hemi-earn/_constants/slippage.ts
```diff
@@ -0,0 +1,27 @@
+// UX_SPEC §5.2 / §5.3: 0.5% deposit, 1.0% redeem. Redeem is wider because
+// on the cooldown path `amountOutMin` is frozen for ~7 days on the Agent
+// (Gotcha #6) — any yield drift or Vetro fee change in the meantime would
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_constants/slippage.ts]: `Gotcha #6`


What's this?

portal/app/[locale]/hemi-earn/_constants/slippage.ts
```diff
@@ -0,0 +1,27 @@
+// UX_SPEC §5.2 / §5.3: 0.5% deposit, 1.0% redeem. Redeem is wider because
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_constants/slippage.ts]: `UX_SPEC §5.2 / §5.3:`

where is that spec referring to?

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_utils/formState.ts
```diff
@@ -0,0 +1,64 @@
...
+  // blocked, not that the requested amount was too small.
+  peggedAmount: bigint | undefined
+  validInput: boolean
+}): PreviewIssue | undefined {
+  if (!validInput || isPreviewLoading) return undefined
+  if (isPreviewError) return 'network-error'
+  if (peggedAmount === BigInt(0)) return 'asset-unavailable'
+  if (!hasShares) return 'amount-too-small'
+  return undefined
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_utils/formState.ts:39]: sounds like we can add tests to this function

portal/package.json
```diff
@@ -25,7 +25,7 @@
     "@tanstack/react-table": "8.21.3",
     "@tanstack/react-virtual": "3.13.24",
     "@vetro-protocol/earn": "1.0.0",
-    "@vetro-protocol/gateway": "1.0.0",
+    "@vetro-protocol/gateway": "1.1.0",
```
gndelia: same here re: bump to `1.2.0`

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/deposit.tsx
```diff
@@ -122,37 +135,52 @@ export const Deposit = function ({ onSwitchToWithdraw }: Props) {
...
+
+  const {
+    data: shares,
+    isError: isSharesError,
+    isLoading: isSharesLoading,
+  } = useDepositShares({
+    peggedAmount: quote?.peggedAmount,
+    shareAddress: pool.shareAddress,
+  })
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/deposit.tsx:164]: I see that all calls to `useDepositShares` are always preceeded by `useQuoteDeposit`

Worth checking (can be done in a future PR though) if we can solve this inside the fetcher. So in `useDepositShares` we do

```ts
queryFn() {
  const quote = await queryClient.ensureQueryData(quoteDepositOptions(options))
  // at this point, quote will be defined (unless there's an error)
  // so no need to use `quote?.peggedAmount`
}
```

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts
```diff
@@ -155,17 +158,15 @@ export const useWithdrawFees = function ({
...
 
-  const { data: quote, isError: isQuoteError } = useQuoteRedeem({
-    account: receiver,
-    asset,
-    shareAddress,
-    shares: canWithdraw ? shares : BigInt(0),
-  })
+  const canWithdraw = validInput && shares > BigInt(0)
+  const assetsOutMin =
+    amount > BigInt(0) ? applySlippage(amount, REDEEM_SLIPPAGE_BPS) : BigInt(0)
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts]: So `applySlippage` handles `BigInt(0)` as `'asset-unavailable'`. Is it ok to skip that scenario here?

packages/hemi-earn-actions/package.json
ArturDolzan: Done in [`e438163e`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/e438163e688eea839bd840348b12473ae54c96f0), thanks!

portal/app/[locale]/hemi-earn/_constants/slippage.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_constants/slippage.ts]: Sorry, that was an internal scratch doc that won't outlive the PR — trimmed in [`cc60cc81`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/cc60cc81f061f5a77f8d1acf008c45d6fa846542). Left the actual reason (frozen `amountOutMin` across the cooldown) inline.

portal/app/[locale]/hemi-earn/_constants/slippage.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_constants/slippage.ts]: Same — that was the internal UX scratch doc. Removed the ref in [`cc60cc81`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/cc60cc81f061f5a77f8d1acf008c45d6fa846542) and kept the policy why inline (50bps deposit / 100bps redeem + issue link).

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_utils/formState.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_utils/formState.ts:39]: Yep — added 19 tests covering `resolvePreviewIssue`, `resolveErrorKey`, `computeIsLoading` and `resolveValidationError` in [`172eb9a4`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/172eb9a4fb1b56fde08915d37e6ac29da3435ef1).

portal/package.json
ArturDolzan: Done in [`e438163e`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/e438163e688eea839bd840348b12473ae54c96f0) — bumped both portal and hemi-earn-actions in the same commit.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/deposit.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/deposit.tsx:164]: Good call — went ahead and applied it in [`6f0a962c`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/6f0a962ca6ad63d433c5b4df921c9b4aca06268a). One nuance: I used `queryClient.fetchQuery` instead of `ensureQueryData` because the latter returns whatever is in cache regardless of staleness — with default `staleTime=0` that'd let a stale `peggedAmount` slip into `sharesOutMin` on remount. `fetchQuery` still dedupes against the in-flight `useQuoteDeposit` observer, so no double-fetch.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts]: Yeah — the form's `resolvePreviewIssue` already maps `peggedAmount = 0n` to `asset-unavailable` and disables submit before we get here, so gating again in the hook would be redundant. If you'd rather have the hook also guard explicitly, happy to open an issue to revisit 🙂

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_utils/formState.ts
```diff
@@ -0,0 +1,64 @@
...
+  // when the upstream quote/preview hasn't produced one yet. `0n` is the
+  // signal that the Gateway returned nothing — the asset is paused or
+  // blocked, not that the requested amount was too small.
+  peggedAmount: bigint | undefined
+  validInput: boolean
+}): PreviewIssue | undefined {
+  if (!validInput || isPreviewLoading) return undefined
+  if (isPreviewError) return 'network-error'
+  if (peggedAmount === BigInt(0)) return 'asset-unavailable'
+  if (!hasShares) return 'amount-too-small'
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_utils/formState.ts:37]: The JSDoc for `peggedAmount` correctly documents that `undefined` means "not yet resolved", and the comment scopes the function to *"after the preview resolves"*. But that precondition isn't enforced in code — the function relies on the `isPreviewLoading` guard above holding. If `validInput` is true and `isPreviewLoading` is false but `peggedAmount` is still `undefined` (e.g. a disabled query in a future refactor), the `=== 0n` check silently skips and `!hasShares` fires, showing "Amount too small" when there's nothing wrong. One line makes the implicit contract explicit:

```ts
if (peggedAmount === undefined) return undefined
```

— Gabi 🤖

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts
```diff
@@ -186,14 +135,16 @@ export const useWithdrawFees = function ({
   const layerZeroFee = quote?.nativeFee ?? BigInt(0)
 
   return {
+    approvalGasFees,
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts]: The "mirror of `useDepositFees`" comment explains the symmetry intent, but `approvalGasFees` is unused by the only caller (`withdraw.tsx` destructures `{ assetsOutMin, canWithdraw, isFeesError, totalFees }`). The two hooks serve different UI needs: the deposit form renders the approval split separately (`depositGasFees + (needsApproval ? approvalGasFees : 0n)`), while the withdraw form renders a single "Total gas fee" with the split already baked into `computeTotalFees`. Structural symmetry in the return shape is misleading here — it implies the hooks are interchangeable when they aren't. Also worth noting: `layerZeroFee` is already a pre-existing dead return from this same hook. This PR adds another.

— Gabi 🤖

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_utils/formState.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_utils/formState.ts:37]: Good catch — added the explicit `peggedAmount === undefined` guard in [`3ac9a10c`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/3ac9a10c45333221e1f89c5157f817b5f4aabc3f) and flipped the existing test that was locking in the old behavior. Thanks!

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useWithdrawFees.ts]: Yeah, fair — those three were unused. Trimmed `approvalGasFees`, `layerZeroFee` and `withdrawGasFees` from the return shape and rewrote the comment so the asymmetry vs `useDepositFees` is documented instead of papered over in [`d2663f86`](https://github.com/hemilabs/ui-monorepo/pull/1984/commits/d2663f86ac5ed871cd37024f7e576dcfded9ead0).

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_fetchers/fetchQuoteDeposit.ts
```diff
@@ -0,0 +1,90 @@
...
+  amount: bigint
+  asset: Address
+  shareAddress: Address
+}
+
+// Plain async fetcher — react-query agnostic, so it can be called from any
+// chained queryFn (via `queryClient.ensureQueryData(quoteDepositOptions(...))`),
+// from server-side scripts, or unit-tested directly without mocking the
+// react-query observer.
+export async function fetchQuoteDeposit({
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_fetchers/fetchQuoteDeposit.ts:35]: I had to update this on #1985  so we'll have to coordinate..

gndelia (approved):
uACK

gabmontes (approved):

PR #1986, opened 2026-06-08
Iterate withdraw drawer to 3-step with cooldown

ArturDolzan:
### Description

This PR iterates the Hemi Earn withdraw drawer to match the deposit drawer's 3-step shape: Approve / Unstake / Receive. A cooldown sub-step under Unstake ticks down toward zero before the underlying lands on the user's Hemi wallet.

  The cooldown timer reads `claimableAt` from `getRequestDetails` on Ethereum (the Vetro Agent's observation timestamp), gated on the subgraph reporting the Agent has acted (FULFILLED+) to avoid querying before the request exists on chain. New hooks: `useRequestDetails` (thin RPC reader with unit tests) and `useEarnCooldownRemaining` (local interval ticker over `claimableAt`).

  `useWithdraw` mirrors the deposit hook's local-store + retry mechanics: persists `approvalTxHash`/`initiateTxHash` to the per-browser store, accepts `priorApprovalTxHash` and `supersedesInitiateTxHash` for retry callers, and soft-settles superseded entries. The withdraw success toast moves from inline (firing on `WITHDRAW_TX_CONFIRMED`) to the hemi-earn layout via a generalized `EarnSuccessToast({ kind, title })` shared with deposit — both kinds now fire on subgraph CLAIMED, matching the full cross-chain completion instead of just the Hemi-side mine.

  The cross-chain delivery watcher also invalidates the user's Hemi token balances on CLAIMED/RECOVERED, so a redeem's underlying arriving back actually refreshes the wallet card immediately. The watcher's kind filter and the polling predicate dropped their `kind === 'DEPOSIT'` checks so REDEEM gets the same treatment.

  ## Notes

  - The countdown shows "0d 0h" in my local test because my dev cooldown is configured to 30 seconds — well below the day/hour granularity of the display. The `claimableAt` itself is read straight
  from the Vetro contract, so in production with a 7-day cooldown the countdown ticks normally ("Cooldown ends in 6d 22h").
  - Recovery UX (RECOVERED for REDEEM returns shares, not the asset) is intentionally not handled here — tracked in a separate follow-up issue.

### Screenshots

https://github.com/user-attachments/assets/04de43af-9a48-4c4d-a908-d95620b0ffb8

Retry:

https://github.com/user-attachments/assets/c1d55307-dfb6-490d-aedd-aa5cfc3c99cc

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_components/earnSuccessToast.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/earnSuccessToast.tsx
@@ -22,10 +22,11 @@ type Props = {
 }
 
 // Fires when a request reaches the subgraph `CLAIMED` state — the moment
-// the full cross-chain flow completes (shares on Ethereum for deposits,
-// underlying back on Hemi for redeems). Mounted once per kind at the
-// hemi-earn layout level so it survives navigation between the pool page
-// and the home page.
+// the full cross-chain flow completes (share OFTs landing on the user's
+// Hemi wallet for deposits, underlying tokens landing on the user's Hemi
+// wallet for redeems). Mounted once per kind at the hemi-earn layout
+// level so it survives navigation between the pool page and the home
+// page.
 //
 // To avoid noise on first mount (a session with historical CLAIMED rows),
 // the component snapshots the already-claimed hashes on the first non-
--- a/portal/app/[locale]/hemi-earn/_components/earnSuccessToast.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/earnSuccessToast.tsx
@@ -1,31 +1,39 @@
 'use client'
 
 import { hemi } from 'hemi-viem'
-import { useTranslations } from 'next-intl'
 import { useEffect, useMemo, useRef, useState } from 'react'
 import { type Address, type Hash } from 'viem'
 import { useAccount } from 'wagmi'
 
 import { useEarnTransactionsQuery } from '../_hooks/useEarnTransactionsQuery'
 import { PoolToast } from '../pool/[shareAddress]/_components/poolToast'
+import { type EarnTransactionKindType } from '../types'
 
 // Visible window before the toast unmounts. Matches the `<Toast>` primitive's
 // own auto-close (10s) so the inner close-animation isn't cut short.
 const TOAST_MS = 10_000
 
-// Fires when a deposit reaches the subgraph `CLAIMED` state — the moment
-// the full cross-chain flow completes and the shares actually land on the
-// user's wallet. Mounted once at the hemi-earn layout level so it survives
-// navigation between the pool page and the home page.
+type Props = {
+  // Subgraph kind to watch — `DEPOSIT` for the shares-landing toast,
+  // `REDEEM` for the underlying-received toast. Each consumer mounts its
+  // own instance with the matching kind + title.
+  kind: EarnTransactionKindType
+  title: string
+}
+
+// Fires when a request reaches the subgraph `CLAIMED` state — the moment
+// the full cross-chain flow completes (shares on Ethereum for deposits,
+// underlying back on Hemi for redeems). Mounted once per kind at the
+// hemi-earn layout level so it survives navigation between the pool page
+// and the home page.
 //
 // To avoid noise on first mount (a session with historical CLAIMED rows),
 // the component snapshots the already-claimed hashes on the first non-
 // empty render and only latches the toast for rows that become CLAIMED
 // after that.
-export const DepositSuccessToast = function () {
+export const EarnSuccessToast = function ({ kind, title }: Props) {
   const { address } = useAccount()
   const { data } = useEarnTransactionsQuery()
-  const t = useTranslations('hemi-earn.pool')
 
   const seenClaimedOnMountRef = useRef<Set<string> | null>(null)
   // Hashes already surfaced this session. Without this ref the toast would
@@ -47,7 +55,7 @@ export const DepositSuccessToast = function () {
   if (seenClaimedOnMountRef.current === null && data !== undefined) {
     seenClaimedOnMountRef.current = new Set(
       data
-        .filter(tx => tx.kind === 'DEPOSIT' && tx.status === 'CLAIMED')
+        .filter(tx => tx.kind === kind && tx.status === 'CLAIMED')
         .map(tx => tx.initiateTxHash.toLowerCase()),
     )
   }
@@ -59,7 +67,7 @@ export const DepositSuccessToast = function () {
       const fresh = data
         .filter(
           tx =>
-            tx.kind === 'DEPOSIT' &&
+            tx.kind === kind &&
             tx.status === 'CLAIMED' &&
             !!tx.claimTxHash &&
             !snapshot.has(tx.initiateTxHash.toLowerCase()),
@@ -67,7 +75,7 @@ export const DepositSuccessToast = function () {
         .sort((a, b) => Number(b.initiatedAt) - Number(a.initiatedAt))
       return fresh[0]
     },
-    [data],
+    [data, kind],
   )
 
   const [latched, setLatched] = useState<{
@@ -105,7 +113,7 @@ export const DepositSuccessToast = function () {
     <PoolToast
       chainId={hemi.id}
       key={latched.initiateTxHash}
-      title={t('deposit-successful')}
+      title={title}
       transactionHash={latched.claimTxHash}
     />
   )
--- a/portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts
+++ b/portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts
@@ -1,12 +1,17 @@
 'use client'
 
 import { useQueryClient } from '@tanstack/react-query'
-import { tokenBalanceQueryKeyPrefix } from 'hooks/useBalance'
+import { getHemiEarnSupportedAssets } from 'hemi-earn-actions'
+import { hemi } from 'hemi-viem'
+import { getTokenBalanceQueryKey } from 'hooks/useBalance'
 import { useEffect, useRef } from 'react'
+import { type Address, isAddressEqual } from 'viem'
 import { useAccount } from 'wagmi'
 
 import { earnPositionsKeyPrefix } from '../_fetchers/fetchEarnPositions'
 import {
+  type EarnTransaction,
+  type EarnTransactionKindType,
   type EarnTransactionStatusType,
   type LocalEarnOperation,
 } from '../types'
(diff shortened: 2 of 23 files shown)


portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts
```diff
@@ -47,10 +47,12 @@ const isInFlightStatus = (status: EarnTransactionStatusType) =>
...
 
 // Walks the previous-vs-current subgraph status maps and reports whether any
-// deposit hash transitioned from an in-flight state into a terminal delivery
-// state (CLAIMED / RECOVERED) since the last poll. First-time observations
-// of already-terminal hashes (page reload after CLAIMED) do NOT count —
-// those rows are historical and don't move balances.
+// request hash transitioned from an in-flight state into a terminal delivery
+// state (CLAIMED / RECOVERED) since the last poll. Applies to both deposits
+// (shares landing on Ethereum) and redeems (underlying tokens landing on
+// Hemi) — both move the user's pool position and TVL. First-time
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts]: Fixed on ccd243b7cf2681e4939499ae5b63d7c469bcc149. Thanks!

gabmontes (changes requested):
**1. `withdraw-token` i18n key is orphaned after this PR**

The only source reference — `label={t('withdraw-token', ...)}` in `reviewWithdraw.tsx` — is removed by this PR (replaced by `unstake-token`). The key remains in `en.json`, `es.json`, and `pt.json` with no consumer. Safe to remove from all three locale files.

---

**2. Cooldown countdown shows "0d Xh" in the final day**

`deriveCooldownPostAction` always formats using `wait-cooldown-countdown` (`{days}d {hours}h`). With a 7-day cooldown, the last 23 hours will render as "0d 22h", "0d 1h", etc. Consider a conditional key or ICU select that drops the `0d` prefix so the copy doesn't look broken when `days` is 0.

---

**3. No unit tests for `useEarnCooldownRemaining`**

`fetchRequestDetails` gets tests, but the new `useEarnCooldownRemaining` hook — which owns the tick logic, the `claimableAt`-vs-now math, and the `0`-clamp — has none. The timer math is trivially unit-testable with `vi.useFakeTimers`.

---

— Gabi 🤖

portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts
```diff
@@ -63,11 +66,10 @@ function detectCrossChainDelivery(
   return false
 }
 
-function invalidateVetroBalances(
-  queryClient: ReturnType<typeof useQueryClient>,
-) {
+function invalidateOnDelivery(queryClient: ReturnType<typeof useQueryClient>) {
   queryClient.invalidateQueries({ queryKey: vetroPoolsPrefix })
   queryClient.invalidateQueries({ queryKey: vetroUserPoolBalancePrefix })
+  queryClient.invalidateQueries({ queryKey: ['tokenBalance'] })
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts]: is it me or will this invalidate all tokenBalances? regardless, let's avoid hardcoding keys

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useEarnCooldownRemaining.ts
```diff
@@ -0,0 +1,45 @@
...
+// remaining seconds (clamped at 0).
+const TICK_MS = 60_000
+
+const nowInSeconds = () => Number(unixNowTimestamp())
+
+export function useEarnCooldownRemaining({
+  claimableAt,
+}: {
+  claimableAt: bigint | number | undefined
+}) {
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useEarnCooldownRemaining.ts]: an object is not needed for just 1 parameter

```suggestion
export function useEarnCooldownRemaining(
  claimableAt: bigint | number | undefined
) {
```

portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts]: Good call — extracted `tokenBalanceQueryKeyPrefix` from `hooks/useBalance` and used it from the watcher in 447847fbd667a58b16cf1a8daa4338a48a80c7bc.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useEarnCooldownRemaining.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useEarnCooldownRemaining.ts]: Yep, fair point — switched to positional in 5d4104ea42dc553bd7c0f48c4a30ba9aaaff89a1. Thanks!

ArturDolzan:
> **1. `withdraw-token` i18n key is orphaned after this PR**
> 
> The only source reference — `label={t('withdraw-token', ...)}` in `reviewWithdraw.tsx` — is removed by this PR (replaced by `unstake-token`). The key remains in `en.json`, `es.json`, and `pt.json` with no consumer. Safe to remove from all three locale files.
> 
> **2. Cooldown countdown shows "0d Xh" in the final day**
> 
> `deriveCooldownPostAction` always formats using `wait-cooldown-countdown` (`{days}d {hours}h`). With a 7-day cooldown, the last 23 hours will render as "0d 22h", "0d 1h", etc. Consider a conditional key or ICU select that drops the `0d` prefix so the copy doesn't look broken when `days` is 0.
> 
> **3. No unit tests for `useEarnCooldownRemaining`**
> 
> `fetchRequestDetails` gets tests, but the new `useEarnCooldownRemaining` hook — which owns the tick logic, the `claimableAt`-vs-now math, and the `0`-clamp — has none. The timer math is trivially unit-testable with `vi.useFakeTimers`.
> 
> — Gabi 🤖

Thanks Gabi! Addressed all three:

  - (a) Orphan `withdraw-token`: removed from en/pt/es in bab1e7abc4940103da81b7ba22731b9cca7ac51f.
  - (b) "0d Xh" in the final day: switched `wait-cooldown-countdown` to an ICU plural that drops the `Xd ` segment when `days === 0` (bef18b54a8adee35ade4765feb0e23e8f5782869). Now the last 23 hours read as "Cooldown ends in 22h".
  - (c) Tests for the hook: extracted the claimable-vs-now math into a pure `computeCooldownRemaining` helper and covered the undefined/clamp/coercion cases (027346261de9d73e755fa809c90f96fc90c41529).

gabmontes (commented):
Just one comment...

portal/app/[locale]/hemi-earn/_components/earnSuccessToast.tsx
```diff
@@ -1,31 +1,39 @@
...
+type Props = {
+  // Subgraph kind to watch — `DEPOSIT` for the shares-landing toast,
+  // `REDEEM` for the underlying-received toast. Each consumer mounts its
+  // own instance with the matching kind + title.
+  kind: EarnTransactionKindType
+  title: string
+}
+
+// Fires when a request reaches the subgraph `CLAIMED` state — the moment
+// the full cross-chain flow completes (shares on Ethereum for deposits,
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_components/earnSuccessToast.tsx]: The comment says `"shares on Ethereum for deposits"` — shares are OFTs that land on the user's **Hemi** wallet, not Ethereum.

— Gabi 🤖

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/earnSuccessToast.tsx]: Good catch. Fixed on c1eeee780e033099097854b08eb22abbd1b018b0. Thanks!

gabmontes (dismissed):

portal/hooks/useBalance.ts
```diff
@@ -8,6 +8,8 @@ import { isNativeAddress } from 'utils/nativeToken'
 import { type Address, isAddress } from 'viem'
 import { useAccount, usePublicClient } from 'wagmi'
 
+export const tokenBalanceQueryKeyPrefix = ['tokenBalance'] as const
```
gndelia: This will break if `@hemilabs/react-hooks/useTokenBalance` changes

Perhaps you want to do


```suggestion
export const tokenBalanceQueryKeyPrefix = `tokenBalanceQueryKey[0]`
```

portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts
```diff
@@ -63,11 +67,10 @@ function detectCrossChainDelivery(
   return false
 }
 
-function invalidateVetroBalances(
-  queryClient: ReturnType<typeof useQueryClient>,
-) {
+function invalidateOnDelivery(queryClient: ReturnType<typeof useQueryClient>) {
   queryClient.invalidateQueries({ queryKey: vetroPoolsPrefix })
   queryClient.invalidateQueries({ queryKey: vetroUserPoolBalancePrefix })
+  queryClient.invalidateQueries({ queryKey: tokenBalanceQueryKeyPrefix })
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts]: I'll ask again - this will invalidate _all_ balances of the app. Seems incorrect to me. Why don't we build the exact query we want to invalidate?

portal/messages/en.json
```diff
@@ -32,13 +32,17 @@
           "heading": "Review stake deposit",
           "subheading": "Your transaction has been submitted and is awaiting confirmation on the blockchain."
         },
+        "cooldown-ended": "Cooldown period complete",
         "stake-token": "Stake {symbol}",
         "get-share-tokens": "Get share tokens",
+        "receive-token": "Receive {symbol}",
+        "unstake-token": "Unstake {symbol}",
+        "wait-cooldown-countdown": "{days, plural, =0 {Cooldown ends in {hours}h} other {Cooldown ends in #d {hours}h}}",
```
gndelia, line 40: would this render `0` when the countdown is left, say, 25 minutes?

portal/messages/es.json
```diff
@@ -32,13 +32,17 @@
           "heading": "Revisar depósito de stake",
           "subheading": "Su transacción ha sido enviada y está esperando confirmación en la blockchain."
         },
+        "cooldown-ended": "Período de cooldown completado",
         "stake-token": "Hacer stake de {symbol}",
         "get-share-tokens": "Recibir share tokens",
+        "receive-token": "Recibir {symbol}",
+        "unstake-token": "Hacer unstake de {symbol}",
+        "wait-cooldown-countdown": "{days, plural, =0 {Cooldown termina en {hours}h} other {Cooldown termina en #d {hours}h}}",
+        "wait-cooldown-pending": "Espere el período de cooldown de {days, plural, one {# día} other {# días}}",
```
gndelia: On Vetro, we use `Período de espera` instead of `cooldown`... the literal translation would be "Waiting time", which may be redundant ("Wait the waiting time..").

but we could phrase it as 


```suggestion
        "wait-cooldown-pending": "El período de espera es de  {days, plural, one {# día} other {# días}}",
```

dvnahuel:
LGTM!

portal/hooks/useBalance.ts
ArturDolzan: Good point — went with `satisfies ReturnType<typeof tokenBalanceQueryKey>[0]` in 2b5c3b9c so the literal is checked against the lib's type. Then ended up dropping the constant entirely in 00c4c3b6 after c97472a1 switched to per-event invalidation (no more prefix consumer). Thanks!

portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useEarnDeliveryWatcher.ts]: Yeah you're right — switched to per-event invalidation in c97472a1. `detectCrossChainDelivery` now returns the delivered events with `{ asset, kind, receiver }`, and we build the exact `getTokenBalanceQueryKey(...)` per event (resolving the share OFT for DEPOSIT via `getHemiEarnSupportedAssets`, asset directly for REDEEM). Tested locally end-to-end. Thanks for pushing on this!

portal/messages/en.json
ArturDolzan: Good catch — would have shown "Cooldown ends in 0h" in the last hour. Added a `wait-cooldown-countdown-soon` key in d20746fb that renders "Cooldown ends in less than an hour" (+ PT/ES) when the remaining drops below 3600s. Thanks!

portal/messages/es.json
ArturDolzan: Done — aligned the four cooldown copies (`wait-cooldown-pending`, `wait-cooldown-countdown`, `wait-cooldown-countdown-soon`, `cooldown-ended`) in es/pt to "período de espera" in f46f4bfb. EN keeps "cooldown" as the literal. Thanks for the heads up on the Vetro terminology!

gabmontes (dismissed):

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx
```diff
@@ -37,13 +47,258 @@ type Props = {
...
+  t,
+  withdrawStatus,
+}: {
+  cooldownRemainingSec: number | undefined
+  isCooldownEligible: boolean | undefined
+  receiveToken: EvmToken
+  subgraphRow:
+    | {
+        claimTxHash: string | null
+        recoverTxHash: string | null
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx]: should these be


```suggestion
        claimTxHash: Address | null
        recoverTxHash: Address | null
```

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx
```diff
@@ -189,18 +432,25 @@ export const ReviewWithdraw = function ({ onClose }: Props) {
...
       description: (
-        <ChainLabel
-          active={withdrawStatus === WithdrawStatus.WITHDRAW_TX_PENDING}
-          chainId={chainId}
-          label={t('withdraw-token', { symbol: selectedAsset.token.symbol })}
-        />
+        <div className="flex items-center gap-x-2">
+          <TokenLogo size="small" token={pool.shareToken} />
+          <span>{t('unstake-token', { symbol: pool.shareToken.symbol })}</span>
+        </div>
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx:438]: Should we use [displayAmount](https://github.com/hemilabs/ui-monorepo/blob/main/portal/components/displayAmount.tsx) for this? So when using small amounts, we have the hover option. I think we use this in the rest of the app

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/cooldownWarning.tsx
```diff
@@ -11,9 +12,13 @@ type Props = {
...
 export const CooldownWarning = function ({ shareAddress }: Props) {
   const t = useTranslations()
-  const { data: days, isPending: isLoading } = useCooldownDuration({
+  const { data: durationSec, isPending: isLoading } = useCooldownDuration({
     shareAddress,
   })
+  const days =
+    durationSec !== undefined
+      ? Math.round(secondsToDays(durationSec))
+      : undefined
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/cooldownWarning.tsx]: I see this is used in both calls of `useCooldownDuration`. Perhaps the `Math.round(secondsToDays(durationSec))` is worth to be done inside the `queryFn`, or even as a `select` in the `useQuery`definition. Thoughts?

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx
```diff
@@ -37,13 +47,258 @@ type Props = {
...
+    ? ProgressStatus.COMPLETED
+    : unstakeMined && (!needsCooldown || cooldownElapsed)
+      ? ProgressStatus.PROGRESS
+      : ProgressStatus.NOT_READY
+  return {
+    description: (
+      <div className="flex items-center gap-x-2">
+        <TokenLogo size="small" token={receiveToken} />
+        <span>{t('receive-token', { symbol: receiveToken.symbol })}</span>
+      </div>
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx:123]: same here Re: DisplayAmount

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx:438]: Those two lines are the step labels.. they just render the token symbol ("Unstake sVetBTC" / "Receive hemiBTC"), there's no amount being formatted at that spot.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx:123]: Same.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx]: Good catch — tightened to `Hash | null` in d8b04a08 to match the canonical type in `hemi-earn/types.ts`. Used `Hash` instead of `Address` since these are 32-byte tx hashes (`claimTxHash`/`recoverTxHash`), same spirit though 🙂

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/cooldownWarning.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/cooldownWarning.tsx]: Nice — extracted a `secondsToWholeDays` helper in `utils/time.ts` (with unit tests) and pointed both call sites at it in d8b04a08. Kept it as a pure helper rather than a `select` because the drawer still needs the raw seconds for the cooldown tick math, so the `select` would only cover the warning banner.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx
Comment (gndelia) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewWithdraw.tsx:438]: ah, sorry, you're right, I got confused

gndelia (approved):
I think we discussed this already, but the status can only be updated if the user is in the Earn page.

gabmontes (approved):

PR #1997, opened 2026-06-15
Add Envio earn-requests endpoint

### LINKED ISSUE #1935: Add endpoint to expose hemi earn deposits

Deposits were added in this subgraph https://github.com/hemilabs/ui-monorepo/pull/1921

We need to expose them per user through the portal backend

~~Note that this issue will only cover deposits, unless withdrawals are implemented in the subgraph~~

Once #1937  is finished, we can also expose redeems

### LINKED ISSUE #1963: Hemi Earn: wire transactions table to subgraph once indexer is live

Replace the mocked `fetchEarnTransactions` in the hemi-earn transactions table with the real subgraph-indexed data once the indexer API is deployed.

The transactions table currently renders hardcoded sample rows from `buildMockTransactions` in `portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts`. This unblocks PR #1953 and local testing, but the table doesn't reflect actual user activity. Once the subgraph + indexer API are deployed, the table needs to consume real data to be useful end-to-end.

### Proposed implementation

Replace the body of `fetchEarnTransactions` with a `GET /:chainId/earn-requests/:address` call against the indexer (per PR #1946, which already returns `{ requests: EarnRequestRow[] }` shaped identically to `EarnTransaction`).

Notes:
- A `// TODO(api):` comment already flags this spot in the file.
- The local-subgraph scaffolding that's been added/removed during testing confirms the data shape works against a real GraphQL endpoint, so the swap is essentially mechanical.

---

Follow-up from #1953.

ArturDolzan:
### Description

This PR wires the portal to the live `hemi-earn-requests` Envio subgraph by adding a proxy endpoint on portal-backend, replacing the mock fetcher the previous PR left behind. The portal never holds the indexer URL or Bearer token in its bundle — both stay in env vars on the backend.

  The backend gets `GET /subgraphs/:chainId/earn-requests/:address`, gated to Hemi mainnet, mirroring the architecture of the closed PR #1946 (same route shape, handler/helper layout, response envelope, and reusable paginator). What diverges is the URL builder and auth: instead of The Graph's `${gateway}/api/{key}/subgraphs/id/{id}` URL embedding, Envio uses a static endpoint plus a Bearer-token header. Both are read from `SUBGRAPH_HEMI_EARN_REQUESTS_API_URL` and `SUBGRAPH_HEMI_EARN_REQUESTS_API_KEY`, mapped via `custom-environment-variables.json`. Pagination uses plain
  `limit`/`offset` because Envio runs on Postgres and doesn't carry graph-node's 5000-skip cap, so the cursor-reset dance in the existing `paginateSubgraph` isn't needed.

  The boundary translator `toEarnRequestRow` lives on the backend and collapses the subgraph's `failed` flag into the synthetic `'FAILED'` status the portal already understands. It also passes `failed` and `failureReason` through to the response, so a future UI can surface the revert reason without another backend change.

  Portal-side, `fetchEarnTransactions` drops the mock and its boundary helpers and becomes a thin GET to the new endpoint, parsing the response as `EarnTransaction[]` directly.

  Validated locally end-to-end against the user's running Envio Hasura at `localhost:8080` + portal-backend on `3006` + portal on `3000`.

  **Heads up for the merge:** prod needs `SUBGRAPH_HEMI_EARN_REQUESTS_API_URL` and `SUBGRAPH_HEMI_EARN_REQUESTS_API_KEY` injected into the portal-backend deployment from 1Password before the
  `/hemi-earn` page works in production. Ticket with the infra team is being opened in parallel.

### Screenshots

No UI changes.

### Related issue(s)

Related to #1848 
Closes #1963 
Closes #1935 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal-backend/README.md
+++ b/portal-backend/README.md
@@ -163,6 +163,15 @@ $ curl http://localhost:3006/subgraphs/43111/locks/0x000000000000000000000000000
 {"positions":[]}
 ```
 
+##### `GET /subgraphs/:chain-id/earn-requests/:address`
+
+Returns the Hemi Earn cross-chain requests (deposits and redeems) initiated by the given address. Hemi mainnet only. Powered by the Envio multichain indexer (`hemi-earn-requests-subgraph`); rows are filtered by `initiator` (`msg.sender` on the Router call).
+
+```console
+$ curl http://localhost:3006/subgraphs/43111/earn-requests/0x0000000000000000000000000000000000000001
+{"requests":[{"amountIn":"1000000000000000000","amountOut":"1000000000000000000","asset":"0x...","automatic":true,"claimableAt":"1759162804","claimTxHash":"0xabc...","failed":false,"failureReason":null,"kind":"DEPOSIT","receiver":"0x1234...","recoverTxHash":null,"requestedAt":"1759162804","requestId":"6","requestTxHash":"0xdef...","status":"FINALIZED"}]}
+```
+
 ### Configuration
 
 These environment variables control how the cache works:
--- a/portal-backend/README.md
+++ b/portal-backend/README.md
@@ -172,6 +172,11 @@ $ curl http://localhost:3006/subgraphs/43111/earn-requests/0x0000000000000000000
 {"requests":[{"amountIn":"1000000000000000000","amountOut":"1000000000000000000","asset":"0x...","automatic":true,"claimableAt":"1759162804","claimTxHash":"0xabc...","failed":false,"failureReason":null,"kind":"DEPOSIT","receiver":"0x1234...","recoverTxHash":null,"requestedAt":"1759162804","requestId":"6","requestTxHash":"0xdef...","status":"FINALIZED"}]}
 ```
 
+Requires the following env vars to be set on the backend:
+
+- `SUBGRAPH_HEMI_EARN_REQUESTS_API_URL` — the full GraphQL URL of the Envio `hemi-earn-requests` indexer.
+- `SUBGRAPH_HEMI_EARN_REQUESTS_API_KEY` — Bearer token for the Envio indexer (only sent when set; can be left empty against an unauthenticated local Envio).
+
 ### Configuration
 
 These environment variables control how the cache works:
--- a/portal-backend/api/src/subgraphs/subgraph.ts
+++ b/portal-backend/api/src/subgraphs/subgraph.ts
@@ -13,6 +13,7 @@ import { hemi, hemiSepolia, mainnet, sepolia } from 'viem/chains'
 import { postJson } from '../post-json.ts'
 
 import { UpstreamGraphQLError } from './errors.ts'
+import type { EarnRequestRow, EarnRequestStatus } from './types/earn.ts'
 import type {
   BtcDepositOperation,
   EvmDepositOperation,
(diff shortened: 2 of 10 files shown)


portal-backend/api/src/subgraphs/subgraph.ts
```diff
@@ -698,3 +699,149 @@ export const getLockedPositions = function ({
...
+// won't be mis-labeled.
+const deriveStatus = (row: SubgraphRequest): EarnRequestStatus =>
+  row.failed ? 'FAILED' : row.status
+
+const toEarnRequestRow = (row: SubgraphRequest): EarnRequestRow => ({
+  amountIn: row.amountIn,
+  amountOut: row.amountOut,
+  // @ts-expect-error address comes lowercased from the indexer
+  asset: row.asset,
+  automatic: row.automatic ?? true,
```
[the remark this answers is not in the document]
ArturDolzan: Good catch — done in `e1cb5640`. Replaced the `@ts-expect-error` with an `as Address` cast and a comment explaining `asset` stays lowercase on purpose (matches the portal-side tokens registry; only `receiver` is checksummed).

portal-backend/api/src/subgraphs/subgraph.ts
```diff
@@ -698,3 +699,149 @@ export const getLockedPositions = function ({
...
+// Envio HyperIndex uses Bearer-token auth. Header is only added when an
+// API key is configured so local dev against an unauthenticated indexer
+// (Hasura at `localhost:8080`) keeps working.
+function requestHemiEarn<TResponse>(schema: Schema): Promise<TResponse> {
+  const { apiKey, apiUrl } = subgraphConfig.hemiEarnRequests
+  const headers = apiKey ? { Authorization: `Bearer ${apiKey}` } : undefined
+  return postJson(apiUrl, schema, { headers }).catch(function (err) {
+    throw new UpstreamGraphQLError(err.message, { cause: err })
+  }) as Promise<TResponse>
+}
```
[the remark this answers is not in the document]
ArturDolzan: Nice one — done in `e1cb5640`. `requestHemiEarn` now throws an `UpstreamGraphQLError` with a clear "URL is not configured" message before hitting `fetch`, so misconfigured deploys surface immediately.

portal-backend/api/src/subgraphs/subgraph.ts
```diff
@@ -698,3 +699,149 @@ export const getLockedPositions = function ({
...
+ * wallets, and the portal's "my transactions" view follows the signer.
+ */
+export const getEarnRequests = async function ({
+  address,
+}: {
+  address: Address
+  chainId: Chain['id']
+}) {
+  const initiator = address.toLowerCase()
+
```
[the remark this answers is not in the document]
ArturDolzan: Yep, dropped it in `e1cb5640`. The route is already gated to Hemi mainnet by `validateChainIsHemiMainnet` and the subgraph is mainnet-only, so the param was pure dead weight — also got rid of the `@ts-expect-error` in the handler as a bonus.

portal-backend/README.md
```diff
@@ -163,6 +163,15 @@ $ curl http://localhost:3006/subgraphs/43111/locks/0x000000000000000000000000000
 {"positions":[]}
 ```
 
+##### `GET /subgraphs/:chain-id/earn-requests/:address`
+
+Returns the Hemi Earn cross-chain requests (deposits and redeems) initiated by the given address. Hemi mainnet only. Powered by the Envio multichain indexer (`hemi-earn-requests-subgraph`); rows are filtered by `initiator` (`msg.sender` on the Router call).
```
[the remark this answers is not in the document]
ArturDolzan: Good call — done in `e1cb5640`. Added them as an inline note right under the endpoint`s curl example. Kept them out of the global env-var table on purpose because the new names are wider than every existing entry, and adding them there forces Prettier to realign the whole table (massive cosmetic diff).

portal-backend/api/src/subgraphs/subgraph.ts
```diff
@@ -698,3 +699,157 @@ export const getLockedPositions = function ({
...
+  failed: row.failed,
+  failureReason: row.failureReason,
+  kind: row.kind,
+  receiver: toChecksum(row.receiver as `0x${string}`),
+  recoverTxHash: row.recoverTxHash as Hash | null,
+  requestedAt: row.requestedAt,
+  requestId: row.requestId,
+  requestTxHash: row.requestTxHash as Hash,
+  status: deriveStatus(row),
+})
```
gndelia, line 730: sounds like we can spread `...row` for those fields that have no change.

portal-backend/api/src/subgraphs/subgraph.ts
```diff
@@ -698,3 +699,157 @@ export const getLockedPositions = function ({
...
+  failed: boolean
+  failureReason: string | null
+  kind: 'DEPOSIT' | 'REDEEM'
+  receiver: string
+  recoverTxHash: string | null
+  requestedAt: string
+  requestId: string
+  requestTxHash: string
+  status: 'PENDING' | 'FULFILLED' | 'FINALIZED' | 'CANCELLED' | 'RECOVERED'
+}
```
gndelia: More than half of these fields are the same as in `EarnRequestRow`. Perhaps we can define one, and the other one overrides what's needed. Plus, the types are coupled already (one is transformed to the other) so that'd make it easier to keep them in sync

portal-backend/api/src/subgraphs/subgraph.ts
ArturDolzan: Good call — done in `31efc3b7`. `toEarnRequestRow` now spreads `...row` for the pass-through fields and only spells out the seven that actually transform (addresses, hashes, nullable `automatic`, synthetic status).

portal-backend/api/src/subgraphs/subgraph.ts
ArturDolzan: Nice — done in `31efc3b7`. Both types now derive from an internal `EarnRequestCommonFields` base (in `types/earn.ts`), so new subgraph fields only need to be added in one place. Also moved `SubgraphRequest` over there next to its sibling since they are conceptually a pair.

gndelia (dismissed):

portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts
```diff
@@ -1,111 +1,22 @@
...
-  ]
-}
+const getEarnRequestsUrl = (account: Address) =>
+  `${process.env.NEXT_PUBLIC_PORTAL_API_URL}/subgraphs/${hemi.id}/earn-requests/${account}`
 
 export const fetchEarnTransactions = async function ({
   account,
 }: {
   account: Address
   networkType: string
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts]: `networkType` is accepted here but silently ignored — `getEarnRequestsUrl` unconditionally uses `hemi.id`, so a testnet user's session fires a real request to the mainnet endpoint and silently receives mainnet earn transactions.

`useEarnTransactionsQuery` already includes `networkType` in the `queryKey`, so React Query treats mainnet and testnet as separate cache entries — but both resolve to the same mainnet data.

Simplest fix: drop `networkType` from the fetcher signature and add `enabled: !!address && networkType === 'mainnet'` to the `useQuery` call in `useEarnTransactionsQuery` (same pattern used by other mainnet-only features).

— Gabi 🤖

gabmontes (changes requested):
One finding — see inline comment.

portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchEarnTransactions.ts]: Good catch — done in `c816b01e`. Dropped `networkType` from the fetcher and gated `useEarnTransactionsQuery` with `enabled: !!address && networkType === 'mainnet'`, exactly as you suggested. Kept `networkType` in the `queryKey` so React Query still segments the cache properly, the testnet entry is just inert now. Thanks!

gabmontes (dismissed):

gndelia (dismissed):

gndelia (approved):

gabmontes (approved):

PR #2004, opened 2026-06-15
Surface REDEEM rows on Hemi Earn tx table

ArturDolzan:
### Description

This PR surfaces REDEEM rows on the Hemi Earn transactions table.

  The deposit-only filter in `useEarnTransactions` is lifted and the local-store ↔ subgraph merge pipeline is mirrored for REDEEM, so withdrawals reconcile in-flight the same way deposits do. The status column now branches by kind: REDEEM rows go through a new `WithdrawStatusCell` that pulls `claimableAt` from the Envio Request entity and renders an adaptive cooldown countdown ("Cooldown · ready in Nd/h/m"), flips to "Ready to claim" once the cooldown matures, and surfaces terminal labels for `FINALIZED` (Withdrawn) and `CANCELLED` (Withdrawal canceled). A trash button shows up next to View for REDEEM rows in `PENDING`/`FULFILLED` matching the UX spec's cancel window — the actual `cancelRedeem` wiring is left for a follow-up PR.

  `AmountCell` now prefers `amountOut` for REDEEM, so the cell shows the real asset-out amount (hemiBTC) once Vetro reports it, falling back to `amountIn` (shares) before fulfillment when shares ≈ pegged 1:1.

  ## Notes

  - Out of scope: drawer for withdraw View, wiring of `cancelRedeem` / `claimRedeem`, recover flow for CANCELLED rows.

### Screenshots

https://github.com/user-attachments/assets/da4ca860-8127-4abe-bf9f-83cd782b53f0

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx
```diff
@@ -0,0 +1,121 @@
...
+  t: ReturnType<typeof useTranslations<'hemi-earn.transactions'>>
+}): string | undefined {
+  if (isCooldownEligible !== true) return undefined
+  if (status === 'FULFILLED' && hasClaimableAt && remainingSec === 0) {
+    return t('status.ready-to-claim')
+  }
+  if (!isCooldownPhase(status)) return undefined
+  const displaySec = hasClaimableAt ? remainingSec ?? 0 : cooldownDurationSec
+  if (displaySec === undefined || displaySec <= 0) return undefined
+  return formatCooldownText(displaySec, t)
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx:66]: Good catch — fixed in `b032d6c4` (gated on `isCooldownPhase(status)` now so the PENDING+matured window also surfaces the label).

portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
```diff
@@ -16,21 +16,32 @@ import { type EarnTransaction } from '../../types'
...
 // Hemi-side asset address.
+// For REDEEM, `amountIn` is in share-token units (svetBTC), so prefer
+// `amountOut` (asset units) once Vetro reports it; until then the share
+// amount is a close enough approximation. For DEPOSIT, `amountIn` is
+// already in asset units.
+const displayAmountFor = (transaction: EarnTransaction) =>
+  transaction.kind === 'REDEEM'
+    ? transaction.amountOut ?? transaction.amountIn
+    : transaction.amountIn
+
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx]: IT seems to me that the token logo should change accordingly, depending on us showing pegged token or shares

Comment (gndelia) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx]: similarly, formatting is using `decimals` of `assets`, but technically speaking, shares may use a different `decimals` value

portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx
```diff
@@ -25,17 +23,47 @@ export const RowActions = function ({ transaction }: Props) {
...
+  // The Router stays PENDING for the whole cooldown window; FULFILLED is
+  // the brief post-claimUnstake window before the return OFT lands. Both
+  // are user-cancelable.
+  const showCancelButton =
+    transaction.kind === 'REDEEM' &&
+    (transaction.status === 'PENDING' || transaction.status === 'FULFILLED')
+
+  const onViewClick = function (e: MouseEvent) {
     e.stopPropagation()
+    if (transaction.kind === 'REDEEM') return
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx:37]: Clicking here does nothing? should the button be disabled/hiddenm? or is this incorrect?

portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx]: Good catch on both points — done in `b870c2d0`. AmountCell now resolves the pool and picks `shareToken` (logo + decimals) while we are showing `amountIn` (pre-FULFILLED), and falls back to the Hemi-side asset token once `amountOut` lands. DEPOSIT keeps the single `useToken(asset)` lookup.

portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx:37]: You are right, that is a known gap — the REDEEM drawer does not exist yet so the click intentionally no-ops. Left a TODO comment in `b870c2d0` marking it, and the wire-up lands in the follow-up PR that introduces the withdraw drawer.

portal/app/[locale]/hemi-earn/_components/icons/xCircleIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      clipRule="evenodd"
+      d="M8 15A7 7 0 1 0 8 1a7 7 0 0 0 0 14Zm-2.03-9.03a.75.75 0 0 1 1.06 0L8 6.94l.97-.97a.75.75 0 1 1 1.06 1.06L9.06 8l.97.97a.75.75 0 1 1-1.06 1.06L8 9.06l-.97.97a.75.75 0 0 1-1.06-1.06L6.94 8l-.97-.97a.75.75 0 0 1 0-1.06Z"
+      fill="currentColor"
+      fillRule="evenodd"
+    />
+  </svg>
+)
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/icons/xCircleIcon.tsx]: I think this is the same component as [portal/components/icons/redErrorIcon.tsx](https://github.com/hemilabs/ui-monorepo/blob/2478ae4d521cb49325015bc01d10e7ff5d37322c/portal/components/icons/redErrorIcon.tsx) (but with different colors)

gndelia:
**Pattern A: pool-by-asset lookup is now triplicated**

`withdrawStatusCell.tsx` (new in this PR) re-implements a lookup that already exists in the codebase:

```ts
// withdrawStatusCell.tsx (new)
const pool = pools.find(p =>
  p.assets.some(a => isAddressEqual(a.address, transaction.asset)),
)
```

This is the same operation as `resolvePoolAndAsset` in the sibling file `transactionDrawer/index.tsx:19-25` (which just additionally returns the matched asset):

```ts
const resolvePoolAndAsset = function (pools: EarnPool[], asset: Address) {
  for (const pool of pools) {
    const match = pool.assets.find(a => isAddressEqual(a.address, asset))
    if (match) return { asset: match, pool }
  }
  return undefined
}
```

And the asset/share→pool mapping shows up a third time in `useEarnDeliveryWatcher.ts:98` (asset→share via `isAddressEqual`). Two more call sites do the same conceptual lookup but with raw `.toLowerCase()`/`===` instead of `isAddressEqual` — `poolPageContent.tsx:31` and `poolFormContext.tsx:48` — which is a latent checksum-case mismatch bug.

Suggestion: extract a single shared helper (e.g. a pure `findPoolByAsset(pools, asset)` in `_utils`, or a `useEarnPoolByAsset(asset)` hook wrapping `useEarnPools`) and use it in `withdrawStatusCell` + the drawer now. That removes this PR's duplication and sets up consolidating the inconsistent call sites (and fixing the raw-comparison ones) in a follow-up. Not a blocker, but better to not land the third copy.

portal/app/[locale]/hemi-earn/_components/icons/xCircleIcon.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/icons/xCircleIcon.tsx]: Good call — done in `b52cb35b`. Renamed `RedErrorIcon` to `ErrorIcon`, swapped its hardcoded fill for `currentColor` so callers control the color via `className` (`text-rose-500` for FAILED + the validity row, `text-neutral-400` for REDEEM canceled), and removed the `XCircleIcon` copy.

ArturDolzan:
> **Pattern A: pool-by-asset lookup is now triplicated**
> 
> `withdrawStatusCell.tsx` (new in this PR) re-implements a lookup that already exists in the codebase:
> 
> ```ts
> // withdrawStatusCell.tsx (new)
> const pool = pools.find(p =>
>   p.assets.some(a => isAddressEqual(a.address, transaction.asset)),
> )
> ```
> 
> This is the same operation as `resolvePoolAndAsset` in the sibling file `transactionDrawer/index.tsx:19-25` (which just additionally returns the matched asset):
> 
> ```ts
> const resolvePoolAndAsset = function (pools: EarnPool[], asset: Address) {
>   for (const pool of pools) {
>     const match = pool.assets.find(a => isAddressEqual(a.address, asset))
>     if (match) return { asset: match, pool }
>   }
>   return undefined
> }
> ```
> 
> And the asset/share→pool mapping shows up a third time in `useEarnDeliveryWatcher.ts:98` (asset→share via `isAddressEqual`). Two more call sites do the same conceptual lookup but with raw `.toLowerCase()`/`===` instead of `isAddressEqual` — `poolPageContent.tsx:31` and `poolFormContext.tsx:48` — which is a latent checksum-case mismatch bug.
> 
> Suggestion: extract a single shared helper (e.g. a pure `findPoolByAsset(pools, asset)` in `_utils`, or a `useEarnPoolByAsset(asset)` hook wrapping `useEarnPools`) and use it in `withdrawStatusCell` + the drawer now. That removes this PR's duplication and sets up consolidating the inconsistent call sites (and fixing the raw-comparison ones) in a follow-up. Not a blocker, but better to not land the third copy.

Done in `b52cb35b` — extracted `findPoolByAsset` and `findPoolByShare` into `_utils` and routed every pool lookup through them (`WithdrawStatusCell`, `AmountCell`, the drawer's resolver, `PoolPageContent`, and the form context). Also swapped the raw `.toLowerCase()` / `===` checks in `PoolPageContent` + `poolFormContext` for `isAddressEqual`, picking up that latent checksum-case fix you flagged.

gndelia (approved):

gabmontes (approved):
LGTM — Auto Gabi

PR #2009, opened 2026-06-16
Add withdraw drawer to Hemi Earn tx table

ArturDolzan:
### Description

This PR adds the withdraw drawer to the Hemi Earn transactions table — clicking View on a REDEEM row now opens a read-only timeline that mirrors the deposit drawer, with an Approval step (when local-store carries the hash), an Unstake step with the cooldown post-action sub-step, and a Receive step that lights up on FINALIZED. FAILED rows get a Try again CTA via the new `RetryFailedWithdraw` wrapper around `useWithdraw`, redirecting the drawer's `earnTxId` on `user-signed-withdraw` so the FAILED view transitions in place to the new in-flight view.

  The retry CTA is gated on `isLocalEarnTransactionRow` so only Hemi-side reverts (request deposit/redeem) retry — Agent-side failures (subgraph `failed=true` after a successful Hemi tx) wait for the recovery flow that ships in a follow-up PR. The predicate is extracted into `_utils` and reused in both historical reviews. The retry wrapper also mirrors `retryWithdraw.tsx`'s safeguards: the button stays disabled until `useSharesToAssets` resolves (so `assetsOutMin` can never collapse to zero), and `isInstant` is read straight from the redeem quote so the request body, gas reservation, and Agent check always agree.

  While in there, `deriveCooldownPostAction` was pulled out of `reviewWithdraw` into a shared helper consumed by both drawers, the historical `unstakeMined` was aligned to the resolved step state so a failed local row no longer shows a misleading cooldown spinner under a FAILED Unstake step, and `useWithdraw`'s `updateWithdrawOperation` became optional (matching `useDeposit`) so the historical retry doesn't need a no-op callback. Three new i18n keys land for the withdraw drawer copy plus a `common.withdrawing` filler the button label was missing.

  ## Note
  - Out of scope: manual claim/recover CTAs and the dedicated "something went wrong" state for Agent failures land in a follow-up that also revisits CANCELLED step states.

### Screenshots

Flow:

https://github.com/user-attachments/assets/3816020e-defd-4b13-829c-3d02a64afde0

### Related issue(s)

Related to #1848 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalWithdrawReview.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalWithdrawReview.tsx
@@ -135,12 +135,85 @@ const buildApprovalStep = (
   t: Translator,
 ) => ({
   description: (
-    <span>{t('step.approve-token', { symbol: shareToken.symbol })}</span>
+    <div className="flex items-center gap-x-2">
+      <TokenLogo size="small" token={shareToken} />
+      <span>{t('step.approve-token', { symbol: shareToken.symbol })}</span>
+    </div>
   ),
   explorerChainId: hemi.id,
   status: ProgressStatus.COMPLETED,
   txHash: approvalTxHash,
 })
+
+function deriveReceiveStatus({
+  cooldownRemainingSec,
+  isCooldownEligible,
+  receiveFromStatus,
+  unstakeMined,
+}: {
+  cooldownRemainingSec: number | undefined
+  isCooldownEligible: boolean | undefined
+  receiveFromStatus: ProgressStatusType
+  unstakeMined: boolean
+}): ProgressStatusType {
+  if (receiveFromStatus !== ProgressStatus.NOT_READY) return receiveFromStatus
+  if (!unstakeMined) return receiveFromStatus
+  const needsCooldown = isCooldownEligible !== false
+  const cooldownElapsed = cooldownRemainingSec === 0
+  return !needsCooldown || cooldownElapsed
+    ? ProgressStatus.PROGRESS
+    : receiveFromStatus
+}
+
+function buildSteps({
+  cooldownPostAction,
+  needsApproval,
+  pool,
+  receive,
+  receiveToken,
+  t,
+  transaction,
+}: {
+  cooldownPostAction: ReturnType<typeof deriveCooldownPostAction>
+  needsApproval: boolean
+  pool: EarnPool
+  receive: ProgressStatusType
+  receiveToken: EvmToken | undefined
+  t: Translator
+  transaction: EarnTransaction
+}): StepPropsWithoutPosition[] {
+  const steps: StepPropsWithoutPosition[] = []
+  if (
+    transaction.status === 'FAILED' &&
+    isLocalEarnTransactionRow(transaction) &&
+    needsApproval
+  ) {
+    steps.push({
+      description: (
+        <div className="flex items-center gap-x-2">
+          <TokenLogo size="small" token={pool.shareToken} />
+          <span>{t('step.approval-needed')}</span>
+        </div>
+      ),
+      status: ProgressStatus.NOT_READY,
+    })
+  }
+  if (transaction.approvalTxHash) {
+    steps.push(
+      buildApprovalStep(transaction.approvalTxHash, pool.shareToken, t),
+    )
+  }
+  const unstakeStep = buildUnstakeStep(transaction, pool.shareToken, t)
+  steps.push(
+    cooldownPostAction
+      ? { ...unstakeStep, postAction: cooldownPostAction }
+      : unstakeStep,
+  )
+  if (receiveToken) {
+    steps.push(buildReceiveStep(transaction, receiveToken, t, receive))
+  }
+  return steps
+}
 
 export const HistoricalWithdrawReview = function ({
   callToAction,
@@ -178,14 +251,19 @@ export const HistoricalWithdrawReview = function ({
     stakingVault: pool.stakingVault,
   })
 
+  const claimableAt = transaction.claimableAt ?? null
   const cooldownRemainingSec = useEarnCooldownRemaining(
-    transaction.claimableAt != null
-      ? BigInt(transaction.claimableAt)
-      : undefined,
+    claimableAt !== null ? BigInt(claimableAt) : undefined,
   )
 
-  const { receive, unstake } = resolveStepStates(transaction)
+  const { receive: receiveFromStatus, unstake } = resolveStepStates(transaction)
   const unstakeMined = unstake === ProgressStatus.COMPLETED
+  const receive = deriveReceiveStatus({
+    cooldownRemainingSec,
+    isCooldownEligible,
+    receiveFromStatus,
+    unstakeMined,
+  })
 
   const cooldownPostAction = deriveCooldownPostAction({
     cooldownDurationSec,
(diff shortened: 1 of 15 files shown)


portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/cooldownPostAction.ts
```diff
@@ -0,0 +1,102 @@
+import {
+  ProgressStatus,
+  type ProgressStatusType,
+} from 'components/reviewOperation/progressStatus'
+import { type useTranslations } from 'next-intl'
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/cooldownPostAction.ts:5]: Verified — this pattern is already in use in `columns.tsx:11` (merged earlier) and TypeScript accepts `typeof X` in a type-query position with a type-only import, since no value symbol is needed there. Local build passes. Leaving as-is.

gndelia:
In the first video, for a few seconds this is rendered

image


is this correct?

gndelia:
### Minor: cooldown reads may be unnecessary for terminal REDEEM rows

In `HistoricalWithdrawReview`, the three cooldown reads fire unconditionally on every drawer open:

```ts
const { data: isCooldownEligible } = useIsCooldownEligible({ account: address, stakingVault: pool.stakingVault })
const { data: cooldownDurationSec } = useCooldownDuration({ stakingVault: pool.stakingVault })
const cooldownRemainingSec = useEarnCooldownRemaining(
  transaction.claimableAt != null ? BigInt(transaction.claimableAt) : undefined,
)
```

**Why some are wasted:** for terminal rows (`FINALIZED` / `RECOVERED`, and arguably local `FAILED`), `deriveCooldownPostAction` short-circuits before it ever consumes `cooldownDurationSec` — `cooldownOver` is decided purely from `subgraphStatus`, returning a `COMPLETED` milestone. So the on-chain `useCooldownDuration` read (and to a lesser extent the eligibility read) does work whose result is thrown away. The cooldown sub-step only genuinely needs these values for *in-flight* rows (`FULFILLED` / `PENDING` waiting on the timer).

**How it could be fixed** — gate the reads on whether the row is still in a cooldown-relevant state, e.g.:

```ts
const needsCooldownReads =
  transaction.status !== 'FINALIZED' && transaction.status !== 'RECOVERED'

const { data: cooldownDurationSec } = useCooldownDuration({
  stakingVault: pool.stakingVault,
  query: { enabled: needsCooldownReads }, // or whatever `enabled` knob the hook exposes
})
```

This is non-blocking — it's a few extra reads, not a correctness bug, and the live `ReviewWithdraw` behaves the same way (though there it's always an in-flight row, so the reads aren't wasted). Flagging mostly so it's a conscious choice; happy to leave as-is if the hooks don't cleanly expose an `enabled` switch.

gndelia:
image

Should the sVETBTC logo be rendered in the first step? It is in the 2nd one

gndelia:
The "hemi" logo seems incorrect in the first step

image

(See this in your video)

Leaving this to Nahuel, but I think these 2 logos should use the Hemi version

image

as these balances are on Hemi side

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalWithdrawReview.tsx
```diff
@@ -0,0 +1,253 @@
...
+    account: address,
+    stakingVault: pool.stakingVault,
+  })
+
+  const { data: cooldownDurationSec } = useCooldownDuration({
+    stakingVault: pool.stakingVault,
+  })
+
+  const cooldownRemainingSec = useEarnCooldownRemaining(
+    transaction.claimableAt != null
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalWithdrawReview.tsx]:
```suggestion
    transaction.claimableAt !== null
```

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedWithdraw.tsx
```diff
@@ -0,0 +1,126 @@
...
+    assetOut > BigInt(0)
+      ? applySlippage(assetOut, REDEEM_SLIPPAGE_BPS)
+      : BigInt(0)
+
+  const { data: quote } = useQuoteRedeem({
+    account: address,
+    asset: asset.address,
+    shareAddress: pool.shareAddress,
+    shares,
+  })
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedWithdraw.tsx:71]: not for this PR, but if `account` is _always_ what comes from `useAccount`, maybe we can just read internally instead of being a parameter in every call

portal/app/[locale]/hemi-earn/_utils/index.ts
```diff
@@ -46,3 +46,8 @@ export const findPoolByShare = (
   shareAddress: Address,
 ): EarnPool | undefined =>
   pools.find(p => isAddressEqual(p.shareAddress, shareAddress))
+
+// Local rows mean the Hemi `request*` tx reverted; subgraph rows mean the
+// Agent failed after a successful Hemi tx (recover, not retry).
+export const isLocalEarnTransactionRow = (tx: EarnTransaction) =>
+  tx.requestId.startsWith('local-')
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_utils/index.ts:53]: it should be trivial to add some tests in the functions here

Please whenever unit tests are simple, let's try to add them. They help providing guardrails for future AI-related changes

dvnahuel:
> The "hemi" logo seems incorrect in the first step
> 
> image
> (See this in your video)
> 
> Leaving this to Nahuel, but I think these 2 logos should use the Hemi version
> 
> image
> as these balances are on Hemi side

Agree, let's add the Hemi network logo to those...

ArturDolzan:
> In the first video, for a few seconds this is rendered
> 
> image
> is this correct?

My contract has the cooldown as 3 minutes. That's why you see 0.

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedWithdraw.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedWithdraw.tsx:71]: Great point — opened #2011 to track it so we don't lose it. Both quote hooks (`useQuoteRedeem`/`useQuoteDeposit`) would benefit, all 6 call sites pass through `useAccount().address` today. Thanks!

ArturDolzan:
> ### Minor: cooldown reads may be unnecessary for terminal REDEEM rows
> In `HistoricalWithdrawReview`, the three cooldown reads fire unconditionally on every drawer open:
> 
> ```ts
> const { data: isCooldownEligible } = useIsCooldownEligible({ account: address, stakingVault: pool.stakingVault })
> const { data: cooldownDurationSec } = useCooldownDuration({ stakingVault: pool.stakingVault })
> const cooldownRemainingSec = useEarnCooldownRemaining(
>   transaction.claimableAt != null ? BigInt(transaction.claimableAt) : undefined,
> )
> ```
> 
> **Why some are wasted:** for terminal rows (`FINALIZED` / `RECOVERED`, and arguably local `FAILED`), `deriveCooldownPostAction` short-circuits before it ever consumes `cooldownDurationSec` — `cooldownOver` is decided purely from `subgraphStatus`, returning a `COMPLETED` milestone. So the on-chain `useCooldownDuration` read (and to a lesser extent the eligibility read) does work whose result is thrown away. The cooldown sub-step only genuinely needs these values for _in-flight_ rows (`FULFILLED` / `PENDING` waiting on the timer).
> 
> **How it could be fixed** — gate the reads on whether the row is still in a cooldown-relevant state, e.g.:
> 
> ```ts
> const needsCooldownReads =
>   transaction.status !== 'FINALIZED' && transaction.status !== 'RECOVERED'
> 
> const { data: cooldownDurationSec } = useCooldownDuration({
>   stakingVault: pool.stakingVault,
>   query: { enabled: needsCooldownReads }, // or whatever `enabled` knob the hook exposes
> })
> ```
> 
> This is non-blocking — it's a few extra reads, not a correctness bug, and the live `ReviewWithdraw` behaves the same way (though there it's always an in-flight row, so the reads aren't wasted). Flagging mostly so it's a conscious choice; happy to leave as-is if the hooks don't cleanly expose an `enabled` switch.

Good catch.. you're right that the reads are technically wasted for terminal rows. The reason I'd lean toward leaving as-is for now is that both hooks have `staleTime: 4h`, so it's a single governance read per session per `stakingVault`, not per drawer open. Opened #2012 to track it though, so we can revisit if the hooks ever need to expose an `enabled` knob for other reasons.

ArturDolzan:
> image
> Should the sVETBTC logo be rendered in the first step? It is in the 2nd one

Fixed on e34f81d5d64d534bcae55ada8d57002017026dd1. Thanks!

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalWithdrawReview.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalWithdrawReview.tsx]: Applied on e34f81d5d64d534bcae55ada8d57002017026dd1. Thanks!

portal/app/[locale]/hemi-earn/_utils/index.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/index.ts:53]: Added on e34f81d5d64d534bcae55ada8d57002017026dd1. Thanks!

ArturDolzan:
> The "hemi" logo seems incorrect in the first step
> 
> image
> (See this in your video)
> 
> Leaving this to Nahuel, but I think these 2 logos should use the Hemi version
> 
> image
> as these balances are on Hemi side

The Hemi logo in the first step is now fixed. As for the very small Hemi logo in the bottom right (the one indicating it's an L2 token), that's tied to my local fork config. In any case, I've uploaded a new video to the PR description.

gndelia (approved):

gabmontes (approved):

Issue #1977, opened 2026-06-02
Hemi Earn: surface manual claim/recover CTAs when auto-finalization fails or is opted out


ArturDolzan:
## Context

The Router auto-transfers tokens to the receiver inside `_handleRequestFulfillment` / `_handleRequestCancel` when the request was submitted with `automatic=true`. The transfer runs inside a `try/catch` — if it reverts (e.g. receiver is a contract without a fallback / `receive`), the Router emits `AutoFinalizationFailed(requestId, reason)` and the request stays in a non-finalized terminal-ish state (`FULFILLED` or `CANCELLED`) until someone calls `claim*` or `recover*` explicitly.

`automatic=false` requests follow the same path by design — they're never auto-finalized in the first place.

Today the portal assumes `automatic=true` is always self-finalizing, so the UI doesn't expose the manual escape hatches. That leaves users with funds locked at the Router and no way to retrieve them from the app.

## What needs to land

The portal already has the contract wrappers (`claimDeposit`/`claimRedeem`/`recoverDeposit`/`recoverRedeem` in `packages/hemi-earn-actions/src/actions/wallet/*.ts`). The subgraph also indexes `AutoFinalizationFailed` as `autoFinalizationFailed: Boolean!` on the `Request` entity (PR #1974). What's missing is the Activity surface that turns this state into a CTA the user can click.

### Trigger condition

Render a CTA on the Activity row when:

- `receiver === connectedAccount` (the contract reverts otherwise — `NotReceiver`)
- AND either:
  - `automatic === false`, OR
  - `automatic === true && autoFinalizationFailed === true`

### Two CTAs, same UI pattern

| Status | CTA label | Action | Token released |
|---|---|---|---|
| `FULFILLED` | Claim | `claimDeposit(id)` / `claimRedeem(id)` | shares (deposit) or asset (redeem) |
| `CANCELLED` | Recover | `recoverDeposit(id)` / `recoverRedeem(id)` | original asset (deposit) or shares (redeem) |

Both are single-tx Hemi-side writes (no allowance, no quote, no native fee) — same `SettlementEvents` vocabulary from `hemi-earn-actions/types.ts`. A single component parameterized by `(kind, status)` covers both.

### Subgraph query for the Activity bucket

```graphql
{
  requests(where: {
    receiver: $account,
    status_in: [FULFILLED, CANCELLED],
    or: [
      { automatic: false },
      { autoFinalizationFailed: true }
    ]
  }) {
[...]

PR #2019, opened 2026-06-19
Add manual claim/recover to Hemi Earn deposits

ArturDolzan:
### Description

This PR implements the manual claim and recover flows for Hemi Earn deposits. When a deposit has auto-finalize off, it stops at FULFILLED (the user signs the claim to receive shares) or CANCELLED (the user signs the recover to pull the original asset back) — both need a Router tx on Hemi that the UI wasn't offering yet. Now the table row and both the live and historical drawers surface a Claim / Recover CTA, and the user signs it right there.

  If that signed tx reverts on Hemi, the attempt is tracked on the existing `hemi-earn:local-operations` entry so the drawer offers a "Try again" instead of silently dropping back to the idle state. That marker is reconciled against the authoritative subgraph status in the merge and dropped once the row reaches FINALIZED/RECOVERED, so a finished deposit never flickers its CTA back during the indexing lag or shows a stale "Tx Failed".

  There's also a status fix: a reverted Agent step leaves a `failed` flag the indexer never clears on cancel, which was masking a recoverable (CANCELLED) deposit as FAILED and hiding the recover CTA. `deriveStatus` now lets a terminal Router status win over a still-PENDING failure. This is a temporary portal-side bridge — once the Envio endpoint PR lands, this translation moves to portal-backend.

  On the subgraph, CANCELLED is demoted from terminal rank so it can transition to RECOVERED, and the transactions query keeps polling a cancelled deposit until it recovers (otherwise it could get stuck on CANCELLED after the user's recover).

  Tested locally end-to-end against the anvil + Envio sandbox (claim, recover, retry-on-revert, and the full CANCELLED → RECOVERED cycle).

### Screenshots

Claim

https://github.com/user-attachments/assets/2eeab88b-d1ba-42e2-8fd6-23fec232cd0a

Recover

https://github.com/user-attachments/assets/523dcf3b-2527-4673-bd0a-ff5cf6254c82

### Related issue(s)

Related to #1848 
Related to #1977 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
@@ -13,7 +13,12 @@ import Skeleton from 'react-loading-skeleton'
 import { formatUnits } from 'viem'
 
 import { useEarnPools } from '../../_hooks/useEarnPools'
-import { findPoolByAsset } from '../../_utils'
+import {
+  findPoolByAsset,
+  hasFailedSettlement,
+  needsManualClaim,
+  needsRecover,
+} from '../../_utils'
 import { type EarnTransaction } from '../../types'
 
 import { RowActions } from './rowActions'
@@ -104,18 +109,27 @@ export const buildColumns = ({
       row.original.kind === 'REDEEM' ? (
         <WithdrawStatusCell transaction={row.original} />
       ) : (
-        <StatusBadge kind={row.original.kind} status={row.original.status} />
+        <StatusBadge
+          kind={row.original.kind}
+          manualClaimNeeded={needsManualClaim(row.original)}
+          manualRecoverNeeded={needsRecover(row.original)}
+          settlementFailed={hasFailedSettlement(row.original)}
+          status={row.original.status}
+        />
       ),
     header: () => <Header text={t('column.status')} />,
     id: 'status',
-    meta: { className: 'justify-start flex-grow-0', width: 190 },
+    // Wide enough to keep the longest status ("Something went wrong - Recover
+    // funds" / the manual-claim copy) on a single line across locales.
+    meta: { className: 'justify-start flex-grow-0', width: 340 },
   },
   {
     cell: ({ row }) => <RowActions transaction={row.original} />,
     id: 'actions',
     // On mobile this column is reordered to the front, so it hugs the left
     // edge (the table's first-cell pl-4 gives the 16px lead). From 'md' up it
-    // returns to the trailing edge and right-aligns.
-    meta: { className: 'justify-start md:justify-end', width: 90 },
+    // returns to the trailing edge and right-aligns. Wide enough to fit the
+    // "Claim share tokens" / "Recover funds" CTA without overflowing the cell.
+    meta: { className: 'justify-start md:justify-end', width: 180 },
   },
 ]
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
@@ -13,12 +13,7 @@ import Skeleton from 'react-loading-skeleton'
 import { formatUnits } from 'viem'
 
 import { useEarnPools } from '../../_hooks/useEarnPools'
-import {
-  findPoolByAsset,
-  hasFailedSettlement,
-  needsManualClaim,
-  needsRecover,
-} from '../../_utils'
+import { findPoolByAsset } from '../../_utils'
 import { type EarnTransaction } from '../../types'
 
 import { RowActions } from './rowActions'
@@ -109,13 +104,7 @@ export const buildColumns = ({
       row.original.kind === 'REDEEM' ? (
         <WithdrawStatusCell transaction={row.original} />
       ) : (
-        <StatusBadge
-          kind={row.original.kind}
-          manualClaimNeeded={needsManualClaim(row.original)}
-          manualRecoverNeeded={needsRecover(row.original)}
-          settlementFailed={hasFailedSettlement(row.original)}
-          status={row.original.status}
-        />
+        <StatusBadge transaction={row.original} />
       ),
     header: () => <Header text={t('column.status')} />,
     id: 'status',
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx
@@ -4,41 +4,28 @@ import { Button, ButtonIcon } from 'components/button'
 import { useTranslations } from 'next-intl'
 import { type MouseEvent } from 'react'
 
-import { needsManualClaim, needsRecover } from '../../_utils'
-import {
-  type EarnTransaction,
-  type EarnTransactionStatusType,
-} from '../../types'
+import { isEarnRowInFlight, needsManualClaim, needsRecover } from '../../_utils'
+import { type EarnTransaction } from '../../types'
 import { LoaderIcon } from '../icons/loaderIcon'
 import { TrashIcon } from '../icons/trashIcon'
 
 import { DepositRowCta } from './transactionDrawer/settleDeposit'
 import { useTxDrawerQueryString } from './transactionDrawer/useTxDrawerQueryString'
 
-const TERMINAL_STATUSES: EarnTransactionStatusType[] = [
-  'CANCELLED',
-  'FAILED',
-  'FINALIZED',
-  'RECOVERED',
-]
-
 type Props = {
   transaction: EarnTransaction
 }
 
-// Spinner while the row is in flight: a mining claim/recover settlement (even on
-// a CANCELLED row, which the status set treats as terminal) or a non-terminal
-// status. A reverted settlement shows no spinner — it's "Tx Failed" with a
-// Retry. The inline Claim/Recover CTA is only for the untouched manual state;
-// once a settlement exists (mining or reverted) the row hands back to View and
-// the drawer carries the spinner/Retry.
+// Spinner while the row is in flight (auto-progressing, including an auto-recover
+// CANCELLED deposit, or a claim/recover settlement tx mining), but never on a
+// reverted settlement — that's "Tx Failed" with a Retry. The inline
+// Claim/Recover CTA is only for the untouched manual state; once a settlement
+// exists (mining or reverted) the row hands back to View and the drawer carries
+// the spinner/Retry.
 function resolveActionState(transaction: EarnTransaction) {
   const { settlement } = transaction
-  const settlementPending = !!settlement && !settlement.failed
   return {
-    showLoaderIcon:
-      settlementPending ||
-      (!TERMINAL_STATUSES.includes(transaction.status) && !settlement?.failed),
+    showLoaderIcon: isEarnRowInFlight(transaction) && !settlement?.failed,
     showManualCta:
       !settlement &&
       (needsManualClaim(transaction) || needsRecover(transaction)),
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx
@@ -4,28 +4,53 @@ import { Button, ButtonIcon } from 'components/button'
 import { useTranslations } from 'next-intl'
 import { type MouseEvent } from 'react'
 
-import { type EarnTransaction } from '../../types'
+import { needsManualClaim, needsRecover } from '../../_utils'
+import {
+  type EarnTransaction,
+  type EarnTransactionStatusType,
+} from '../../types'
 import { LoaderIcon } from '../icons/loaderIcon'
 import { TrashIcon } from '../icons/trashIcon'
 
+import { DepositRowCta } from './transactionDrawer/settleDeposit'
 import { useTxDrawerQueryString } from './transactionDrawer/useTxDrawerQueryString'
+
+const TERMINAL_STATUSES: EarnTransactionStatusType[] = [
+  'CANCELLED',
+  'FAILED',
+  'FINALIZED',
+  'RECOVERED',
+]
 
 type Props = {
   transaction: EarnTransaction
 }
 
+// Spinner while the row is in flight: a mining claim/recover settlement (even on
+// a CANCELLED row, which the status set treats as terminal) or a non-terminal
+// status. A reverted settlement shows no spinner — it's "Tx Failed" with a
+// Retry. The inline Claim/Recover CTA is only for the untouched manual state;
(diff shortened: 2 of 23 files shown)


portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx
```diff
@@ -4,28 +4,53 @@ import { Button, ButtonIcon } from 'components/button'
...
+  const { settlement } = transaction
+  const settlementPending = !!settlement && !settlement.failed
+  return {
+    showLoaderIcon:
+      settlementPending ||
+      (!TERMINAL_STATUSES.includes(transaction.status) && !settlement?.failed),
+    showManualCta:
+      !settlement &&
+      (needsManualClaim(transaction) || needsRecover(transaction)),
+  }
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/rowActions.tsx]: Good catch — fixed in `78432232`. The loader now shares an `isEarnRowInFlight` check with the polling loop, so an auto-recover CANCELLED deposit shows the spinner while it's still progressing. Thanks!

portal/app/[locale]/hemi-earn/_hooks/useEarnTransactionsQuery.ts
```diff
@@ -41,9 +43,11 @@ export const useEarnTransactionsQuery = function () {
       const hasSubgraphInFlight = subgraphData.some(
         t =>
           t.status !== 'FINALIZED' &&
-          t.status !== 'CANCELLED' &&
           t.status !== 'RECOVERED' &&
-          t.status !== 'FAILED',
+          t.status !== 'FAILED' &&
+          // A DEPOSIT CANCELLED still walks to RECOVERED; only a REDEEM
+          // CANCELLED is terminal (withdrawal canceled).
+          !(t.status === 'CANCELLED' && t.kind === 'REDEEM'),
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useEarnTransactionsQuery.ts]: Thought about this one — I'm leaning toward keeping CANCELLED deposits polling. They're the user's funds awaiting recovery (auto or manual), same as a FULFILLED deposit awaiting claim, which already polls; stopping it would make the table go stale when the recovery lands (including from another device), and the cost is modest (one query/10s while the page is open).

gndelia (commented):
(Unless they were already added to your wallet)

Double check that the "Add token to wallet" appears after claiming the shares, both automatically and at the bottom of the drawer as we do in the tunnel history (I did not see it in the video)

gndelia:
On the "Recover funds", after the TX is confirmed, and you click on `manage`, it takes a while for the correct balance to appear. IS some invalidation failing, or an optimistic update missing?

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
```diff
@@ -98,20 +104,75 @@ function buildStakeStep(tx: EarnTransaction, token: EvmToken, t: Translator) {
...
+  baseStatus: ProgressStatusType
+  settlementTxHash: Hash | undefined
+  t: Translator
+  token: EvmToken
+  tx: EarnTransaction
+}): StepPropsWithoutPosition {
+  const status = resolveTerminalStatus(tx, baseStatus, settlementTxHash)
+  const txHash = settlementTxHash ?? getTerminalDeliveryTxHash(tx)
+  // Only a manual claim/recover has a user-signed tx worth linking.
+  const explorerChainId = tx.automatic === false ? hemi.id : undefined
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx]: I would link all transactions that are available, even if the user did not sign them. In the tunnel, you get the links to the automatic transactions that happen when minting tokens on the remote chain. If the tx hashes are available, I'd just show them

They are also useful if, for some reason, the user can't see the funds

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx
```diff
@@ -7,18 +7,50 @@ import { isAddressEqual } from 'viem'
...
 
+// The drawer is already open, so the deposit CTAs don't redirect on sign. A
+// `failedKind` (the user's claim/recover Hemi tx reverted) surfaces the CTA as
+// a Retry even though the subgraph status no longer matches.
+function depositCallToAction(
+  asset: EarnAsset,
+  pool: EarnPool,
+  transaction: EarnTransaction,
+  failedKind: 'CLAIM' | 'RECOVER' | undefined,
+) {
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx]: should probably use an object instead of 4 parameters

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx
```diff
@@ -7,18 +7,50 @@ import { isAddressEqual } from 'viem'
...
+  }
+  if (needsManualClaim(transaction) || failedKind === 'CLAIM') {
+    return <ClaimDeposit asset={asset} pool={pool} transaction={transaction} />
+  }
+  if (needsRecover(transaction) || failedKind === 'RECOVER') {
+    return (
+      <RecoverDeposit asset={asset} pool={pool} transaction={transaction} />
+    )
+  }
+  return undefined
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx]: Nit: Components in react usually return `null` when rendering nothing (`undefined` works as well though)

```suggestion
  return null
```

portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
```diff
@@ -104,18 +109,27 @@ export const buildColumns = ({
...
         <WithdrawStatusCell transaction={row.original} />
       ) : (
-        <StatusBadge kind={row.original.kind} status={row.original.status} />
+        <StatusBadge
+          kind={row.original.kind}
+          manualClaimNeeded={needsManualClaim(row.original)}
+          manualRecoverNeeded={needsRecover(row.original)}
+          settlementFailed={hasFailedSettlement(row.original)}
+          status={row.original.status}
+        />
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx]: seems to me that a simpler api would be passing `row.original` to `StatusBadge` and access `status`, `kind` and call the 3 functions inside

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
```diff
@@ -164,6 +171,27 @@ export const LocalEarnOperationsProvider = function ({
...
+          mutated = true
+          return { ...e, settlement }
+        })
+        if (!mutated) return prev
+        return { ...prev, [targetAccount]: updated }
+      })
+    },
+    [address, setStore],
+  )
+
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:194]: this is getting more and more complex. Would we need all of this once the endpoint is integrated? or can we handle everything with the subgraph's reading + onchain data?

portal/app/[locale]/hemi-earn/_hooks/useEarnTransactionsQuery.ts
```diff
@@ -38,14 +38,9 @@ export const useEarnTransactionsQuery = function () {
...
       const subgraphData = (query.state.data ?? []) as EarnTransaction[]
-      const hasSubgraphInFlight = subgraphData.some(
-        t =>
-          t.status !== 'FINALIZED' &&
-          t.status !== 'CANCELLED' &&
-          t.status !== 'RECOVERED' &&
-          t.status !== 'FAILED',
-      )
-      return hasSubgraphInFlight || inFlightLocalsExist ? 10_000 : false
+      return subgraphData.some(isEarnRowInFlight) || inFlightLocalsExist
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_hooks/useEarnTransactionsQuery.ts]: Nit: I would reverse this

```suggestion
      return inFlightLocalsExist || subgraphData.some(isEarnRowInFlight)
```

faster to check a boolean that an array of data 😅  those nanoseconds will make the difference (?)

portal/app/[locale]/hemi-earn/_hooks/useSettleDeposit.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  return useMutation({
+    async mutationFn() {
+      if (!address) {
+        throw new Error('No account connected')
+      }
+
+      await ensureConnectedTo(hemi.id)
+
+      const walletClient = await getWalletClient(config, { chainId: hemi.id })
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_hooks/useSettleDeposit.ts]: There's `useHemiWalletClient` which does this, so we access the same wallet instance as the whole app. I think this will create a new one..

portal/app/[locale]/hemi-earn/_hooks/useSettleDeposit.ts
```diff
@@ -0,0 +1,129 @@
...
+  // Claim drops shares into the user's wallet; recover returns the original
+  // asset. Refresh whichever lands.
+  const deliveredBalanceQueryKey = getTokenBalanceQueryKey({
+    account: address,
+    chainId: hemi.id,
+    tokenAddress: kind === 'CLAIM' ? pool.shareAddress : asset.address,
+  })
+
+  return useMutation({
+    async mutationFn() {
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_hooks/useSettleDeposit.ts:70]: I presume here is missint what I mentioned [in this comment](https://github.com/hemilabs/ui-monorepo/pull/2019#issuecomment-4752456197)

gabmontes (dismissed):

portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_context/localEarnOperationsContext.tsx:194]: Yeah, fair — it grew. Some of it can definitely be slimmed: most of this is local-only because the subgraph can't give us "just-signed, not yet indexed" or "the claim/recover I just watched revert" in real time. But the bits we'd drop — the `settlement` overlay ("Tx Failed"/"Try again"), the optimistic in-flight rows, the mining spinner — are real UX feedback, not just complexity, so losing them is a trade-off rather than free. The only piece that's genuinely impossible from subgraph + onchain is the FAILED request-deposit retry: a reverted `requestDeposit` never reaches the subgraph, so only the local entry has it.

Since some of that is UX, I'd loop in Nahuel on what's worth keeping vs simplifying, and open an issue to scope the trim together with design — rather than balloon this PR. Want me to open one?

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx]: Good idea — done in `decbef83`. Dropped the `automatic`-only gate, so the terminal step links the claim/recover hash whenever one's available (auto-finalized ones included), in both the live and historical drawers.

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx]: Done in `decbef83` — takes a single object now.

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx]: Done in `decbef83`, thanks.

portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx]: Nice — done in `decbef83`. `StatusBadge` takes the `transaction` now and derives the kind/status + predicates internally, so the callers stop threading them.

portal/app/[locale]/hemi-earn/_hooks/useEarnTransactionsQuery.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useEarnTransactionsQuery.ts]: Ha, fair — done in `decbef83`, the local in-flight check goes first now.

portal/app/[locale]/hemi-earn/_hooks/useSettleDeposit.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useSettleDeposit.ts]: Good call — switched to `useHemiWalletClient` in `decbef83` so it shares the app's wallet instance instead of spinning up a new client. Thanks!

ArturDolzan:
> On the "Recover funds", after the TX is confirmed, and you click on `manage`, it takes a while for the correct balance to appear. IS some invalidation failing, or an optimistic update missing?

Yeah.. the invalidation was there but it was `invalidateQueries`, so the deposit form (it mounts when you hit "manage" after recovering) flashed the stale balance before the background refetch. Switched the delivered balance to `resetQueries` in `decbef83`.

ArturDolzan:
> (Unless they were already added to your wallet)
> 
> Double check that the "Add token to wallet" appears after claiming the shares, both automatically and at the bottom of the drawer as we do in the tunnel history (I did not see it in the video)

On the "Add token to wallet": the bottom-of-drawer CTA was already in the live drawer (on FINALIZED) but was missing from the historical one.. added it there too in `decbef83` (shared `AddShareTokenToWallet`), so it matches the tunnel history now.

portal/app/[locale]/hemi-earn/_hooks/useSettleDeposit.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useSettleDeposit.ts:70]: Fixes on decbef83. Thanks!

gabmontes (approved):

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
```diff
@@ -98,20 +104,76 @@ function buildStakeStep(tx: EarnTransaction, token: EvmToken, t: Translator) {
...
+  settlementTxHash: Hash | undefined
+  t: Translator
+  token: EvmToken
+  tx: EarnTransaction
+}): StepPropsWithoutPosition {
+  const status = resolveTerminalStatus(tx, baseStatus, settlementTxHash)
+  const txHash = settlementTxHash ?? getTerminalDeliveryTxHash(tx)
+  // Link whenever a delivery hash is available — an auto-finalized claim/recover
+  // has one too, even though the user didn't sign it.
+  const explorerChainId = txHash ? hemi.id : undefined
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx:146]: https://github.com/hemilabs/ui-monorepo/pull/2019#discussion_r3443139442

Either this was duplicated or wasn't applied. I don't think we need the conditional

```suggestion
  const explorerChainId = hemi.id
```

just always provide the explorer id

gndelia (approved):
uACK

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx:146]: will review it in the upcoming PR. thanks!

PR #2023, opened 2026-06-19
Fix network selector row not being clickable

### LINKED ISSUE #1957: Network selector doesn't change unless clicking the network name
### Description

When selecting an option in the network selector, it should work when clicking the entire row (that's shown as hovered) - not just when clicking "Mainnet" or "Testnet".

See video

https://github.com/user-attachments/assets/11ec26e5-ba56-4ff8-bdbb-eb6a3d579db4

arielmergen:
### Description

  Fixes the Mainnet/Testnet network selector so the entire row is clickable,
  not just the network name.

  The dropdown rows show the hover background on the full-width `<li>`, but the
  `onClick` lived on the inner `<button>`, which is sized to its content. So
  clicking the hovered row anywhere except on the "Mainnet"/"Testnet" label did
  nothing. Added `w-full` to the button so its click target spans the full row
  width, matching the visible hover area.

### Screenshots

https://github.com/user-attachments/assets/26435d80-2637-4c32-a442-eb9c81c84037

### Related issue(s)

Closes #1957 
Fixes #
Related to #

### Checklist

- [ ] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx
+++ b/portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx
@@ -141,7 +141,7 @@ const NetworkSwitchImpl = function ({
                 return {
                   content: (
                     <button
-                      className={`flex w-full items-center gap-x-2 ${
+                      className={`-mx-2 -my-1 flex w-full items-center gap-x-2 px-2 py-1 ${
                         selected ? 'text-neutral-950' : ''
                       }`}
                       disabled={selected}
--- a/portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx
+++ b/portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx
@@ -141,7 +141,7 @@ const NetworkSwitchImpl = function ({
                 return {
                   content: (
                     <button
-                      className={`-mx-2 -my-1 flex w-full items-center gap-x-2 px-2 py-1 ${
+                      className={`flex w-full items-center gap-x-2 ${
                         selected ? 'text-neutral-950' : ''
                       }`}
                       disabled={selected}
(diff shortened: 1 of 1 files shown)


portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx
```diff
@@ -141,7 +141,7 @@ const NetworkSwitchImpl = function ({
                 return {
                   content: (
                     <button
-                      className={`flex items-center gap-x-2 ${
+                      className={`flex w-full items-center gap-x-2 ${
                         selected ? 'text-neutral-950' : ''
                       }`}
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx]: Good catch. Addressed in a317f76 by expanding the button hit area into the `<li>` padding with negative margins (`-mx-2 -my-1`) plus matching `px-2 py-1`, so the whole hovered row is now clickable.

portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx
```diff
@@ -141,7 +141,7 @@ const NetworkSwitchImpl = function ({
                 return {
                   content: (
                     <button
-                      className={`flex items-center gap-x-2 ${
+                      className={`-mx-2 -my-1 flex w-full items-center gap-x-2 px-2 py-1 ${
```
Comment (gndelia) [inline portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx]: I think margin and padding are not needed for this change

```suggestion
                      className={`flex w-full items-center gap-x-2 ${
```

this should be enough (but verify!)

Comment (arielmergen) [reply portal/app/[locale]/_components/navbar/_components/networkSwitch.tsx]: Done — kept just `w-full`. The button now spans the full row width, so the whole row is clickable. The only residual is the `<li>`'s `px-2 py-1` padding strip (hovered but not clickable), which is negligible in practice.

gndelia (approved):

gabmontes (approved):

ArturDolzan (approved):

PR #2025, opened 2026-06-22
Fix wrong deposit amount in Hemi Earn table

### LINKED ISSUE #2024: Hemi Earn: transactions table shows wrong deposit amount (shares instead of asset)
### Description

In the **My earn transactions** table, the `Amount` column shows a wrong value for a **finalized deposit**. The table renders e.g. `1,000,000 hemiBTC` while the **Stake deposit details** drawer correctly shows `0.0001 hemiBTC` for the very same transaction. The drawer value is the correct one.

**Root cause.** The table's `AmountCell` (`portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx`) selects the raw amount with:

```ts
const rawAmount = showingShares
  ? transaction.amountIn
  : transaction.amountOut ?? transaction.amountIn
```

For a finalized **DEPOSIT**, `amountOut` is populated with the **minted share tokens** (sVetToken / svetBTC, **18 decimals**), not the deposited asset. The `amountOut ?? amountIn` fallback was only meant for REDEEM (where post-FULFILLED `amountOut` is the asset received). So the cell grabs the 18-decimal share amount but formats it against the **8-decimal asset token** (hemiBTC):

`formatUnits(100000000000000, 8) → 1,000,000` instead of `formatUnits(10000, 8) → 0.0001`.

The drawer is correct because `historicalDepositReview.tsx` always renders `amount={transaction.amountIn}` with the asset token.

**Confirmed against real data** (tx `0xaec5daf1e86d1bbc0ea064e2301b1d507a11706042df401a6b0df0145f47d377`, requestId `15`):
- Production API returns `amountIn=10000`, `amountOut=100000000000000`, asset `0xaa40…ef28` (hemiBTC, 8 dec).
- On-chain `Transfer` and Router `DepositRequested.assets` = `0x2710` = `10000` = **0.0001 hemiBTC**. ✓

**Scope.** The withdraw (REDEEM) path is **not** affected — both the table and the withdraw drawer already pair the right amount with the right token in both phases (shares before FULFILLED, asset after).

### Steps to reproduce

1. Make a Hemi Earn deposit (e.g. `0.0001 hemiBTC`) and let it reach the `Deposited` (FINALIZED) status.
2. Open the **My earn transactions** table.
3. Observe the `Amount` column shows a value ~10^10× too large (e.g. `1,000,000 hemiBTC`).
4. Open the row's **Stake deposit details** drawer — `Total Amount` correctly shows `0.0001 hemiBTC`.

### Environment

- **OS**: macOS
- **Runtime/Browser/Version**: portal web app
- **Wallet**: any

### Proposed fix

Extract the amount/token selection into a pure helper `pickEarnRowAmount(transaction, { assetToken, shareToken })` and make the **DEPOSIT** branch always use `amountIn` + the asset token; keep REDEEM behavior unchanged (shares + shareToken until FULFILLED, then `amountOut` + assetToken). Add unit tests covering the deposit regression (deposit with `amountOut` populated) and both REDEEM phases.

Files:
- `portal/app/[locale]/hemi-earn/_utils/index.ts`
- `portal/app/[locale]/hemi-earn/_components/transactionsSection/columns.tsx`
- `portal/test/app/[locale]/hemi-earn/_utils/index.test.ts`

ArturDolzan:
### Description

This PR fixes the Hemi Earn transactions table showing a wrong amount for finalized deposits.. it rendered values about 10^10x too large (e.g. 1,000,000 hemiBTC instead of the deposited 0.0001), even though the Stake deposit details drawer showed the correct value all along.

  The root cause was in the table's amount cell: it picked the raw amount with `amountOut ?? amountIn`. For a finalized deposit, `amountOut` is the minted share tokens (sVetToken, 18 decimals), not the deposited asset, so the cell took the 18-decimal share amount and formatted it against the 8-decimal asset token. That fallback only makes sense for redeems, where the post-fulfilled `amountOut` really is the asset received. The drawer was fine because it always reads `amountIn`. Confirmed against the reported tx (production API returns `amountIn=10000`, `amountOut=100000000000000`; on-chain the transfer and the Router event are `10000` = 0.0001 hemiBTC).

  The fix moves the amount/token selection into a small pure helper `pickEarnRowAmount` so a deposit always shows `amountIn` with the asset token, while the redeem behavior (shares + share token until fulfilled, then the returned asset) stays exactly as before. The withdraw path was already correct and is left untouched. Added unit tests for the deposit regression and both redeem phases.

### Screenshots

No UI changes.

### Related issue(s)

Closes #2024 
Fixes #2024 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_utils/index.ts
+++ b/portal/app/[locale]/hemi-earn/_utils/index.ts
@@ -1,3 +1,4 @@
+import { type Token } from 'types/token'
 import { formatPercentage } from 'utils/format'
 import { type Address, isAddressEqual } from 'viem'
 
@@ -84,6 +85,22 @@ export const canRetryRow = (tx: EarnTransaction) =>
 // misleading "needed" state.
 export const hasFailedSettlement = (tx: EarnTransaction) =>
   tx.settlement?.failed === true
+
+// Picks the amount + token to render for an earn transaction row. A DEPOSIT
+// always shows the deposited asset (`amountIn`, asset units) — its `amountOut`
+// is the minted share amount (sVetToken, 18 decimals) and must never be rendered
+// against the 8-decimal asset token. A REDEEM shows shares (`amountIn` +
+// shareToken) until FULFILLED, then the returned asset (`amountOut` + assetToken).
+export const pickEarnRowAmount = (
+  transaction: EarnTransaction,
+  { assetToken, shareToken }: { assetToken?: Token; shareToken?: Token },
+): { rawAmount: string; token: Token | undefined } =>
+  transaction.kind === 'REDEEM'
+    ? {
+        rawAmount: transaction.amountOut ?? transaction.amountIn,
+        token: transaction.amountOut == null ? shareToken : assetToken,
+      }
+    : { rawAmount: transaction.amountIn, token: assetToken }
 
 // A row the table is still actively tracking — drives both the polling loop and
 // the row spinner. Terminal: FINALIZED/RECOVERED/FAILED for any kind, plus a
--- a/portal/app/[locale]/hemi-earn/_utils/index.ts
+++ b/portal/app/[locale]/hemi-earn/_utils/index.ts
@@ -99,7 +99,7 @@ export const pickEarnRowAmount = (
   transaction.kind === 'REDEEM'
     ? {
         rawAmount: transaction.amountOut ?? transaction.amountIn,
-        token: transaction.amountOut == null ? shareToken : assetToken,
+        token: transaction.amountOut === null ? shareToken : assetToken,
       }
     : { rawAmount: transaction.amountIn, token: assetToken }
 
--- a/portal/app/[locale]/hemi-earn/_utils/index.ts
+++ b/portal/app/[locale]/hemi-earn/_utils/index.ts
@@ -90,7 +90,8 @@ export const hasFailedSettlement = (tx: EarnTransaction) =>
 // always shows the deposited asset (`amountIn`, asset units) — its `amountOut`
 // is the minted share amount (sVetToken, 18 decimals) and must never be rendered
 // against the 8-decimal asset token. A REDEEM shows shares (`amountIn` +
-// shareToken) until FULFILLED, then the returned asset (`amountOut` + assetToken).
+// shareToken) while `amountOut` is unset, then the returned asset (`amountOut` +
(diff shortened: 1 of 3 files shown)


portal/app/[locale]/hemi-earn/_utils/index.ts
```diff
@@ -85,6 +86,22 @@ export const canRetryRow = (tx: EarnTransaction) =>
 export const hasFailedSettlement = (tx: EarnTransaction) =>
   tx.settlement?.failed === true
 
+// Picks the amount + token to render for an earn transaction row. A DEPOSIT
+// always shows the deposited asset (`amountIn`, asset units) — its `amountOut`
+// is the minted share amount (sVetToken, 18 decimals) and must never be rendered
+// against the 8-decimal asset token. A REDEEM shows shares (`amountIn` +
+// shareToken) until FULFILLED, then the returned asset (`amountOut` + assetToken).
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/index.ts]: Good point — reworded in `869b75e9` to describe the actual `amountOut` check instead of leaning on "FULFILLED". Thanks!

portal/app/[locale]/hemi-earn/_utils/index.ts
```diff
@@ -85,6 +86,23 @@ export const canRetryRow = (tx: EarnTransaction) =>
...
+// shareToken) while `amountOut` is unset, then the returned asset (`amountOut` +
+// assetToken) once it's populated at fulfillment.
+export const pickEarnRowAmount = (
+  transaction: EarnTransaction,
+  { assetToken, shareToken }: { assetToken?: Token; shareToken?: Token },
+): { rawAmount: string; token: Token | undefined } =>
+  transaction.kind === 'REDEEM'
+    ? {
+        rawAmount: transaction.amountOut ?? transaction.amountIn,
+        token: transaction.amountOut == null ? shareToken : assetToken,
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_utils/index.ts]:
```suggestion
        token: transaction.amountOut === null ? shareToken : assetToken,
```

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/index.ts]: Sure thing — applied in `b73e5e3b`. Thanks!

gndelia (dismissed):

gabmontes (dismissed):

gabmontes (approved):

gndelia (approved):

Issue #1993, opened 2026-06-11
Create storybook for base components


gndelia:
### Feature or change description

In order to allow IA to better interact with creating new base components, or updating existing ones, we should use a storybook.

LET's build one for the [components](https://github.com/hemilabs/ui-monorepo/tree/main/portal/components).

I don't think all components should be on Storybook - I'd go for the simplest. A list could be

- [x] Button
- [x] ButtonLoader
- [x] ChainLogo
- [x] Drawer
- [x] Heading styles (not really a component)
- [x] Link
- [x] Spinner
- [x] Table
- [x] Tabs
- [x] Toast
- [ ] TokenInput
- [x] Tooltip
- [ ] TxLink
- [x] VerticalLines
- [x] Warning box

(open to other suggestions)

I'd recommend separating this into a few PRs, like the first one doing the setup + 1 component to verify it works, and then subsequent PRs implementing multiple components

Let's just focus on implementing the stories, and we can later focus on improvements or fixing inconsistencies accross the app

dvnahuel:
Toast component [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=20261-30873&t=cRKBfQOjHKZkTSEe-1) 

cc: @arielmergen

arielmergen:
The Table story is covered: it landed on `main` as part of #2129, with `Default`, `Loading` and `Empty` stories mirroring the staking dashboard table. That PR also brought the shared tweaks the story relied on, the `xxSmall` Button size and the `font-medium` table `Header`. Because of that, the draft PR #2061 is fully superseded and is being closed. What is left on this list is TokenInput and TxLink.

PR #2053, opened 2026-06-30
Add Tabs story to Storybook

arielmergen:
### Description

Adds a Storybook story for the `Tabs`/`Tab` components (`components/tabs`).

The `Default` story is based on a real usage in the app (the Active/Withdrawn filter in
`stakeTableFilter`, and `tunnelProviderToggle`): an interactive set of tabs using the `onClick`
variant with a `selected` state, so clicking a tab selects it. The link (`href`) variant looks
identical and only differs in navigation, so the story uses the router-free `onClick` variant. The
components are not modified.

- New `stories/tabs.stories.tsx` — an interactive `Default` story (state-driven selection).

### Screenshots

components-tabs--default--ui

components-tabs--default--withdrawn

components-tabs--default

### Related issue(s)

Related to #1993

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/stories/tabs.stories.tsx
+++ b/portal/stories/tabs.stories.tsx
@@ -1,30 +1,44 @@
 import type { Meta, StoryObj } from '@storybook/nextjs'
 import { Tab, Tabs } from 'components/tabs'
+import type { ComponentProps } from 'react'
 import { useState } from 'react'
 
+type StoryProps = ComponentProps<typeof Tabs> & {
+  size: ComponentProps<typeof Tab>['size']
+}
+
 const meta = {
-  component: Tabs,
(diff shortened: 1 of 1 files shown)


gabmontes (dismissed):

gndelia (commented):
Tabs have different sizes, so you. may want to add that control option

arielmergen:
Added a `size` control in 6772b8b6 — it's an inline radio (`xSmall` / `small`) wired to both tabs, so you can switch the tab size from the panel. Thanks @gndelia!

ArturDolzan (approved):

gndelia (approved):

PR #2058, opened 2026-07-02
Add Storybook story for Toast

arielmergen:
### Description

Adds a Storybook story for the `Toast` component (`components/toast`).

The `Default` story renders the toast with editable Controls (`title`, `description`, `variant`,
`autoCloseMs`), with `autoCloseMs` set to `0` so it stays visible in the canvas. Two extra stories
cover the variants that use link props, mirroring real app usage: `WithTransaction` shows the `tx`
external link (as in the deposit/withdraw toasts), and `WithGoTo` shows the `goTo` internal link
alongside `tx` (as in `StakeToast`, which links to the staking dashboard). The component is not
modified.

### Screenshots

components-toast--default--ui

components-toast--default--ui

### Related issue(s)

Related to #1993

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/stories/toast.stories.tsx
+++ b/portal/stories/toast.stories.tsx
@@ -46,3 +46,24 @@ export const WithTransaction: Story = {
     />
   ),
 }
+
+export const WithGoTo: Story = {
+  parameters: {
+    controls: { disable: true },
+  },
+  render: () => (
+    <Toast
+      autoCloseMs={0}
+      description="Here's your staking transaction:"
+      goTo={{
+        href: '/stake/dashboard',
+        label: 'Go to staking dashboard',
+      }}
+      title="Staking successful"
+      tx={{
+        href: 'https://explorer.hemi.xyz/tx/0x1234abcd5678ef90',
+        label: '0x1234…ef90',
+      }}
+    />
+  ),
+}
--- a/portal/stories/toast.stories.tsx
+++ b/portal/stories/toast.stories.tsx
@@ -1,0 +1,48 @@
+import type { Meta, StoryObj } from '@storybook/nextjs'
+import { Toast } from 'components/toast'
+
+const meta = {
+  args: {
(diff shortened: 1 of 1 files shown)


portal/stories/toast.stories.tsx
gabmontes, line 1: Nit: The `Toast` component also accepts a `goTo` prop (internal navigation link), which is actively used in production (e.g., `StakeToast` links to `/stake/dashboard`). Consider adding a `WithGoTo` story to cover that variant too — similar to how `WithTransaction` covers `tx`.

— Gabi 🤖

gabmontes (commented):
Clean and small PR — looks good overall. One minor suggestion about covering the `goTo` prop.

gndelia:
It seems that `variant=none` is showing something like `success` ?

Variant: None

image

Variant: Success

image

IT also seems that the "Error" icon is larger - given the same texts, it causes a different toast size


https://github.com/user-attachments/assets/a1ef802a-3e42-47d1-b765-e38043035535


It may be worth fixing these problems - maybe not for this PR, but we can open an issue

gndelia (dismissed):

portal/stories/toast.stories.tsx
arielmergen: Fixed in 4626bb03.

arielmergen:
Good catches, both confirmed. Since they're issues with the `Toast` component itself (not the story), I've split them out into #2062 to keep this PR scoped to just adding the story.

gndelia (approved):

ArturDolzan (approved):

gabmontes (approved):

dvnahuel:
Let's iterate the toast component to this [updated one](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=20261-30873&t=cRKBfQOjHKZkTSEe-1)

PR #2065, opened 2026-07-03
Handle unresolvable earnTxId in earn drawer

### LINKED ISSUE #1954: Hemi Earn: handle unresolvable txId in transactions drawer

Improve the empty-state handling in the earn transactions drawer when the `?earnTxId=...` URL param doesn't resolve to a valid transaction or pool.

Today the drawer renders an empty `<div class="drawer-content" />` (no heading, no message, no close button) whenever the txId can't be resolved. The user can only dismiss it via backdrop click — there's no signal of what went wrong.

Two scenarios trigger this:
1. A shared URL opens during the brief window before `useEarnTransactions` resolves (subgraph + local store still loading).
2. The deposit's asset was removed from the `HEMI_EARN_TOKENS` registry — `resolvePoolAndAsset` returns `undefined` permanently.

### Proposed implementation

Distinguish loading vs not-found using `useEarnTransactions`'s `isPending` flag:
- While `isPending`: render a skeleton/loading state inside the drawer.
- Once `!isPending` and the tx or pool can't be resolved: render a "Transaction not found" message with an explicit close button.

Avoid auto-clearing the URL during the loading window (`setTxDrawerQueryString(null)`) — that would cause a flicker when someone opens a shared link.

Code reference: `portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx` (lines 38–53, the two empty-drawer branches).

---

Original feedback: https://github.com/hemilabs/ui-monorepo/pull/1953#discussion_r3319517408
Follow-up from #1953.

ArturDolzan:
### Description

This PR fixes the Hemi Earn transactions drawer opening as an empty container when a shared `?earnTxId=` link can't be resolved — previously the user got no heading, no message and no close button, only a backdrop click to escape, and there was no way to tell "still loading" apart from "the deposit's asset was delisted so it'll never resolve".

Now the drawer differentiates the two using `isPending` from `useEarnTransactions` and `useEarnPools`. While it's still loading it shows a skeleton that mirrors the review layout and keeps the URL query string intact, so opening a shared link doesn't flicker. Once loading is done and the transaction/pool/asset still can't be resolved, it shows a "Transaction not found" empty state (same `TableEmptyState` + inbox icon used elsewhere in the feature) with an explicit Close button.

Along the way I exposed a small `useDrawerAnimatedClose` hook so the not-found Close button plays the drawer's exit animation instead of unmounting abruptly, and widened `DrawerTopSection`'s `heading` to `ReactNode` (backward-compatible — every caller still passes a string) so the skeleton can reuse it for a title placeholder.

### Screenshots

https://github.com/user-attachments/assets/827500ff-36d5-446a-8204-282790316c56

### Related issue(s)

Closes #1954 
Related to #1848 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/index.tsx
@@ -24,6 +24,8 @@ import { HistoricalWithdrawReview } from './historicalWithdrawReview'
 import { RetryFailedDeposit } from './retryFailedDeposit'
 import { RetryFailedWithdraw } from './retryFailedWithdraw'
 import { AddTokenToWalletCta, SettleCta } from './settleShared'
+import { TransactionDrawerSkeleton } from './transactionDrawerSkeleton'
+import { TransactionNotFound } from './transactionNotFound'
 import { useTxDrawerQueryString } from './useTxDrawerQueryString'
 
 // The drawer is already open, so the CTAs don't redirect on sign. A reverted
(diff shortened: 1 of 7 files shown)


gabmontes (dismissed):

gndelia:
Hashes have a specific length. We can use [isHash](https://viem.sh/docs/utilities/isHash#ishash) and if it's not a valid hash, I wouldn't even show the drawer. I'll just not show anything and unset the `earnTxId`. If it's a valid hash (in format), but we can't find it, then I would show what this PR implements.

I don't think it makes sense to show "Transaction not found" if `earnTxId=0x123` is obviously not a tx at all

Thoughts?

ArturDolzan:
> Hashes have a specific length. We can use [isHash](https://viem.sh/docs/utilities/isHash#ishash) and if it's not a valid hash, I wouldn't even show the drawer. I'll just not show anything and unset the `earnTxId`. If it's a valid hash (in format), but we can't find it, then I would show what this PR implements.
> 
> I don't think it makes sense to show "Transaction not found" if `earnTxId=0x123` is obviously not a tx at all
> 
> Thoughts?

Good call — makes sense to not treat garbage like 0x123 as a "not found" tx. Added the isHash guard in 9c058d8f: an invalid format now just clears earnTxId and shows nothing, and the "Transaction not found" state only kicks in for well-formed hashes we can't resolve.

gndelia (approved):

gabmontes (approved):

Issue #2028, opened 2026-06-23
Hemi Earn — redeem cooldown maturity finalize (Agent.claimUnstake) + unified state machine


ArturDolzan:
## Context

The redeem manual claim/recover PR mirrors the deposit **Router-side** settlement flow:

- `claimRedeem` on `FULFILLED` (delivers the underlying asset) → "Claim funds"
- `recoverRedeem` on `CANCELLED` (returns the shares) → "Recover shares"
- recover-path display + retry-on-revert

It deliberately does **not** cover the redeem **cooldown maturity** step, which is the dominant redeem path. Tracking that here.

## The gap

Per the UX spec (§3.3, §4, §5.5):

- The default redeem path has a ~7-day vault cooldown. During the entire wait the **Router stays `PENDING`** — there is no Router-side acknowledgment of cooldown (gotcha #1).
- At maturity the spec designs a primary CTA **"Claim from vault"** that calls **`Agent.claimUnstake(requestId)`** on the remote chain (anyone-callable) to make the Agent fulfill the redeem and bridge the asset back → Router `FULFILLED`.
- The `automatic` flag does **not** cover this. `automatic` only auto-finalizes the Router-side `claimRedeem`/`recoverRedeem` *after* the Router already reached `FULFILLED`/`CANCELLED`. So even `automatic = true` redeems sit at `PENDING` until someone calls `claimUnstake`.

The portal today only shows a cooldown countdown (`useEarnCooldownRemaining`) — it has **no** `claimUnstake` / Agent-state handling.

## Scope

1. **Unified state machine (spec §4):** compute the user-visible state from Router status **+** Agent reads — `OUTBOUND_FLIGHT` / `REMOTE_EXECUTING` / `COOLDOWN_WAIT` / `COOLDOWN_MATURE` / `RETURN_FLIGHT` — instead of the raw indexer status. Needs `Agent.unstakeRequests(id)` / `claimableAt` and the `UnstakeRequested` event (§6.2, §6.3).
2. **`COOLDOWN_MATURE` → "Claim from vault" CTA (spec §5.5):** wire `Agent.claimUnstake` as a UI fallback (remote-chain mutation; likely needs a return-leg `nativeFee` quote, cf. `Agent.quoteRedeemFulfillment`).
[...]

Issue #1943, opened 2026-05-26
hemi-earn-actions: add Agent-side write wrappers (cancel, retry, claimUnstake)


ArturDolzan:
Follow-up to #1942, which migrated `hemi-earn-actions` to the latest Router/Agent contract interface and added the four Router-side claim/recover wrappers. The Agent-side writes were intentionally scoped out and need their own PR before any UI feature that lets the user act on a stuck/cooldown request can ship.

## What's missing

`Agent` exposes three write functions that the portal will need to call from Ethereum, none of which have wrappers in `packages/hemi-earn-actions/` today:

- `Agent.cancel(uint256 requestId)` (Agent.sol:309) — operator-only. Branches internally on whether the request is in `failedRequests` (failed handler) or `unstakeRequests` (redeem cooldown). Sends original tokens back via LZ and the Router auto-recovers (because `automatic=true` is the hard-wired default).
- `Agent.retry(uint256 requestId)` (Agent.sol:292) — permissionless. Re-invokes `_receive` with the stored msg. Useful for gas-spike failures (less so for slippage, but the contract allows it).
- `Agent.claimUnstake(uint256 requestId)` (Agent.sol:262) — permissionless. Finalizes a matured Vetro cooldown and dispatches the LZ message back to Hemi. Normally the keeper handles it, but the UI needs the manual fallback button.

## Why we need it

With `automatic=true` and `operator = msg.sender` hard-wired in the UI (no advanced panel), the only user-facing error/recovery action collapses to a single "Cancel" or "Finalize manually" button. Both require a tx on Ethereum, and both need a typed wrapper with the same DX as `requestDeposit` / `claimDeposit` already have on the Router side.

## Scope of the PR

In `packages/hemi-earn-actions/`:

- `src/agentAbi.ts` — add ABI entries for `cancel`, `retry`, `claimUnstake`. Also add the read views the UI needs to compute state (`unstakeRequests(id)`, `failedRequests(id)`).
- `src/actions/wallet/` — add `cancelRequest.ts` (single wrapper since `Agent.cancel` branches internally on failed vs unstake), `retryRequest.ts`, `claimUnstake.ts`. Follow the `requestDeposit` pattern: event emitter via `toPromiseEvent`, validation (`requestId > 0n`), top-up support via `msg.value`, optional `payable` value.
- `src/actions/public/` — add `getUnstakeRequest` and `getFailedRequest` read wrappers around the new agent views, returning typed structs.
[...]

PR #2080, opened 2026-07-09
Add Agent retry/cancel/failedRequests wrappers

ArturDolzan:
### Description

This PR adds the Agent-side wrappers the portal needs to handle a REMOTE_FAILED request — the case where the Agent's remote execution reverts (slippage, insufficient native fee, gateway error), the request lands in Agent.failedRequests, and the Router stays stuck on PENDING with no action available in the UI today.

The key piece is getFailedRequest, a read of failedRequests(id): it's the authoritative "is this genuinely stuck or already being auto-recovered?" signal, because the contract deletes the entry the moment a retry or cancel succeeds (so tokenIn != 0x0 means still stuck). On top of that it adds the two payable writes — retryRequest (Agent.retry, re-runs the failed leg with a gas top-up) and cancelRequest (Agent.cancel, returns the original tokens). Note cancelRequest calls Agent.cancel on Ethereum and is deliberately distinct from the existing cancelRedeem, which is Router.cancel on Hemi.

Everything mirrors the existing claimUnstake / getUnstakeRequest siblings — new agentAbi entries (cancel, retry, failedRequests), event types, barrels and tests. It's package-only and purely additive; nothing consumes it yet, the portal wiring comes in a follow-up PR.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #1943
Related to #2028 
Related to #1848 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/packages/hemi-earn-actions/src/actions/wallet/cancelRequest.ts
+++ b/packages/hemi-earn-actions/src/actions/wallet/cancelRequest.ts
@@ -1,0 +1,92 @@
+import { EventEmitter } from 'events'
+import { toPromiseEvent } from 'to-promise-event'
+import type { Address, TransactionReceipt, WalletClient } from 'viem'
+import {
+  getBalance,
+  waitForTransactionReceipt,
+  writeContract,
+} from 'viem/actions'
+
+import { agentAbi } from '../../agentAbi'
+import type { CancelRequestEvents } from '../../types'
+
+const runCancelRequest = ({
+  account,
+  agentAddress,
+  nativeFee = BigInt(0),
+  requestId,
+  walletClient,
+}: {
+  account: Address
+  agentAddress: Address
+  nativeFee?: bigint
+  requestId: bigint
+  walletClient: WalletClient
+}) =>
+  async function (emitter: EventEmitter<CancelRequestEvents>) {
+    try {
+      if (!walletClient.chain) {
+        throw new Error('Chain is not defined on wallet')
+      }
+
+      if (requestId <= BigInt(0)) {
+        emitter.emit('tx-failed-validation', 'invalid requestId')
+        return
+      }
+
+      if (nativeFee > BigInt(0)) {
+        const balance = await getBalance(walletClient, { address: account })
+        if (balance < nativeFee) {
+          emitter.emit('tx-failed-validation', 'insufficient balance for fee')
+          return
+        }
+      }
+
+      emitter.emit('pre-tx')
+
+      const txHash = await writeContract(walletClient, {
+        abi: agentAbi,
+        account,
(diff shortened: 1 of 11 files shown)


packages/hemi-earn-actions/src/types.ts
```diff
@@ -65,6 +65,9 @@ export type RecoverRedeemEvents = CommonEvents & SettlementEvents
 // Router.cancel only emits CancellationRequested; the keeper's Agent.cancel then
 // sets CANCELLED and the user calls recoverRedeem to pull shares back.
 export type CancelRedeemEvents = CommonEvents & SettlementEvents
+// Agent-side writes for a failed request on Ethereum (cancel ≠ Router-side CancelRedeem).
+export type CancelRequestEvents = CommonEvents & SettlementEvents
+export type RetryRequestEvents = CommonEvents & SettlementEvents
```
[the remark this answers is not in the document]
ArturDolzan: Good catch — you're right, only `retry` is failed-only; `cancel` also covers an in-cooldown request. Tightened the comment in [`83f52d22`](https://github.com/hemilabs/ui-monorepo/pull/2080/commits/83f52d22).

packages/hemi-earn-actions/src/actions/wallet/cancelRequest.ts
```diff
@@ -0,0 +1,92 @@
...
+  writeContract,
+} from 'viem/actions'
+
+import { agentAbi } from '../../agentAbi'
+import type { CancelRequestEvents } from '../../types'
+
+const runCancelRequest = ({
+  account,
+  agentAddress,
+  nativeFee = BigInt(0),
```
gndelia, line 16: is this  default value correct? Wouldn't a nativeFee of 0 revert?

gndelia (approved):
I left a question, but LGTM

may also want to consider checking native token balance in the package, but not a blocker

arielmergen (approved):

packages/hemi-earn-actions/src/actions/wallet/cancelRequest.ts
ArturDolzan: Good question! The `0` is just "no extra top-up" and mirrors `claimUnstake` (same default). On-chain `cancel` sends with `failedRequest.nativeFee + msg.value`, so it rides on the fee already pre-funded with the original request, and the caller passes a top-up when that's short (the UI hook quotes before calling) — keeping the wrapper fee-agnostic like the other actions. You're right that a bare `0` on an under-funded request would revert though; whether the caller tops up *enough* for the cancel path specifically is worth a closer look, so happy to open an issue to dig into that fee quote 🙂

PR #2083, opened 2026-07-10
Add remote-failure recovery to Hemi Earn redeems

ArturDolzan:
### Description

This PR implements a recovery path for Hemi Earn redeems that get stuck on Ethereum. When the Agent's cross-chain execution reverts — slippage, an insufficient LayerZero fee, or a gateway failure — the request lands in `Agent.failedRequests` while the Router stays `PENDING`, and until now the row just showed a spinner with no way forward.

Important context: these are **fallbacks for edge cases the keeper bot couldn't auto-resolve on its own**. In normal operation the keeper retries/cancels these automatically, so the user should never see or use these controls — they only surface after a short grace window (giving the bot first crack) and only while the request is still stuck on-chain.

When it does surface, the user gets a **Retry** (top up the fee and re-run the redeem) and/or **Cancel** (return the share tokens), which call `Agent.retry` / `Agent.cancel` on Ethereum, mirroring the existing Claim-from-vault flow. The failure category decides what's offered: slippage shows only Cancel (a retry would just slip again against the frozen min-out), while gas/unknown lead with Retry and keep Cancel as an escape hatch. It's redeem-only for now; deposit recovery is a follow-up.

Under the hood it consumes the `receivedAt` / `retryCount` / `failureReason` fields from the earn-requests API, and wires the CTA, banner, and a grace-aware status badge + cross-chain step into both the live and historical withdraw drawers.

### Screenshots

slippage
Captura de Tela 2026-07-10 às 11 45 18

fees/unknown

Captura de Tela 2026-07-10 às 10 31 29

Table

Captura de Tela 2026-07-10 às 13 31 59

### Related issue(s)

Related to #1943
Related to #2028 
Related to #1848 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts
+++ b/portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts
@@ -1,3 +1,5 @@
+'use client'
+
 import { useEnsureConnectedTo } from '@hemilabs/react-hooks/useEnsureConnectedTo'
 import { useNativeBalance } from '@hemilabs/react-hooks/useNativeBalance'
 import { useUpdateNativeBalanceAfterReceipt } from '@hemilabs/react-hooks/useUpdateNativeBalanceAfterReceipt'
@@ -10,6 +12,7 @@ import {
 import {
   cancelRequest,
   getFailedRequest,
+  quoteDepositFulfillment,
   quoteRedeemFulfillment,
   retryRequest,
 } from 'hemi-earn-actions/actions'
@@ -83,14 +86,23 @@ const useRemoteFailedAction = function ({
         })
       }
 
-      const { remoteAsset } = await queryClient.ensureQueryData(
+      const { remoteAsset, remoteShare } = await queryClient.ensureQueryData(
         assetDataQueryOptions(asset),
       )
-      const quote = await quoteRedeemFulfillment({
-        agentAddress,
-        asset: remoteAsset,
-        client,
-      })
+      // Retry re-sends the asset fulfillment (asset OFT); cancel returns the shares (share OFT).
+      // Quote each on the OFT it actually uses so the top-up isn't sized against the wrong path.
+      const quote =
+        kind === 'RETRY'
+          ? await quoteRedeemFulfillment({
+              agentAddress,
+              asset: remoteAsset,
+              client,
+            })
+          : await quoteDepositFulfillment({
+              agentAddress,
+              client,
+              share: remoteShare,
+            })
       // retry/cancel run with failedRequest.nativeFee + msg.value on-chain, so only top up the shortfall.
       const nativeFee = maxBigInt(BigInt(0), quote - failedRequest.nativeFee)
 
--- a/portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts
+++ b/portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts
@@ -1,0 +1,173 @@
+import { useEnsureConnectedTo } from '@hemilabs/react-hooks/useEnsureConnectedTo'
+import { useNativeBalance } from '@hemilabs/react-hooks/useNativeBalance'
+import { useUpdateNativeBalanceAfterReceipt } from '@hemilabs/react-hooks/useUpdateNativeBalanceAfterReceipt'
+import { useMutation, useQueryClient } from '@tanstack/react-query'
+import { type EventEmitter } from 'events'
+import {
+  type CancelRequestEvents,
+  type RetryRequestEvents,
+} from 'hemi-earn-actions'
+import {
+  cancelRequest,
+  getFailedRequest,
+  quoteRedeemFulfillment,
+  retryRequest,
+} from 'hemi-earn-actions/actions'
+import { mainnet } from 'networks/mainnet'
+import { maxBigInt } from 'utils/bigint'
+import { getEvmL1PublicClient } from 'utils/chainClients'
+import { isAddressEqual, zeroAddress } from 'viem'
+import { useAccount, useWalletClient } from 'wagmi'
+
+import { earnTransactionsKeyPrefix } from '../_fetchers/fetchEarnTransactions'
+import { type EarnSettlement, type EarnTransaction } from '../types'
+
+import { assetDataQueryOptions } from './useAssetData'
+import { getFailedRequestQueryKey } from './useFailedRequest'
+import { agentAddressQueryOptions } from './useHemiEarnAgentAddress'
+import { useLocalEarnOperations } from './useLocalEarnOperations'
+
+type RemoteFailedKind = Extract<
+  EarnSettlement['kind'],
+  'RETRY' | 'CANCEL_REQUEST'
+>
+
+type UseRemoteFailedAction = {
+  action: typeof retryRequest
+  kind: RemoteFailedKind
+  on?: (emitter: EventEmitter<RetryRequestEvents>) => void
+  transaction: EarnTransaction
+}
+
+const useRemoteFailedAction = function ({
+  action,
+  kind,
+  on,
+  transaction,
+}: UseRemoteFailedAction) {
+  const { address } = useAccount()
+  const ensureConnectedTo = useEnsureConnectedTo()
+  const queryClient = useQueryClient()
+  const { setSettlement } = useLocalEarnOperations()
+  const { data: l1WalletClient } = useWalletClient({
+    chainId: mainnet.id,
+  })
+  const updateNativeBalanceAfterFees = useUpdateNativeBalanceAfterReceipt(
+    mainnet.id,
+  )
+  const { queryKey: nativeTokenBalanceQueryKey } = useNativeBalance(mainnet.id)
+
+  const { asset, requestId, requestTxHash } = transaction
+
+  return useMutation({
+    async mutationFn() {
+      if (!address) {
+        throw new Error('No account connected')
+      }
+
+      const agentAddress = await queryClient.ensureQueryData(
+        agentAddressQueryOptions(),
+      )
+      const client = getEvmL1PublicClient(mainnet.id)
+      const id = BigInt(requestId)
+
+      const failedRequest = await getFailedRequest({
+        agentAddress,
+        client,
+        requestId: id,
+      })
+      // Entry already gone — a keeper retried/cancelled first; nothing to sign, just refresh.
+      if (isAddressEqual(failedRequest.tokenIn, zeroAddress)) {
+        return queryClient.invalidateQueries({
+          queryKey: earnTransactionsKeyPrefix,
+        })
+      }
+
+      const { remoteAsset } = await queryClient.ensureQueryData(
+        assetDataQueryOptions(asset),
+      )
+      const quote = await quoteRedeemFulfillment({
+        agentAddress,
+        asset: remoteAsset,
+        client,
+      })
+      // retry/cancel run with failedRequest.nativeFee + msg.value on-chain, so only top up the shortfall.
+      const nativeFee = maxBigInt(BigInt(0), quote - failedRequest.nativeFee)
+
+      await ensureConnectedTo(mainnet.id)
+
+      const { emitter, promise } = action({
+        account: address,
+        agentAddress,
+        nativeFee,
+        requestId: id,
+        walletClient: l1WalletClient!,
+      })
+
+      const fail = () => setSettlement(requestTxHash, { failed: true, kind })
+
+      emitter.on('user-signed-tx', function (txHash) {
+        setSettlement(requestTxHash, { failed: false, kind, txHash })
+      })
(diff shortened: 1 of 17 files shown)


portal/app/[locale]/hemi-earn/_utils/decodeFailureReason.ts
```diff
@@ -0,0 +1,43 @@
...
+const agentErrorsAbi = [
+  {
+    inputs: [
+      { name: 'provided', type: 'uint256' },
+      { name: 'required', type: 'uint256' },
+    ],
+    name: 'InsufficientFee',
+    type: 'error',
+  },
+] as const
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/decodeFailureReason.ts:20]: Thanks! viem's `decodeErrorResult` prepends the standard `Error(string)` and `Panic(uint256)` selectors itself, so they decode even though `agentErrorsAbi` only lists `InsufficientFee` — the `errorName === 'Error'` branch does fire. The decode tests cover exactly this (slippage `Error(string)` → `'slippage'`, `Panic` → `'unknown'`) and pass. Happy to add explicit `Error`/`Panic` entries to the ABI if you'd rather have it spelled out, but it's not needed for correctness 🙂

portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts
```diff
@@ -0,0 +1,173 @@
+import { useEnsureConnectedTo } from '@hemilabs/react-hooks/useEnsureConnectedTo'
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts:3]: nit: The sibling hooks added in this PR (`useFailedRequest.ts`, `useRemoteFailedState.ts`) include the `'use client'` directive — this one is missing it. Not a runtime bug (transitively client-bundled via `remoteFailed.tsx`), but adding it keeps the pattern consistent and gives a clearer error if someone accidentally imports it from a server component later.

— Gabi 🤖

gabmontes (commented):
Clean PR — well-structured, defensive, and thoroughly tested. One minor nit.

gndelia:
Hi @ArturDolzan 👋 — did a pass over this. Nicely decomposed and the pure-logic test coverage is great. A few things to check, none blocking:

**1. Cancel fee sized with the redeem-fulfillment quote (worth verifying)**
In `useRemoteFailedActions.ts`, `quoteRedeemFulfillment` sizes the top-up for **both** retry and cancel:
```ts
const quote = await quoteRedeemFulfillment({ agentAddress, asset: remoteAsset, client })
const nativeFee = maxBigInt(BigInt(0), quote - failedRequest.nativeFee)
```
`Agent.cancel` sends a different cross-chain message (return shares) than a redeem fulfillment. If cancel's required LZ fee differs from `quoteRedeemFulfillment`, the top-up could under/over-fund it. Under-funding self-heals (tx reverts → `fail()` → user retries), so not a hard bug — but can you confirm the Agent forwards the stored `nativeFee` on cancel and that the redeem quote is the right ceiling? If cancel has its own quote fn, use that.

**2. Disconnected users never see "Action needed" (UX)**
`useFailedRequest` is `enabled: !!address && isRemoteFailed(...)`. Disconnected → `failedRequest` undefined → `isStuck` false → `show` false, so `StatusBadge` in the table falls back to the in-progress spinner instead of the amber "Action needed". Acting needs a connection anyway, but the badge is always visible in the table — worth deciding whether that amber state should be reachable pre-connection.

**3. Custom-error slippage decodes to `unknown` → leads with Retry (acknowledged)**
A slippage revert that arrives as a custom Vetro error (not `Error(string)`) maps to `unknown`, which leads with Retry — and a retry against the frozen min-out just slips again. You already call this out in the description and Cancel stays available, so it degrades fine. Flagging only so it's a conscious trade-off; if the Vetro slippage error selectors are known, adding them to `agentErrorsAbi` would tighten it.

**4. Nit:** `BigInt(requestId!)` in `useFailedRequest.queryFn` leans on `enabled` for the non-null — a small guard/narrowing would drop the assertion.

Things I explicitly checked and are fine: CTA precedence (`showRemoteFailed` vs manual/claim, and `FAILED_STATUSES` vs `isRemoteFailed` — `withdrawStatus` is the local op status so no shadowing), the concurrent-keeper race handling in the mutation, the fee shortfall math, and the `on` callback shared across retry/cancel (`RetryRequestEvents`/`CancelRequestEvents` are identical aliases). 👍

portal/app/[locale]/hemi-earn/_utils/decodeFailureReason.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  if (decoded.errorName === 'InsufficientFee') return 'gas'
+  if (decoded.errorName === 'Error') {
+    return slippageMessage.test(String(decoded.args?.[0] ?? ''))
+      ? 'slippage'
+      : 'unknown'
+  }
+  return 'unknown'
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_utils/decodeFailureReason.ts:43]: I wonder if this should go into the Agent package (if there's such a package?)

portal/app/[locale]/hemi-earn/types.ts
```diff
@@ -46,13 +47,19 @@ export type EarnTransaction = {
   claimTxHash: Hash | null
   // Agent-side failure flag; lingers through CANCELLED/RECOVERED (failed:false recover = deliberate cancel).
   failed: boolean
+  // Raw Agent revert reason (bytes) on a remote failure; decoded to pick the recovery CTA.
+  failureReason?: string | null
   kind: EarnTransactionKindType
   processedAt?: string | null
+  // Unix seconds the Agent received the request (== the failure block on a remote failure).
+  receivedAt?: string | null
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/types.ts:55]: Just to confirm, this means `receivedAt` may be a string, `null` or `undefined`. Can we double check indeed it can return `undefined` and `null` ? Why the 2 values? The same applies  to the other props

portal/messages/es.json
```diff
@@ -380,6 +380,11 @@
           "heading": "La transacción falló",
           "subheading": "Algo salió mal. Use el botón de abajo para devolver sus share tokens a su billetera."
         },
+        "remote-failed": {
+          "heading": "Su retiro quedó atascado en Ethereum",
+          "subheading": "El paso entre cadenas falló. Reintente para agregar gas y completar el retiro, o cancele para devolver sus share tokens a su billetera.",
```
gndelia: `cadenas` is the literal translation of chains, but in the context of crypto, it should be `red` (singular) and `redes` (plural)

```suggestion
          "subheading": "El paso entre redes falló. Reintente para agregar gas y completar el retiro, o cancele para devolver sus share tokens a su billetera.",
```

gndelia (changes requested):

portal/messages/pt.json
```diff
@@ -380,6 +380,11 @@
           "heading": "Transação falhou",
           "subheading": "Algo deu errado. Use o botão abaixo para devolver seus share tokens à sua carteira."
         },
+        "remote-failed": {
+          "heading": "Seu saque travou na Ethereum",
+          "subheading": "A etapa entre cadeias falhou. Tente novamente para adicionar gas e concluir o saque, ou cancele para devolver seus share tokens à sua carteira.",
```
gndelia: I don't know portuguese but `cadeias` seems to have the same issue as  [here](https://github.com/hemilabs/ui-monorepo/pull/2083/changes#r3571833102)

dvnahuel:
As a minor point of improvement I would suggest iterating some of the texts to the following:

**Drawer**

1. Info card: Withdrawal didn't complete - The price moved past your slippage limit before the final step could run. Your svetBTC is still held by the vault. Return it to your wallet and try the withdrawal again with a higher slippage tolerance.

**CTA**

1. Return svetBTC to wallet

**Table status and CTA**

1. Status: Withdrawal Slippage limit exceeded.
2. CTA: Return share tokens.

cc: @ArturDolzan

portal/app/[locale]/hemi-earn/_utils/decodeFailureReason.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/decodeFailureReason.ts:43]: Makes sense. Want me to open an issue to look at splitting it out?

portal/app/[locale]/hemi-earn/types.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/types.ts:55]: Both are reachable and intentional: `null` is the subgraph value before it's set, `undefined` is the local-store optimistic rows that don't carry these fields until indexed. Consumers collapse both with `?? 0` / `?? null`. Same for the sibling props.

portal/app/[locale]/hemi-earn/_utils/decodeFailureReason.ts
Comment (gndelia) [reply portal/app/[locale]/hemi-earn/_utils/decodeFailureReason.ts:43]: ok

arielmergen:
I'm working on the button sizes and the badge component. Since both are used in this table, once my PR is merged, the buttons and badges here will be fixed as well.

Thoughts?

Here the pr Button size https://github.com/hemilabs/ui-monorepo/pull/2091

cc: @ArturDolzan @gndelia @dvnahuel

portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts:3]: Good catch — added it in [`b34631b1`](https://github.com/hemilabs/ui-monorepo/pull/2083/commits/b34631b1).

portal/messages/es.json
ArturDolzan: Good catch, `red`/`redes` it is. Fixed in [`b34631b1`](https://github.com/hemilabs/ui-monorepo/pull/2083/commits/b34631b1).

portal/messages/pt.json
ArturDolzan: Same fix on `pt` — `cadeias` → `redes` in [`b34631b1`](https://github.com/hemilabs/ui-monorepo/pull/2083/commits/b34631b1).

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2083#issuecomment-4959361028):

Thanks for the thorough pass — going point by point:

**1. Cancel fee sized with the redeem quote** — good catch. Cancel returns the shares through the share OFT while only retry re-sends the asset, so I split the quote by path: retry keeps `quoteRedeemFulfillment(asset)`, cancel now uses `quoteDepositFulfillment(remoteShare)` (same 33-byte msg, just the right OFT). Done in [`b34631b1`](https://github.com/hemilabs/ui-monorepo/pull/2083/commits/b34631b1).

**2. Disconnected users never see "Action needed"** — deliberate: the amber keys off the on-chain `failedRequests` read, which needs a connection, so we don't flash a stale "Action needed" after a keeper already resolved it. Acting needs a wallet anyway. Happy to open an issue if you'd like to explore a subgraph-driven pre-connection badge.

**3. Custom-error slippage → unknown** — conscious trade-off; Cancel stays available so it degrades fine. The revert comes from the Vetro gateway and we only have its interface, so the selectors aren't known here — if they surface we can add them to `agentErrorsAbi`.

**4. `BigInt(requestId!)` nit** — done, swapped the assertion for a guard in [`b34631b1`](https://github.com/hemilabs/ui-monorepo/pull/2083/commits/b34631b1).

ArturDolzan:
> [@dvnahuel commented](https://github.com/hemilabs/ui-monorepo/pull/2083#issuecomment-4959674841):

Thanks for the copy pass — applied all four in [`b34631b1`](https://github.com/hemilabs/ui-monorepo/pull/2083/commits/b34631b1).

portal/messages/es.json
gndelia: @ArturDolzan this still reads as `cadenas`

ArturDolzan: Ah — it was in `b34631b1` but hadn't been pushed yet, so the PR was still on the old file. It's up there now, should read `redes`. Sorry for the noise!

portal/messages/en.json
```diff
@@ -392,6 +398,7 @@
         "keep": "Keep withdrawal",
         "title": "Cancel withdrawal"
       },
+      "cancel-request": "Cancel",
```
gabmontes: Not sure we should use just the "Cancel" text here. It can be read as in an "OK/Cancel" case, not meaning to issue a new "cancel" operation but to dismiss any action. cc/ @dvnahuel

dvnahuel: Indeed! I left my suggestion on the copy [here](https://github.com/hemilabs/ui-monorepo/pull/2083#issuecomment-4959674841)

ArturDolzan: Good point — dropped the ambiguous "Cancel", now every remote-failed cancel reads "Return share tokens" (it always returns the shares), so gas/unknown shows "Retry" + "Return share tokens". Done in [`e69ac165`](https://github.com/hemilabs/ui-monorepo/pull/2083/commits/e69ac165).

gndelia (approved):

gabmontes (approved):

Issue #2092, opened 2026-07-14
Hemi Earn: add local sandbox scripts for end-to-end testing


ArturDolzan:
### Feature or change description

Add a small set of TypeScript scripts under `portal/scripts/hemi-earn/` that spin up and configure a local sandbox for the Hemi Earn UI end-to-end flow (deposit, withdraw, cooldown, recover, retry).

The scripts should mirror the pattern already used in other Hemi frontends: plain `.ts` modules invoked with `tsx`, each one exportable as a function and runnable as a CLI (`parseArgs` from `node:util`, dual-mode via `import.meta.url`).

### Problem or use case addressed

End-to-end testing of the Hemi Earn UI currently requires setup that lives outside this repository. As a result, only a subset of contributors can reproduce the flow locally, and the tooling isn't versioned alongside the code it exercises.

Bringing the minimum scripts in-tree makes the sandbox reproducible for any contributor without extra prerequisites, and keeps the tooling evolving in sync with the UI.

### Proposed implementation

Deliver incrementally across several small PRs, starting with the bare minimum to bring the sandbox up and configured:

1. **Baseline** — deploy the required mocks, fund a test account, control the cooldown state.
2. **Follow-ups** — additional scripts for time manipulation, state control, failure simulation, etc.

Custom mock contracts required by the sandbox will be consumed as pre-compiled JSON artifacts (bytecode + ABI) rather than sources — the monorepo does not need a Solidity toolchain.

**Out of scope**

- Any temporary source-level patches applied per testing session (constants, hooks, subgraph config, env flags). Those remain a manual step for now.
- Publishing or vendoring the private contract sources. Only pre-compiled artifacts land in this repo.

ArturDolzan:
This is done.

PR #2093, opened 2026-07-14
Add local sandbox scripts for Hemi Earn

ArturDolzan:
### Description

This PR adds a local sandbox for running the Hemi Earn UI end-to-end without needing an external repo. The setup lives under `portal/scripts/hemi-earn/` as a self-contained workspace: it forks Hemi mainnet via Anvil, deploys the required mocks, aliases the Vetro contracts at their production addresses via `anvil_setCode`, funds a test account, and enables cooldown by default.

Custom mock bytecode ships as pre-compiled JSON artifacts to avoid pulling a Solidity toolchain into the monorepo. The nested `package.json` scopes `"type": "module"` + Node types to this folder alone, so Next.js's CJS resolution in `portal/` stays untouched.

Captura de Tela 2026-07-14 às 14 17 08

Setup:

Captura de Tela 2026-07-14 às 13 24 49

### Related issue(s)

Related to #2092 

### Checklist

- [x] Manual testing passed.


portal/scripts/hemi-earn/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "hemi-earn-sandbox-scripts",
+  "version": "0.0.0",
+  "scripts": {
+    "setup": "node --experimental-transform-types setup.ts"
+  },
+  "dependencies": {
+    "@vetro-protocol/earn": "1.0.0",
+    "@vetro-protocol/gateway": "1.2.0"
+  },
```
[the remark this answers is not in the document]
ArturDolzan: Good catch — done in `5c20417c`. Pinned `viem@2.28.4` in the workspace so the Vetro peers resolve against the same version portal uses, and dropped the now-redundant `node/no-extraneous-import` override 🙂

portal/scripts/hemi-earn/README.md
```diff
@@ -0,0 +1,65 @@
...
+- A JSON-RPC URL for Hemi mainnet (e.g. `https://rpc.hemi.network/rpc`).
+- A test EOA — used to receive funded ETH and tokens.
+
+## Quick start
+
+Run both commands from the repo root:
+
+```bash
+# Terminal 1 — fork Hemi mainnet
+anvil --fork-url https://rpc.hemi.network/rpc
```
gndelia: Note we have a package for this in [@hemilabs/anvil-fork-setup](https://www.npmjs.com/package/@hemilabs/anvil-fork-setup) that can be called from JS as part of your startup setup

portal/scripts/hemi-earn/artifacts.ts
```diff
@@ -0,0 +1,22 @@
...
+export function loadArtifact(name: string) {
+  const path = join(artifactsDir, `${name}.json`)
+  const raw = readFileSync(path, 'utf8')
+  try {
+    return JSON.parse(raw) as Artifact
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err)
+    throw new Error(`Failed to parse artifact ${path}: ${message}`)
+  }
+}
```
gndelia, line 22: Maybe I'm not following - but these artifacts seem to be deploying everything on your local fork - is that ok? 

The whole structure of hemi-earn already exists on mainnet (and hemi), so maybe just forking from the real blockchain should work 🤔 

`viem` has [impersonateAccount](https://v1.viem.sh/docs/actions/test/impersonateAccount.html#impersonateaccount) as part of the testClient, which allows you to become owner of whatever you need. Have you explored that approach ? E2E tests in Vetro do that, for example

portal/scripts/hemi-earn/README.md
ArturDolzan: Great pointer — done in [`7db295de`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/7db295de). `setup.ts` now calls `startAnvilFork` (auto-installs Foundry if missing, detaches so anvil survives the script exit), which collapses the two-terminal flow into a single command. Also swapped our hardcoded `DEFAULT_DEPLOYER_PK` for the `TEST_PRIVATE_KEY` re-export from `@hemilabs/anvil-fork-setup/utils` — cleaner single source of truth 🙂

portal/scripts/hemi-earn/artifacts.ts
ArturDolzan: Good question — I did consider that! The catch here is that Hemi Earn spans two chains: the Router lives on Hemi (43111), and the Agent/Gateway/Staking live on Ethereum (1). The sandbox collapses both onto a single anvil to avoid running two forks + wiring up a LayerZero relayer between them, so one of the sides has to be "brought over" via `anvil_setCode` — that's where the artifacts come in. A fork+impersonate approach would still need to source the Ethereum-side bytecode somehow (probably `eth_getCode` against a real Ethereum RPC at runtime), which trades committed artifacts for a live-RPC dependency. Worth exploring but a bigger change than fits this PR — happy to open a follow-up issue to dig into that trade-off calmly 🙂

gabmontes (commented):
Lockfile drift?

ArturDolzan:
> [@gabmontes commented](https://github.com/hemilabs/ui-monorepo/pull/2093#pullrequestreview-4697957683):
> Lockfile drift?

Good catch — reverted in [`b0e69af1`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/b0e69af1). pnpm re-resolved `ox@0.6.9`'s crypto subtree when the workspace was added; scoped the diff back to just our additions.

portal/README.md
```diff
@@ -84,6 +84,10 @@ Run tests with the following command:
 pnpm test
 ```
 
+## Local sandbox for Hemi Earn
+
+To test the Hemi Earn UI end-to-end against a controlled Anvil fork, see [`scripts/hemi-earn/README.md`](./scripts/hemi-earn/README.md). The sandbox deploys the required mocks and funds a test account so the UI can be exercised without external tooling.
+
```
gabmontes, line 90: Where are the contracts in the `/artifacts` folder come from? How can those be changed or updated?

portal/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "lib": ["dom", "dom.iterable", "esnext", "WebWorker"],
     "target": "es6"
   },
+  "exclude": ["scripts/hemi-earn/**/*"],
```
gabmontes, line 10: Why?

pnpm-workspace.yaml
```diff
@@ -1,6 +1,7 @@
 packages:
   - 'packages/*'
   - 'portal'
+  - 'portal/scripts/hemi-earn'
```
gabmontes: Should these be a different workspace? If are just scripts used to test the portal, I don't think there is a need to declare a workspace.

portal/README.md
ArturDolzan: Good question — added the sources under [`contracts/`](https://github.com/hemilabs/ui-monorepo/tree/HEAD/portal/scripts/hemi-earn/contracts) in [`23b258d0`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/23b258d0), plus a `contracts/README.md` covering the refresh workflow. Both the `.sol` and `.json` land together in the same PR to avoid drift (it was tagged to land in a upcoming PR).

portal/tsconfig.json
ArturDolzan: `portal/tsconfig.json` targets `es6` (via `@tsconfig/next`), but the sandbox scripts have their own `tsconfig.json` targeting `ES2023` + Node types so they can use `BigInt` literals (`10_000n`) and Node built-ins. Without this exclude, `next build` type-checks the scripts under the wrong config and fails on the bigint literals.

pnpm-workspace.yaml
ArturDolzan: Went ahead and applied this in [`23b258d0`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/23b258d0) — dropped the workspace declaration, moved `@hemilabs/anvil-fork-setup` into portal's devDeps, and lifted `sandbox:hemi-earn:setup` back into portal's scripts so knip still sees the entry. Nested `package.json` stays as `{ "type": "module" }` only, since portal can't set that without breaking Next's CJS resolution. Command is now `pnpm --filter portal sandbox:hemi-earn:setup -- --address 0xEOA`.

gabmontes (dismissed):

portal/scripts/hemi-earn/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+if (!btcGateway) {
+  throw new Error(
+    "No gateway with pegBaseSymbol='BTC' in @vetro-protocol/gateway",
+  )
+}
+export const GATEWAY_PROD: Address = btcGateway.address
+
+export const STAKING_PROD: Address = sVetBtcAddress
+
+export const HEMI_CHAIN_ID = 43111
```
gndelia: not needed - whoever needs this, can just import `hemi` from `viem`

portal/scripts/hemi-earn/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const DEFAULT_FORK_URL = 'http://127.0.0.1:8545'
+export const DEFAULT_UPSTREAM_RPC = 'https://rpc.hemi.network/rpc'
```
gndelia: Same here - can be read from `hemi` in `viem`

portal/scripts/hemi-earn/README.md
```diff
@@ -0,0 +1,77 @@
+# Hemi Earn — Local sandbox
+
+Scripts to spin up and configure a local Hemi Earn sandbox against an Anvil fork of Hemi mainnet. Lets any contributor run the Hemi Earn UI end-to-end without external tooling.
+
+## Prerequisites
+
+- Node 24 (matches the repo `.nvmrc`) — required for native `.ts` execution via `--experimental-transform-types`.
```
gndelia: This flag is not needed.

```suggestion
- Node 24 (matches the repo `.nvmrc`) — required for native `.ts`.
```

As part of node 24, this is already stable. See that `portal-backend` runs in TS without using this flag

portal/scripts/hemi-earn/README.md
```diff
@@ -0,0 +1,77 @@
...
+  --port 8547 \
+  --upstream-rpc https://your-hemi-rpc.example.com
+```
+
+## Available scripts
+
+Each script is a standalone `.ts` module that exports its main function and can also be run as a CLI. Run them from anywhere via:
+
+```bash
+node --experimental-transform-types portal/scripts/hemi-earn/<script>.ts [flags]
```
gndelia:
```suggestion
node portal/scripts/hemi-earn/<script>.ts [flags]
```

portal/package.json
```diff
@@ -8,6 +8,7 @@
     "dev": "next dev",
     "dev:wifi": "WIFI=true LOCAL_IP=$(ipconfig getifaddr en0) PORT=3000 pnpm run dev",
     "generate:htaccess": "node ./scripts/generateServerConfig.js",
+    "sandbox:hemi-earn:setup": "node --experimental-transform-types scripts/hemi-earn/setup.ts",
```
gndelia: Same here - should be able to remove it

```suggestion
    "sandbox:hemi-earn:setup": "node scripts/hemi-earn/setup.ts",
```

portal/scripts/hemi-earn/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "noEmit": true,
+    "resolveJsonModule": true,
+    "strict": true,
+    "target": "ES2023",
+    "types": ["node"]
+  },
+  "include": ["**/*.ts"]
+}
```
gndelia, line 12: We're usually using `@tsconfig/nodeXX` configs to use standard configs across packages, and not maintaining custom ones

See [api tsconfig.json](https://github.com/hemilabs/ui-monorepo/blob/main/portal-backend/api/tsconfig.json) for example. I suggest doing the same here. `"extends": "@tsconfig/node24/tsconfig.json",` should work for you

portal/scripts/hemi-earn/constants.ts
ArturDolzan: Good call — dropped in [`7df451c0`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/7df451c0), now pulls `hemi.id` from `viem/chains`.

portal/scripts/hemi-earn/constants.ts
ArturDolzan: Same — done in [`7df451c0`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/7df451c0), reads `hemi.rpcUrls.default.http[0]` now.

portal/scripts/hemi-earn/README.md
ArturDolzan: Right, dropped the flag in [`7df451c0`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/7df451c0).

portal/scripts/hemi-earn/README.md
ArturDolzan: Done in [`7df451c0`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/7df451c0).

portal/package.json
ArturDolzan: Done in [`7df451c0`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/7df451c0).

portal/scripts/hemi-earn/tsconfig.json
ArturDolzan: Good pointer — done in [`7df451c0`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/7df451c0). Also added a `typescript.config` entry in `.knip.json` so knip follows the extends chain from the nested tsconfig and does not flag the dep as unused.

gndelia (dismissed):

gabmontes (commented):
Minor nit — one stale string leftover from the workspace→portal migration.

portal/scripts/hemi-earn/setup.ts
```diff
@@ -0,0 +1,105 @@
...
+      'upstream-rpc': { short: 'u', type: 'string' },
+    },
+    strict: true,
+  })
+
+  if (!values.address || !isAddress(values.address, { strict: false })) {
+    console.error(
+      'Usage: pnpm --filter hemi-earn-sandbox-scripts run setup -- --address 0xYourEOA',
+    )
+    console.error('  [-p PORT]              port for the fork (default 8545)')
```
gabmontes, line 30: Stale usage string — the `hemi-earn-sandbox-scripts` workspace was removed in `23b258d0`. This should match what the README documents:

```
Usage: pnpm --filter portal sandbox:hemi-earn:setup -- --address 0xYourEOA
```

— Gabi 🤖

gabmontes (commented):
One minor leftover from the workspace→portal migration.

portal/scripts/hemi-earn/setup.ts
ArturDolzan: Good catch — fixed in [`b47c121e`](https://github.com/hemilabs/ui-monorepo/pull/2093/commits/b47c121e). Now matches the README quick-start.

gndelia (approved):

gabmontes (approved):

PR #2095, opened 2026-07-15
Add remote-failure recovery to Hemi Earn deposits

### LINKED ISSUE #1943: hemi-earn-actions: add Agent-side write wrappers (cancel, retry, claimUnstake)

### LINKED ISSUE #2028: Hemi Earn — redeem cooldown maturity finalize (Agent.claimUnstake) + unified state machine

ArturDolzan:
### Description

This PR extends the Hemi Earn remote-failure recovery — shipped for redeems in #2083 — to deposits. On the Agent side the mechanism is symmetric: failed deposits land in the same failedRequests mapping and are resolved by the same retry/cancel, so a deposit that reverts on Ethereum (gateway slippage, out-of-gas) is just as recoverable. Until now it only showed a red "Tx Failed" with no way out.

The core change is dropping the redeem-only gate on isRemoteFailed, which lets the existing badge, banner, CTA and step machinery light up for deposits too. From there it's mostly wiring the pieces into the two deposit drawers (live + historical) and handling the deposit-vs-redeem asymmetries: a deposit cancel returns the funds (not shares), so the CTA reads "Return funds" and the fee top-up is quoted against the asset OFT instead of the share OFT.

I also cleaned up two things left behind in the shared flow when the cancel button was renamed from "Cancel" to "Return ...": the in-flight label still said "Cancelling..." (now "Returning funds/shares..."), and the terminal step kept showing the fulfillment token during a cancel instead of the token actually being returned. Both fixes apply to redeem as well.

### Screenshots

Captura de Tela 2026-07-15 às 09 46 23

Table

Captura de Tela 2026-07-15 às 10 24 20

### Related issue(s)

Closes #1943
Closes #2028 
Related to #1848 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
@@ -15,19 +15,24 @@ import { type ReactNode } from 'react'
 import { type EvmToken } from 'types/token'
 import { type Hash } from 'viem'
 
+import { useRemoteFailedState } from '../../../_hooks/useRemoteFailedState'
 import { SparkleIcon } from '../../../_icons/sparkleIcon'
 import {
   getTerminalDeliveryTxHash,
   isLocalEarnTransactionRow,
   isRecoverPath,
+  isRemoteFailed,
+  isRemoteFailedCancel,
   needsManualClaim,
   needsRecover,
+  remoteFailedStepStatus,
 } from '../../../_utils'
 import {
   type EarnTransaction,
   type EarnTransactionStatusType,
 } from '../../../types'
 
+import { RemoteFailedBanner } from './remoteFailed'
 import { SettleBanner } from './settleShared'
 
 type Props = {
@@ -123,23 +128,27 @@ function resolveTerminalStatus(
 // Recover path returns the asset, so its terminal step is labeled with the asset token.
 function buildTerminalStep({
   baseStatus,
+  remoteFailedReady,
   settlementTxHash,
   t,
   token,
   tx,
 }: {
   baseStatus: ProgressStatusType
+  remoteFailedReady: boolean
   settlementTxHash: Hash | undefined
   t: Translator
   token: EvmToken
   tx: EarnTransaction
 }): StepPropsWithoutPosition {
-  const status = resolveTerminalStatus(tx, baseStatus, settlementTxHash)
+  const status = isRemoteFailed(tx)
+    ? remoteFailedStepStatus(remoteFailedReady, tx.settlement)
+    : resolveTerminalStatus(tx, baseStatus, settlementTxHash)
   const txHash = settlementTxHash ?? getTerminalDeliveryTxHash(tx)
   // Link whenever a delivery hash exists — an auto-finalized claim/recover has one the user never signed.
   const explorerChainId = txHash ? hemi.id : undefined
-  if (isRecoverPath(tx)) {
-    // "Funds returned" only once RECOVERED; CANCELLED still awaits the recover.
+  if (isRecoverPath(tx) || isRemoteFailedCancel(tx)) {
+    // "Funds returned" only once RECOVERED; CANCELLED / a signed remote-failed cancel awaits the recover.
     const recoverLabel =
       tx.status === 'RECOVERED'
         ? 'step.funds-returned'
@@ -200,6 +209,8 @@ export const HistoricalDepositReview = function ({
     spender: getHemiEarnRouterAddress(),
   })
 
+  const { show: remoteFailedReady } = useRemoteFailedState(transaction)
+
   const { waitingForShares } = resolveStepStates(transaction)
   // Only a still-mining settlement drives PROGRESS; a failed one keeps the natural (FAILED) status + Retry CTA.
   const { settlement } = transaction
@@ -231,6 +242,7 @@ export const HistoricalDepositReview = function ({
   steps.push(
     buildTerminalStep({
       baseStatus: waitingForShares,
+      remoteFailedReady,
       settlementTxHash,
       t,
       token,
@@ -240,7 +252,12 @@ export const HistoricalDepositReview = function ({
 
   return (
     <Operation
-      aboveCallToAction={<SettleBanner transaction={transaction} />}
+      aboveCallToAction={
+        <>
+          <SettleBanner transaction={transaction} />
+          <RemoteFailedBanner transaction={transaction} />
+        </>
+      }
       amount={transaction.amountIn}
       callToAction={callToAction}
       heading={t('deposit-heading')}
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
@@ -10,7 +10,6 @@ import { TokenLogo } from 'components/tokenLogo'
 import { getHemiEarnRouterAddress } from 'hemi-earn-actions'
 import { hemi } from 'hemi-viem'
 import { useNeedsApproval } from 'hooks/useNeedsApproval'
-import { mainnet } from 'networks/mainnet'
 import { useTranslations } from 'next-intl'
 import { type ReactNode } from 'react'
 import { type EvmToken } from 'types/token'
@@ -26,8 +25,8 @@ import {
   isRemoteFailedCancel,
   needsManualClaim,
   needsRecover,
-  remoteFailedSettlement,
   remoteFailedStepStatus,
+  resolveStepExplorerChainId,
 } from '../../../_utils'
 import {
   type EarnTransaction,
@@ -147,13 +146,11 @@ function buildTerminalStep({
     ? remoteFailedStepStatus(remoteFailedReady, tx.settlement)
     : resolveTerminalStatus(tx, baseStatus, settlementTxHash)
   const txHash = settlementTxHash ?? getTerminalDeliveryTxHash(tx)
-  // A remote-failed retry/cancel is signed on the Agent's L1, so link it to mainnet, not Hemi.
-  const recoveryTxHash = remoteFailedSettlement(tx.settlement)?.txHash
-  const explorerChainId = txHash
-    ? txHash === recoveryTxHash
-      ? mainnet.id
-      : hemi.id
-    : undefined
+  const explorerChainId = resolveStepExplorerChainId({
+    fallbackChainId: hemi.id,
+    settlement: tx.settlement,
+    txHash,
+  })
   if (isRecoverPath(tx) || isRemoteFailedCancel(tx)) {
     // "Funds returned" only once RECOVERED; CANCELLED / a signed remote-failed cancel awaits the recover.
     const recoverLabel =
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
@@ -10,6 +10,7 @@ import { TokenLogo } from 'components/tokenLogo'
 import { getHemiEarnRouterAddress } from 'hemi-earn-actions'
 import { hemi } from 'hemi-viem'
 import { useNeedsApproval } from 'hooks/useNeedsApproval'
+import { mainnet } from 'networks/mainnet'
 import { useTranslations } from 'next-intl'
 import { type ReactNode } from 'react'
 import { type EvmToken } from 'types/token'
@@ -25,6 +26,7 @@ import {
   isRemoteFailedCancel,
   needsManualClaim,
   needsRecover,
+  remoteFailedSettlement,
   remoteFailedStepStatus,
 } from '../../../_utils'
 import {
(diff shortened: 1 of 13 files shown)


portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx
```diff
@@ -123,23 +128,27 @@ function resolveTerminalStatus(
...
+  const status = isRemoteFailed(tx)
+    ? remoteFailedStepStatus(remoteFailedReady, tx.settlement)
+    : resolveTerminalStatus(tx, baseStatus, settlementTxHash)
   const txHash = settlementTxHash ?? getTerminalDeliveryTxHash(tx)
   // Link whenever a delivery hash exists — an auto-finalized claim/recover has one the user never signed.
   const explorerChainId = txHash ? hemi.id : undefined
-  if (isRecoverPath(tx)) {
-    // "Funds returned" only once RECOVERED; CANCELLED still awaits the recover.
+  if (isRecoverPath(tx) || isRemoteFailedCancel(tx)) {
+    // "Funds returned" only once RECOVERED; CANCELLED / a signed remote-failed cancel awaits the recover.
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/historicalDepositReview.tsx:155]: Good catch — the retry/cancel run on the Agent's L1, so I linked the step to mainnet instead of Hemi for those recovery markers in [`0e647c11`](https://github.com/hemilabs/ui-monorepo/pull/2095/commits/0e647c11).

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx
```diff
@@ -311,16 +322,18 @@ export const ReviewDeposit = function ({ onClose }: Props) {
         </div>
       ),
       explorerChainId: deliveryHash ? chainId : undefined,
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx]: Same fix here — the remote-failed recovery tx now links to mainnet in [`0e647c11`](https://github.com/hemilabs/ui-monorepo/pull/2095/commits/0e647c11).

gndelia:
Oh I like this

image

showing the actual problem and not a regular "Failed" - We definitely should implement things like this on other contract interactions 👏🏽

portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts
```diff
@@ -89,20 +89,22 @@ const useRemoteFailedAction = function ({
...
-          : await quoteDepositFulfillment({
-              agentAddress,
-              client,
-              share: remoteShare,
-            })
+      // Quote the OFT the action actually sends: retry sends the fulfillment token
+      // (deposit → share, redeem → asset), cancel returns tokenIn (deposit → asset,
+      // redeem → share). Sizing against the wrong OFT would under/over-fund the top-up.
+      const usesShareOft =
+        (transaction.kind === 'DEPOSIT') === (kind === 'RETRY')
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts]: I found this hard to read. I think this one is equivalent and reads better:


```suggestion
        transaction.kind === kind &&  ['DEPOSIT', 'RETRY'].includes(kind)
```

Thoughts?

portal/app/[locale]/hemi-earn/_utils/index.ts
```diff
@@ -68,11 +68,7 @@ export const canRetryRow = (tx: EarnTransaction) =>
 // Subgraph FAILED sibling of canRetryRow: the Agent reverted on Ethereum after a good Hemi
 // tx, so the request is stuck remotely and the user drives Agent.retry / Agent.cancel.
 export const isRemoteFailed = (tx: EarnTransaction | undefined) =>
-  !!tx &&
-  tx.kind === 'REDEEM' &&
-  tx.status === 'FAILED' &&
-  tx.failed &&
-  !isLocalEarnTransactionRow(tx)
+  !!tx && tx.status === 'FAILED' && tx.failed && !isLocalEarnTransactionRow(tx)
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_utils/index.ts]: Ultra minor but I think the `!!` are redundant?
```suggestion
  tx && tx.status === 'FAILED' && tx.failed && !isLocalEarnTransactionRow(tx)
```

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx
```diff
@@ -310,17 +325,23 @@ export const ReviewDeposit = function ({ onClose }: Props) {
...
-          depositStatus === DepositStatus.DEPOSIT_TX_CONFIRMED,
-        isFailed,
-        isRecover,
-        settlementFailed,
-      }),
+      explorerChainId: deliveryHash
+        ? deliveryHash === recoveryTxHash
+          ? mainnet.id
+          : chainId
+        : undefined,
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx]: maybe worth extracting as this matches https://github.com/hemilabs/ui-monorepo/pull/2095/changes#diff-d2230ed1d7c98a2f759a9c4f145e6cca7691d5d8feaeadc1778720b058186ec2R152 

Thoughts?

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/remoteFailed.tsx
```diff
@@ -142,19 +149,13 @@ export const RemoteFailedBanner = function ({
...
-            ? 'remote-failed.heading-slippage'
-            : 'remote-failed.heading',
-        )}
-        subheading={t(
-          category === 'slippage'
-            ? 'remote-failed.subheading-slippage'
-            : 'remote-failed.subheading',
-        )}
+        heading={t(`remote-failed.heading${variant}${suffix}`)}
+        subheading={t(`remote-failed.subheading${variant}${suffix}`)}
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/remoteFailed.tsx:158]: We are loosing type safety here. But perhaps fixing that would make the code like over-engineered. Thoughts?

portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useRemoteFailedActions.ts]: Thanks — agree the original was cryptic. Heads up though, your version isn't equivalent: there are two `kind`s in scope — `transaction.kind` (DEPOSIT/REDEEM) and the action `kind` (RETRY/CANCEL_REQUEST) — so `transaction.kind === kind` is always false. Renamed them to `isDeposit`/`isRetry` so it reads clearly now in [`49131286`](https://github.com/hemilabs/ui-monorepo/pull/2095/commits/49131286).

portal/app/[locale]/hemi-earn/_utils/index.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/index.ts]: Good eye — not fully redundant though: without the `!!`, `isRemoteFailed(undefined)` returns `undefined` instead of `false` (and the return type widens to `boolean | undefined`). Swapped it to `tx !== undefined` to drop the `!!` while keeping the strict boolean, in [`49131286`](https://github.com/hemilabs/ui-monorepo/pull/2095/commits/49131286).

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/poolReview/reviewDeposit.tsx]: Good call — pulled it into a shared `resolveStepExplorerChainId` helper (with tests) in [`49131286`](https://github.com/hemilabs/ui-monorepo/pull/2095/commits/49131286).

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/remoteFailed.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/remoteFailed.tsx:158]: Yeah, fair tradeoff — though TS actually still validates it: the template resolves to the four `heading`/`subheading` combos, so a typo'd key fails to compile. E.g. dropping the "g" in `heading` gives `'remote-failed.headin' is not assignable to NamespacedMessageKeys`. Spelling all the keys out explicitly felt more verbose for the same safety, so I'd lean toward keeping it... but happy to change it if you'd prefer.

Captura de Tela 2026-07-15 às 16 54 32

gndelia (approved):

gabmontes (approved):

PR #2100, opened 2026-07-17
Reload tunnel history when the account changes

### LINKED ISSUE #1558: Transaction History does not update when the user switches account
### Description

When the user switches account (through their wallet), assuming both accounts are connected to the portal, the balances in the tunnel form update, but the TX history does not - the tx history from the first account remains rendered

### Steps to reproduce

1. Connect your wallet with one address
2. Go to the tx history (Where there are transactions)
3. USing you wallet, switch to another address. Ensure the new address is connected to the site. The "Connect Wallet" drawer can be used to pick up that the right wallet is being detected

Expected:
The transaction history from the second wallet should be listed

Actual:
The tx history from the previous wallet is still listed

Godbrand0:
Closes #1558

## Problem

Switching accounts in the wallet (both connected to the portal) kept the previous account's transaction history rendered. The history loader tracked restoration with a plain `loadedFromLocalStorage` boolean:

- The restore effect bailed out because the flag was still `true` after a switch, so the reducer kept the old account's operations.
- The reset effect only fired on disconnect/unsupported chain — it didn't react to the address changing.
- Worse: the save effect *does* depend on `address`, so right after a switch it re-ran with the **new** address while the reducer still held the **old** account's history, persisting account A's operations under account B's localStorage keys. After that, even a page refresh showed the wrong history.

## Fix

Replace the boolean with `loadedAddress` — the address whose history is loaded in the reducer:

- The state resets whenever the connected address differs from the loaded one (not just on disconnect), and the restore effect then reloads from the new account's storage keys.
- Saving to localStorage is guarded with `address === loadedAddress`, which also covers the same-commit race where the save effect runs before the reset takes effect.
- The manual resync path clears `loadedAddress`, keeping its behavior unchanged.

The sync workers were already keyed by address, so they restart on their own and now dispatch into a reducer holding the right account's state.

Note: the fix prevents new corruption but does not clean up localStorage already polluted by previous switches — the existing resync button handles that.

## Screenshots

image

image

### DIFF
--- a/portal/context/tunnelHistoryContext/historyLoader.tsx
+++ b/portal/context/tunnelHistoryContext/historyLoader.tsx
@@ -247,6 +247,9 @@ export const HistoryLoader = function ({
       if (!forceResync) {
         return
       }
+      // cancel any pending debounced save - otherwise its trailing call can
+      // run after the clear below and repopulate storage with stale data
+      debouncedSaveToStorage.cancel()
       // clear local storage
       clearHistoryInLocalStorage({
         address: address!,
--- a/portal/context/tunnelHistoryContext/historyLoader.tsx
+++ b/portal/context/tunnelHistoryContext/historyLoader.tsx
@@ -145,32 +145,30 @@ export const HistoryLoader = function ({
   const { remoteNetworks } = useNetworks()
   const [networkType] = useNetworkType()
 
-  // use this boolean to check if the past history was restored from local storage
-  const [loadedFromLocalStorage, setLoadedFromLocalStorage] = useState(false)
+  // the address whose history is loaded in the reducer - tracking it (instead of
+  // a boolean) lets the state reset and reload when the user switches accounts
+  const [loadedAddress, setLoadedAddress] = useState<Address | undefined>(
+    undefined,
+  )
 
   const supportedEvmChain = useConnectedToSupportedEvmChain()
 
   useEffect(
     function resetState() {
-      if (!supportedEvmChain || !loadedFromLocalStorage) {
-        setLoadedFromLocalStorage(false)
+      if (!supportedEvmChain || address !== loadedAddress) {
+        setLoadedAddress(undefined)
         dispatch({ type: 'reset' })
       }
     },
-    [
-      loadedFromLocalStorage,
-      dispatch,
-      setLoadedFromLocalStorage,
-      supportedEvmChain,
-    ],
+    [address, dispatch, loadedAddress, setLoadedAddress, supportedEvmChain],
   )
 
   useEffect(
     function restoreFromLocalStorage() {
-      if (!address || loadedFromLocalStorage || !supportedEvmChain) {
+      if (!address || loadedAddress || !supportedEvmChain) {
         return
       }
-      setLoadedFromLocalStorage(true)
+      setLoadedAddress(address)
       // Load all the deposits given the Hemi address
       const deposits = remoteNetworks
         .map(
@@ -204,9 +202,9 @@ export const HistoryLoader = function ({
       address,
       dispatch,
       l2ChainId,
-      loadedFromLocalStorage,
+      loadedAddress,
       remoteNetworks,
-      setLoadedFromLocalStorage,
+      setLoadedAddress,
       supportedEvmChain,
     ],
   )
@@ -216,7 +214,10 @@ export const HistoryLoader = function ({
       if (
         !address ||
         !supportedEvmChain ||
-        !loadedFromLocalStorage ||
+        // only save the history of the address it belongs to - on an account
+        // switch, this effect re-runs with the new address while the reducer
+        // still holds the previous account's history
+        address !== loadedAddress ||
         !['finished', 'syncing'].includes(history.status) ||
         // if we started resync, do not save!
         forceResync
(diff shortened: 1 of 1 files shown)


Godbrand0:
Good catch, this gap predates the account-switch fix (the comment above it even calls out the need to avoid a stale save overriding the clear, but the code never actually canceled the debounce). Fixed in 04383a6a by calling debouncedSaveToStorage.cancel() right before clearHistoryInLocalStorage runs, so no queued trailing write can survive the reset.

portal/context/tunnelHistoryContext/historyLoader.tsx
```diff
@@ -255,7 +256,7 @@ export const HistoryLoader = function ({
       // reset the history in memory
       dispatch({ type: 'reset' })
       // update flag so data is reloaded again
-      setLoadedFromLocalStorage(false)
+      setLoadedAddress(undefined)
       // mark resync as finished
```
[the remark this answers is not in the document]
Godbrand0: Good catch, this gap predates the account-switch fix (the comment above it even calls out the need to avoid a stale save overriding the clear, but the code never actually canceled the debounce). Fixed in https://github.com/hemilabs/ui-monorepo/commit/04383a6a2fbd678eedab33449b348976c65becb3 by calling debouncedSaveToStorage.cancel() right before clearHistoryInLocalStorage runs, so no queued trailing write can survive the reset.

arielmergen (dismissed):
Thanks for your contribution. Could you please take another look?

portal/context/tunnelHistoryContext/historyLoader.tsx
```diff
@@ -145,32 +145,30 @@ export const HistoryLoader = function ({
...
   const supportedEvmChain = useConnectedToSupportedEvmChain()
 
   useEffect(
     function resetState() {
-      if (!supportedEvmChain || !loadedFromLocalStorage) {
-        setLoadedFromLocalStorage(false)
+      if (!supportedEvmChain || address !== loadedAddress) {
+        setLoadedAddress(undefined)
         dispatch({ type: 'reset' })
       }
```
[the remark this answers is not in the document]
Godbrand0: Good catch — flushed the pending debounced save in the `resetState` effect (before dispatching `reset`) so the outgoing account's queued history/sync cursor is persisted before the incoming account's save call can overwrite the shared debounce's queued args. Kept `cancel()` (not flush) on the resync path, since that one intentionally discards pending data. Fixed in 2885ad42, along with rebasing the branch onto latest main.

Godbrand0:
can you check it out??

portal/context/tunnelHistoryContext/historyLoader.tsx
```diff
@@ -145,32 +145,36 @@ export const HistoryLoader = function ({
   const { remoteNetworks } = useNetworks()
   const [networkType] = useNetworkType()
 
-  // use this boolean to check if the past history was restored from local storage
-  const [loadedFromLocalStorage, setLoadedFromLocalStorage] = useState(false)
+  // the address whose history is loaded in the reducer - tracking it (instead of
```
arielmergen: Hi @Godbrand0, could you remove these comments? We try to keep the code clean, so please only add comments when they’re really necessary.

portal/context/tunnelHistoryContext/historyLoader.tsx
```diff
@@ -145,32 +145,36 @@ export const HistoryLoader = function ({
...
+  )
 
   const supportedEvmChain = useConnectedToSupportedEvmChain()
 
   useEffect(
     function resetState() {
-      if (!supportedEvmChain || !loadedFromLocalStorage) {
-        setLoadedFromLocalStorage(false)
+      if (!supportedEvmChain || address !== loadedAddress) {
+        // flush any pending debounced save for the previous account before
```
arielmergen: Same here — I don’t think this comment is necessary.

arielmergen: @Godbrand0 Hi! When you have a chance, could you take another look at this?

Godbrand0: Thanks for taking care of it directly — I see the comments are already removed in e012a88a. Nothing further needed from my side here.

portal/context/tunnelHistoryContext/historyLoader.tsx
arielmergen: Resolved in [e012a88](https://github.com/hemilabs/ui-monorepo/pull/2100/commits/e012a88aaf7a6f28cd33920e430e6754c6e06b3f)

arielmergen (dismissed):

portal/context/tunnelHistoryContext/historyLoader.tsx
gabmontes: @arielmergen @Godbrand0 there are two more comments that might be worth removing.

Godbrand0: Good catch — found and removed the two remaining comments this PR had added (in offloadToStorage and clearAndResyncHistory), left the pre-existing ones untouched. Fixed in 24bbee60.

arielmergen (approved):

PR #2103, opened 2026-07-18
Validate stake drawer mode before rendering

### LINKED ISSUE #2064: TypeError: Cannot destructure property 'heading' of '{(intermediate value)(intermediate value)}[a]' as it is undefined.
Sentry Issue: [PORTAL-287](https://hemi-labs.sentry.io/issues/131182384/?referrer=github_integration)

```
TypeError: Cannot destructure property 'heading' of '{(intermediate value)(intermediate value)}[a]' as it is undefined.
  at eF (./app/[locale]/stake/_components/manageStake/index.tsx:26:5)
...
(5 additional frame(s) were not displayed)
```

Godbrand0:
Closes #2064

## Problem

Sentry PORTAL-287: `TypeError: Cannot destructure property 'heading' of '...[a]' as it is undefined` in `ManageStake`. The `heading`/`subheading` lookup only has `manage` and `stake` keys, so any other `mode` value crashes the stake page.

## Root cause

Three pages register the **same `drawerMode` query key** with disjoint literal sets:

| Page | Allowed values |
|---|---|
| stake | `manage`, `stake` |
| staking-dashboard | `staking`, `unlocking`, `claimingRewards`, `increasingAmount`, `increasingUnlockTime` |
| hemi-earn pool | `depositing`, `withdrawing` |

Reading from the URL is safe (`parseAsStringLiteral` nulls invalid values), but nuqs setters bypass per-hook parsing: `emitter.emit(urlKey, { state: value })` pushes the raw value to every hook subscribed to that key ("cross-hook key sync"), which stores it without running its own literal check. So if another page's component sets `drawerMode` while the stake page's `SideDrawer` is still mounted (e.g. during an App Router transition, where the previous page stays rendered), the stake hook receives a foreign value. It passes the previous truthy-only guard, and `ManageStake` crashes on the lookup.

The staking-dashboard and hemi-earn drawers are incidentally protected because their guards also require in-memory operation state that only exists mid-flow on their own pages.

## Fix

Export an `isDrawerMode` type guard from `useDrawerStakeQueryString` and use it in `SideDrawer` — the single render site of `ManageStake` — so only `manage`/`stake` open the drawer. A foreign value now renders no drawer (correct, since it belongs to a different page's drawer) instead of crashing. TypeScript narrowing keeps the `mode` prop typed with no casts.

## Follow-up suggestion

Giving each page a unique query key (e.g. `stakeDrawerMode`, `stakingDrawerMode`, `poolDrawerMode`) would eliminate this class of collision entirely. Left out of this PR since it changes user-visible URLs — happy to do it as a follow-up if you want it.

### DIFF
--- a/portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts
+++ b/portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts
@@ -3,6 +3,12 @@ import { parseAsString, parseAsStringLiteral, useQueryState } from 'nuqs'
 const drawerModes = ['manage', 'stake'] as const
 export type DrawerModes = (typeof drawerModes)[number]
 
+// Needed on top of the parser's validation: nuqs syncs values set by other
+// hooks sharing the "drawerMode" url key (other pages define their own modes
+// for it) without re-parsing, so an invalid value can reach consumers.
+export const isDrawerMode = (value: string | null): value is DrawerModes =>
+  drawerModes.includes(value as DrawerModes)
+
 export const useDrawerStakeQueryString = function () {
   const [tokenAddress, setTokenAddress] = useQueryState(
     'tokenAddress',
--- a/portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts
+++ b/portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts
@@ -3,11 +3,8 @@ import { parseAsString, parseAsStringLiteral, useQueryState } from 'nuqs'
 const drawerModes = ['manage', 'stake'] as const
 export type DrawerModes = (typeof drawerModes)[number]
 
-// Needed on top of the parser's validation: nuqs syncs values set by other
-// hooks sharing the "drawerMode" url key (other pages define their own modes
-// for it) without re-parsing, so an invalid value can reach consumers.
-export const isDrawerMode = (value: string | null): value is DrawerModes =>
-  drawerModes.includes(value as DrawerModes)
+export const isDrawerMode = (value: unknown): value is DrawerModes =>
+  (drawerModes as readonly unknown[]).includes(value)
 
 export const useDrawerStakeQueryString = function () {
   const [tokenAddress, setTokenAddress] = useQueryState(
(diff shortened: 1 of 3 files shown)


portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts
```diff
@@ -3,6 +3,12 @@ import { parseAsString, parseAsStringLiteral, useQueryState } from 'nuqs'
 const drawerModes = ['manage', 'stake'] as const
 export type DrawerModes = (typeof drawerModes)[number]
 
+// Needed on top of the parser's validation: nuqs syncs values set by other
+// hooks sharing the "drawerMode" url key (other pages define their own modes
+// for it) without re-parsing, so an invalid value can reach consumers.
```
Comment (gndelia) [inline portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts]: Let's remove these IA generated comments that add no value
```suggestion
```

portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts
```diff
@@ -3,6 +3,12 @@ import { parseAsString, parseAsStringLiteral, useQueryState } from 'nuqs'
...
+export const isDrawerMode = (value: string | null): value is DrawerModes =>
+  drawerModes.includes(value as DrawerModes)
```
Comment (gndelia) [inline portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts]: although very minimal, let's add a unit test to this

Godbrand0:
1 Removed the explanatory comment.
2 Added a unit test (test/app/[locale]/stake/_hooks/useDrawerStakeQueryString.test.ts) covering both valid modes plus invalid inputs (null, undefined, foreign mode strings, non-string values).
3 Tightened isDrawerMode to accept unknown and check against the drawerModes array directly, dropping the unsound cast on the checked value.

portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts
Comment (Godbrand0) [reply portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts]: Added a unit test (test/app/[locale]/stake/_hooks/useDrawerStakeQueryString.test.ts) covering both valid modes plus invalid inputs (null, undefined, foreign mode strings, non-string values).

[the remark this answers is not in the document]
Comment (Godbrand0) [reply portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts]: Tightened isDrawerMode to accept unknown and check against the drawerModes array directly, dropping the unsound cast on the checked value.

portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts
Comment (Godbrand0) [reply portal/app/[locale]/stake/_hooks/useDrawerStakeQueryString.ts]: Removed the explanatory comment.

gndelia (approved):

arielmergen (approved):

PR #2105, opened 2026-07-20
Add debounce

### LINKED ISSUE #1810: Debounce when the user inputs Bitcoin to tunnel to HemiBTC
### Description

When the user enters a value in the Tunnel form for tunneling Bitcoin, there's an async calculation. The value entered is not debounced, so on each keystroke, we're sending a request to calculate the expected hemiBTC output.

### Steps to reproduce

See video

https://github.com/user-attachments/assets/d14d7b6d-f887-4600-9dc3-19647f4546fe

Basically, on each keystroke, there's a request to calculate how much hemiBTC the user may get.

### Environment

- **OS**: _e.g., Windows, macOS, Linux_
- **Runtime/Browser/Version**: _e.g., Node 22, Chrome 120_
- **Wallet**: _e.g., MetaMask, Trust, Coinbase Wallet_

### Proposed fix

We should debounce the input value
Note: You may want to implement #1807  first (as it gives access to useDebounce)

arielmergen:
### Description

Tunneling BTC to hemiBTC fired a fee-estimate request on every keystroke. This debounces (300 ms) the amount that drives that request, so a burst of keystrokes collapses into a single request. Uses useDebounce from @hemilabs/react-hooks (already used elsewhere):

const debouncedFromInput = useDebounce(fromInput, 300)
const amountBigInt = parseTokenUnits(debouncedFromInput, fromToken)

Why debounce the derived amount and not the input itself (as the ticket suggests): fromInput keeps updating instantly, so the field stays responsive and validateSubmit/submit react immediately. Only the network-bound fee estimate is throttled. Each new amount still recalculates its own fee.

### Screenshots

https://github.com/user-attachments/assets/e74d5a12-842c-426c-aca8-53c6835472ea

### Related issue(s)

Closes #1810 
Fixes #
Related to #

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/tunnel/_components/btcDeposit.tsx
+++ b/portal/app/[locale]/tunnel/_components/btcDeposit.tsx
@@ -1,5 +1,6 @@
 'use client'
 
+import { useDebounce } from '@hemilabs/react-hooks/useDebounce'
 import { useAccounts } from 'hooks/useAccounts'
 import { useBitcoin } from 'hooks/useBitcoin'
 import { useBitcoinBalance } from 'hooks/useBitcoinBalance'
@@ -95,7 +96,8 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
     isMinDepositsSatsLoading,
   })
 
-  const amountBigInt = parseTokenUnits(fromInput, fromToken)
+  const debouncedFromInput = useDebounce(fromInput, 300)
+  const amountBigInt = parseTokenUnits(debouncedFromInput, fromToken)
 
   const {
     btcDepositFee,


portal/app/[locale]/tunnel/_components/btcDeposit.tsx
```diff
@@ -95,7 +96,8 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
     isMinDepositsSatsLoading,
   })
 
-  const amountBigInt = parseTokenUnits(fromInput, fromToken)
+  const debouncedFromInput = useDebounce(fromInput, 300)
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/btcDeposit.tsx:99]: 300 is the default value, so it's redundant

```suggestion
  const debouncedFromInput = useDebounce(fromInput)
```

gndelia (approved):
I left a minor suggestion, but LGTM

ArturDolzan (approved):

Issue #2094, opened 2026-07-14
Add support for mooStakeDao-VUSD-crvUSD token in the tunnel


gndelia:
### Feature or change description

We need to add to the tunnel the token `mooStakeDao-VUSD-crvUSD` named `Beefy Finance VUSD/crvUSD` (We need a way to simplify that long symbol) 

L1 Ethereum address: `0x1b47d70e564afF83F09515B405c8af32a0f988a4`
L2 Hemi address: `0xfB2E53fAac872e875fbf2cf21ed9372f4422aa95`

Website https://app.beefy.com/vault/stakedao-vusd-crvusd

TBD token logo

dvnahuel:
I designed this logo, thoughts? 

<img width="32" height="32" alt="Image" src="https://github.com/user-attachments/assets/b951f53e-5322-407b-b06c-120f51ce8efb" />

PR #2107, opened 2026-07-20
Truncate token symbols

arielmergen:
### Description

Token symbols come from the contract, so they can be arbitrarily long.
`mooStakeDao-VUSD-crvUSD` is 23 characters and wrapped onto two lines in the
token selector. Symbols are now truncated by the CSS to the width available, and
the full symbol is shown in a tooltip — same approach already used for tokens in
`tokenTable/tokenCell.tsx`.

Two fixes were needed along the way:

- `TokenLogo` had no `shrink-0`, so it collapsed to 0px once the text no longer
  fit. The sibling `Chevron` already sets it.
- The tooltip's default `trigger` includes `click`, which left it stuck open
  over the modal the selector opens. Restricted to `hover` and `focus`.

### Screenshots

https://github.com/user-attachments/assets/9203ea8d-1f85-4b35-934a-6d7de8f6ad98

Captura de pantalla 2026-07-20 a la(s) 3 14 11 p  m

Captura de pantalla 2026-07-20 a la(s) 3 37 41 p  m

### Related issue(s)

Related to #2094 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/components/tokenSelector/tokenSymbol.tsx
+++ b/portal/components/tokenSelector/tokenSymbol.tsx
@@ -10,7 +10,7 @@ type Props = {
 // Truncation is done by the CSS, so the symbol adapts to the width available.
 // Both the selector and the read-only row bound it to the same width, so they
 // truncate alike.
-export const TokenSymbol = ({ className, symbol }: Props) => (
+export const TokenSymbol = ({ className = '', symbol }: Props) => (
   <Tooltip
     disabled={!isSymbolTooLong(symbol)}
     id={`token-symbol-${symbol}`}
--- a/portal/components/tokenSelector/tokenSymbol.tsx
+++ b/portal/components/tokenSelector/tokenSymbol.tsx
@@ -1,0 +1,24 @@
+import { Tooltip } from 'components/tooltip'
+
+import { isSymbolTooLong } from './utils'
+
+type Props = {
+  className?: string
+  symbol: string
+}
+
+// Truncation is done by the CSS, so the symbol adapts to the width available.
+// Both the selector and the read-only row bound it to the same width, so they
+// truncate alike.
+export const TokenSymbol = ({ className, symbol }: Props) => (
+  <Tooltip
+    disabled={!isSymbolTooLong(symbol)}
+    id={`token-symbol-${symbol}`}
+    text={symbol}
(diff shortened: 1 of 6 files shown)


portal/components/tokenSelector/utils.ts
```diff
@@ -0,0 +1,8 @@
+// Symbols are read from the token contract, so they can be arbitrarily long
+// (i.e. mooStakeDao-VUSD-crvUSD). The CSS truncates them to the width the
+// layout gives them - this is only used to decide whether the full symbol is
+// worth showing in a tooltip.
+export const maxSymbolLength = 10
+
+export const isSymbolTooLong = (symbol: string) =>
+  symbol.length > maxSymbolLength
```
gndelia: 2 comments re: this file

1. I'll probably move that comment into jsdoc of `isSymbolTooLong`
2. As simple as the function is, let's add a few tests to it. With IA, this is a very small prompt away!

gndelia: Ah, nevermind, I found the test! Skip item 2

arielmergen: Moved the comment into the JSDoc of `isSymbolTooLong`. Left `maxSymbolLength` without one, since the explanation is about the function rather than the constant.

portal/components/tokenSelector/tokenSymbol.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const TokenSymbol = ({ className, symbol }: Props) => (
+  <Tooltip
+    disabled={!isSymbolTooLong(symbol)}
+    id={`token-symbol-${symbol}`}
+    text={symbol}
+    // the selector opens a modal on click, so the tooltip must not react to it
+    trigger={['hover', 'focus']}
+    variant="simple"
+  >
+    <span className={`block max-w-20 truncate ${className}`}>{symbol}</span>
```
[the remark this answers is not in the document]
arielmergen: Good catch — fixed with a default value in the destructuring (`className = ''`), so no `undefined` class ends up in the DOM.

gabmontes (approved):

ArturDolzan (approved):

gndelia (approved):

PR #2113, opened 2026-07-21
Restore placeholder for disabled balance queries

### LINKED ISSUE #2112: Balances are stuck on loading when wallet is disconnected
### Description

See how my wallet is disconnected, and balances are stuck in "Loading"

Image

### Steps to reproduce

Open the token selector in the tunnel while your wallet is disconnected. All balances should just read "-" as the queries should be disabled, but they're stuck on `-`

ArturDolzan:
### Description

This PR fixes token balances getting stuck on "Loading" when the wallet is disconnected.

If you open the token selector in the Tunnel without a connected wallet, the balance queries are disabled — so instead of resolving, they sit in react-query's 'pending' + 'idle' state forever. The balance renderers were treating that exactly like an active fetch and showing a perpetual skeleton, when they should just render a "-".

The cause was a refactor that moved the balance components to the data-first pattern and, along the way, dropped the fetchStatus check that told a disabled query apart from a loading one. This restores that check behind a small, tested isBalanceUnavailable helper so both the crypto and fiat balances fall back to "-" when there's nothing to fetch. Callers that already guard on connection state are untouched.

### Screenshots

https://github.com/user-attachments/assets/46c8bd1a-a8a2-4d0c-8d98-8e94b2062f1f

### Related issue(s)

Closes #2112
Fixes #2112 

### Checklist

- [x] Manual testing passed.


portal/components/cryptoBalance.tsx
```diff
@@ -75,10 +91,11 @@ const EvmBalance = (props: Props<EvmToken>) =>
   )
 
 const BtcBalance = function ({ token }: Props<BtcToken>) {
-  const { balance, status } = useBitcoinBalance()
+  const { balance, fetchStatus, status } = useBitcoinBalance()
   return (
     <RenderCryptoBalance
       balance={balance?.confirmed ? BigInt(balance?.confirmed) : undefined}
+      fetchStatus={fetchStatus}
```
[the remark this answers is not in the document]
ArturDolzan: Good catch — now checking `confirmed !== undefined` so a real `0` renders as `0` instead of `-`. Done in [`6ce79129`](https://github.com/hemilabs/ui-monorepo/pull/2113/commits/6ce79129).

portal/components/fiatBalance.tsx
```diff
@@ -97,10 +109,11 @@ const EvmBalance = (props: Props<EvmToken>) =>
   )
 
 const BtcBalance = function ({ token }: Props<BtcToken>) {
-  const { balance, status } = useBitcoinBalance()
+  const { balance, fetchStatus, status } = useBitcoinBalance()
   return (
     <RenderFiatBalance
       balance={BigInt(balance?.confirmed ?? 0)}
+      fetchStatus={fetchStatus}
       queryStatus={status}
```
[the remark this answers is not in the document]
ArturDolzan: Nice one — dropped the `?? 0` and only pass a balance once it's actually loaded, so a real `0` still shows while a pending/disconnected query falls back to `-`. Done in [`6ce79129`](https://github.com/hemilabs/ui-monorepo/pull/2113/commits/6ce79129).

portal/utils/balance.ts
```diff
@@ -0,0 +1,10 @@
+import { type FetchStatus, type QueryStatus } from '@tanstack/react-query'
+
+// A disabled query (e.g. no wallet connected) stays 'pending' + 'idle' and never resolves.
+export const isBalanceUnavailable = ({
+  fetchStatus,
+  status,
+}: {
+  fetchStatus?: FetchStatus
+  status: QueryStatus
+}) => status === 'error' || fetchStatus === 'idle'
```
gndelia: This method has nothing to do with specifics as _balance_, but rather, it's related to dataUnavailable (either by error or query disabled).

I'd make the naming more generic

portal/components/fiatBalance.tsx
```diff
@@ -50,8 +53,10 @@ const RenderFiatBalanceUnsafe = function ({
     )
   }
 
-  // Check for errors from either source
-  if (queryStatus === 'error' || pricesStatus === 'error') {
+  if (
+    isBalanceUnavailable({ fetchStatus, status: queryStatus }) ||
+    pricesStatus === 'error'
```
gndelia: if you do [this change](https://github.com/hemilabs/ui-monorepo/pull/2113/changes#r3623822921) you can probably use it for `price` query as well

gndelia:
**`isBalanceUnavailable` conflates "resolved successfully" with "disabled"**

The helper returns `true` whenever `fetchStatus === 'idle'`, but a **successfully resolved** query also settles to `fetchStatus: 'idle'` (with `status: 'success'`) — it's indistinguishable from the disabled `pending` + `idle` case by `fetchStatus` alone.

In `cryptoBalance.tsx` this is harmless: `balance !== undefined` returns early before the check is reached. But in `fiatBalance.tsx` the success path requires **both** balance and prices, so it isn't fully masked:

```ts
if (balance !== undefined && pricesData !== undefined) { /* success */ }
if (isBalanceUnavailable({ fetchStatus, status: queryStatus }) || pricesStatus === 'error') {
  return <>-</>
}
return <Skeleton />
```

Scenario — connected wallet, **balance resolves before prices** (`queryStatus: 'success'`, `fetchStatus: 'idle'`, `pricesData` still `undefined`, `pricesStatus: 'pending'`):

- success `if` → false (prices not ready yet)
- `isBalanceUnavailable({ fetchStatus: 'idle', status: 'success' })` → `'idle' === 'idle'` → **true** → renders `-`

Pre-refactor this rendered a `Skeleton` (`queryStatus === 'error'` was false). So this reintroduces a milder, inverted version of the same "loading looks unavailable" symptom this PR fixes: a flash of `-` while prices lag behind the balance, before flipping to the real fiat value. It only affects the fiat renderers that now pass `fetchStatus` (tunnel token selector), and it's transient — but avoidable.

**Suggested fix** — scope the disabled check to a still-pending query:

```ts
export const isBalanceUnavailable = ({
  fetchStatus,
  status,
}: {
  fetchStatus?: FetchStatus
  status: QueryStatus
}) => status === 'error' || (status === 'pending' && fetchStatus === 'idle')
```

This still passes all four existing tests and correctly returns `false` for `success` + `idle`. Worth adding a case asserting `{ fetchStatus: 'idle', status: 'success' } → false` to lock in the distinction.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2113#issuecomment-5036421877):
> **`isBalanceUnavailable` conflates "resolved successfully" with "disabled"**

Spot on — scoped the disabled check to `status === 'pending' && fetchStatus === 'idle'` so a resolved `success` + `idle` no longer flashes `-`, and locked it in with a test. Done in [`88141a8c`](https://github.com/hemilabs/ui-monorepo/pull/2113/commits/88141a8c).

portal/utils/balance.ts
ArturDolzan: Good call — renamed to `isDataUnavailable` and moved it to `utils/queryStatus.ts`, since it's really about query state, not balances. Done in [`88141a8c`](https://github.com/hemilabs/ui-monorepo/pull/2113/commits/88141a8c).

portal/components/fiatBalance.tsx
ArturDolzan: Done — using the same helper for the prices query now too (it also covers the case where the prices query itself is disabled). In [`88141a8c`](https://github.com/hemilabs/ui-monorepo/pull/2113/commits/88141a8c).

gabmontes (approved):

gndelia (approved):

Issue #1965, opened 2026-05-30
Hemi Earn: wire 'Total earned' card to real data source


ArturDolzan:
### Feature or change description

Replace the mocked `useEarnedAmount` hook with a real data source so the "Total earned" card on `/hemi-earn` reflects what the user has actually earned across positions.

### Problem or use case addressed

`portal/app/[locale]/hemi-earn/_hooks/useEarnedAmount.ts` currently returns `{ totalUsd: 0 }` after a 2s delay (`TODO: placeholder mock — returns $0 until the earned-amount data source is available`). The home card always shows $0.

### Proposed implementation

Decide on the data source (likely yield accrued = current `convertToAssets(shares)` minus original deposit principal tracked via subgraph) and replace the placeholder. Hook signature is consumed by `_components/earnedAmount.tsx`.

---

Follow-up from #1848.

gndelia:
Follow-up from reviewing #2125 (the cost-basis API step), worth resolving before "Total earned" is trusted.

**The cost-basis replay ignores `status`.** `getRequests` in `portal-backend/api/src/hemi-earn/costBasis.ts` filters rows on `processedAt: { _is_null: false }` only. Terminal transitions that happen *after* processing — `CANCELLED`, `RECOVERED` — are not excluded.

In the indexer these are independent fields: the Agent handlers (`DepositRequestProcessed` / `RedeemRequestProcessed`) write `processedAt` unconditionally, while `status` is driven separately by the Router events through `applyStatus`. So `processedAt != null AND status IN ('CANCELLED', 'RECOVERED')` is representable in the data model, regardless of whether the protocol currently allows it.

If it *is* reachable, the failure mode is silent and one-directional:

- A processed redeem that later goes `RECOVERED` returns the shares to the holder **through the Router**.
- The `ShareTransfer` handler deliberately skips Router legs (`if (routerAddresses.has(from) || routerAddresses.has(to)) return`), so that return leg is invisible to the replay.
- The replay still applies the original disposal, which reduces the position's cost basis proportionally.

Net: the holder's basis is permanently understated and "Total earned" is overstated — with no error surfaced, since the endpoint returns 200 either way.

Two things needed:

1. **Confirm reachability** — can a request reach `CANCELLED`/`RECOVERED` after `processedAt` is set? That's a protocol question the indexer alone can't answer.
[...]

PR #2121, opened 2026-07-23
Add compiled exports to hemi-earn-actions

ArturDolzan:
### Description

This PR makes the `hemi-earn-actions` package consumable from a plain Node runtime so the portal API can import it in a follow-up PR.

Right now the package only ships TypeScript sources, which works for the portal because Next transpiles them, but the API runs in Node and can't. This adds a `build` step plus conditional exports: a new `compiled` condition resolves to the built `dist/*.js`, while `default` keeps pointing at the `.ts` source the portal already uses — so nothing changes for existing consumers. Internal imports get explicit `.ts` extensions (needed by the TS build) and a tiny `actions/` shim keeps the public export path from reaching into `src/`.

It's the first of a short sequence (package → subgraph → api → ui) behind the new "Total earned" card; the API PR depends on this one to build in Docker.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #1965 

### Checklist

- [x] Manual testing passed.


arielmergen (approved):
LGMT

packages/hemi-earn-actions/tsconfig.build.json
```diff
@@ -0,0 +1,11 @@
...
+  "compilerOptions": {
+    "allowImportingTsExtensions": false,
+    "declaration": false,
+    "noEmit": false,
+    "outDir": "./dist",
+    "rewriteRelativeImportExtensions": true
+  },
+  "exclude": ["dist", "node_modules", "test", "vitest.config.ts"],
+  "extends": "./tsconfig.json"
+}
```
gndelia, line 11: I think this one can be simply

```
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "noEmit": false
  },
  "include": ["src/**/*.ts"]
}
```

At least on Vetro, packages are consumed by web and api and work fine using this config. Have you checked that implementation?

ArturDolzan: Good point. The catch is that in the Vetro monorepo the packages are consumed as raw `.ts` source (web and api both point at `./src/*.ts`).. the compiled `_esm` build is only for the npm publish. Our portal API runs as plain compiled JS in Docker, so it actually needs a real `dist` emit through the `compiled` condition, modeled on `ve-hemi-actions`, which the API already consumes the exact same way.

gndelia (approved):

packages/hemi-earn-actions/tsconfig.build.json
gndelia: Ah, the fact you named `ve-hemi-actions` made me remember I made this exact change a few weeks ago for migrating the api to Typescript.

PR #2124, opened 2026-07-23
Add staked amount and receiver index to requests

ArturDolzan:
### Description

This PR extends the `hemi-earn-requests` subgraph so the backend can compute a per-user "Total earned" figure in a follow-up PR.

It adds a `stakedAmount` field to the Request entity, filled from the `staked` param of the Agent's `DepositRequestProcessed` event, and indexes `receiver` so requests can be looked up per share holder. The staked amount is the pegged value actually deposited into the vault — the only quantity comparable to `convertToAssets(shares)` when measuring earnings. The deposit's raw input (`assets`) can't be reused for that because it goes through a gateway
conversion (with slippage) before being staked, and redeems don't need a new field at all since the cost basis is reduced proportionally to the already-indexed shares burned.

Second of the sequence (package → subgraph → api → ui) behind the earned card.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #1965 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/subgraphs/hemi-earn-requests-subgraph/src/mappings/eventHandlers.ts
+++ b/subgraphs/hemi-earn-requests-subgraph/src/mappings/eventHandlers.ts
@@ -41,6 +41,7 @@ const baseRequest = (id: string, requestId: bigint): Request => ({
   requestId,
   requestTxHash: undefined,
   retryCount: 0,
+  stakedAmount: undefined,
   status: 'PENDING',
 })
 
@@ -256,6 +257,7 @@ indexer.onEvent(
         kind: existing.kind ?? 'DEPOSIT',
         processedAt: BigInt(event.block.timestamp),
         processTxHash: event.transaction.hash,
+        stakedAmount: existing.stakedAmount ?? event.params.staked, // pegged staked into the vault
       }),
       requestId: event.params.requestId,
     }),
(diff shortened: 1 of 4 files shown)


gndelia (dismissed):

subgraphs/hemi-earn-requests-subgraph/src/mappings/eventHandlers.ts
```diff
@@ -256,6 +257,7 @@ indexer.onEvent(
         kind: existing.kind ?? 'DEPOSIT',
         processedAt: BigInt(event.block.timestamp),
         processTxHash: event.transaction.hash,
+        stakedAmount: existing.stakedAmount ?? event.params.staked, // pegged staked into the vault
```
gabmontes: `staked` is emitted only by `DepositRequestProcessed`, and this is the only handler that writes `stakedAmount`, so `existing.stakedAmount` is always `undefined` here — the `??` guard can never fire. The sibling single-source fields in this same handler (`processedAt`, `processTxHash`) assign directly without a guard, whereas `??` is reserved for fields written by multiple events (`amountIn`, `amountOut`, `kind`). Suggest simplifying to `stakedAmount: event.params.staked` for consistency.

— Gabi 🤖

ArturDolzan: Good catch — you're right, `stakedAmount` is single-source so the `??` never fires. Simplified to a direct assign, matching `processedAt`/`processTxHash`. [`f4446fa7`](https://github.com/hemilabs/ui-monorepo/pull/2124/commits/f4446fa7)

gabmontes (approved):

gndelia (approved):

PR #2125, opened 2026-07-24
Add Hemi Earn cost-basis API endpoint

ArturDolzan:
### Description

This PR adds the Hemi Earn cost-basis API endpoint and prices peer-to-peer share transfers at fair market value, so "Total earned" reflects what a user actually holds.

It replays a user's processed deposits/redeems keyed by the Hemi share OFT — deposits add their pegged cost basis, redeems reduce it proportionally. On its own that treats a share received via a peer-to-peer transfer as pure profit
(zero cost), inflating earned. So it also consumes the subgraph's ShareTransfer and RateSnapshot signals: transfers are merged into the chronological replay, a transfer in priced at the share->asset rate closest to it (FMV), a transfer out reducing the basis proportionally like a redeem.

Asset resolution is scoped to the caller's own assets and made defensive (a de-registered asset can't 500 the endpoint), and the event timestamp falls back to processedAt when the Hemi request isn't indexed yet.

Third of the sequence (package → subgraph → api → ui). The endpoint returns real data once the subgraph is deployed with the `stakedAmount` field.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #1965 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal-backend/api/src/hemi-earn/costBasis.ts
+++ b/portal-backend/api/src/hemi-earn/costBasis.ts
@@ -48,7 +48,7 @@ const resolveShareByAsset = function (asset: string) {
       // A zero share means the asset isn't registered — fail loudly instead of
       // silently mis-keying the cost basis to the zero address.
       if (resolved === zeroAddress) {
-        throw new Error(`assetsData returned no share for asset ${asset}`)
+        throw new Error(`getAssetData returned a zero share for asset ${asset}`)
       }
       return resolved
     })
@@ -68,8 +68,9 @@ type GetEarnCostBasisQueryResponse = GraphResponse<{ Request: CostBasisRow[] }>
  * Computes a user's per-vault Hemi Earn cost basis by replaying their processed
  * requests. Filtered by `receiver` (the holder of the shares) — unlike
  * `getEarnRequests`, which follows `initiator` — because the earned card values
- * what the user holds. Keyed by the Hemi share OFT (== the portal shareAddress),
- * value in pegged base units (decimal string).
+ * what the user holds. Keyed by the Hemi share OFT (== the portal shareAddress);
+ * the value is a decimal string on the base-unit scale, which may be fractional
+ * after proportional redeems (WAD precision) — so it is not `BigInt`-safe.
  *
  * Known limitation: only Router deposits/redeems are replayed, so peer-to-peer
  * share-OFT transfers are not tracked — shares acquired outside a deposit read
--- a/portal-backend/api/src/hemi-earn/costBasis.ts
+++ b/portal-backend/api/src/hemi-earn/costBasis.ts
@@ -1,0 +1,127 @@
+import { getAssetData } from 'hemi-earn-actions/actions'
+import {
+  type Address,
+  type PublicClient,
+  createPublicClient,
+  formatUnits,
+  http,
+  zeroAddress,
+} from 'viem'
+import { hemi } from 'viem/chains'
+
+import {
+  type GraphResponse,
+  checkGraphQLErrors,
+  paginateHemiEarnSubgraph,
+  requestHemiEarn,
+} from '../subgraphs/subgraph.ts'
+import type { EarnCostBasis } from '../subgraphs/types/earn.ts'
+
+import { type CostBasisRow, WAD_DECIMALS, replayCostBasis } from './replay.ts'
+
+// Lazily created and reused across cost-basis lookups so the Hemi client is only
+// built when the endpoint is hit. Chain-default RPC with batching, matching how
+// the ve-hemi module consumes its package actions.
+let hemiEarnRpcClient: PublicClient | undefined
+const getHemiEarnRpcClient = function () {
+  if (!hemiEarnRpcClient) {
+    hemiEarnRpcClient = createPublicClient({
+      chain: hemi,
+      transport: http(undefined, { batch: true }),
+    })
+  }
+  return hemiEarnRpcClient
+}
+
+// asset -> Hemi share OFT via the Router registry. Memoized because the mapping
+// is immutable and there are only a handful of assets.
+const shareByAssetCache = new Map<string, Promise<Address>>()
+const resolveShareByAsset = function (asset: string) {
+  const key = asset.toLowerCase()
+  const cached = shareByAssetCache.get(key)
+  if (cached) return cached
+  const share = getAssetData(getHemiEarnRpcClient(), {
+    asset: asset as Address,
+  })
+    .then(function (data) {
+      const resolved = data.share.toLowerCase() as Address
+      // A zero share means the asset isn't registered — fail loudly instead of
+      // silently mis-keying the cost basis to the zero address.
+      if (resolved === zeroAddress) {
+        throw new Error(`assetsData returned no share for asset ${asset}`)
+      }
+      return resolved
+    })
+    // Evict on failure so a transient RPC error (or the guard above) can't
+    // poison the cache forever.
+    .catch(function (error) {
+      shareByAssetCache.delete(key)
+      throw error
+    })
+  shareByAssetCache.set(key, share)
+  return share
+}
+
+type GetEarnCostBasisQueryResponse = GraphResponse<{ Request: CostBasisRow[] }>
+
+/**
+ * Computes a user's per-vault Hemi Earn cost basis by replaying their processed
+ * requests. Filtered by `receiver` (the holder of the shares) — unlike
+ * `getEarnRequests`, which follows `initiator` — because the earned card values
+ * what the user holds. Keyed by the Hemi share OFT (== the portal shareAddress),
+ * value in pegged base units (decimal string).
+ *
+ * Known limitation: only Router deposits/redeems are replayed, so peer-to-peer
+ * share-OFT transfers are not tracked — shares acquired outside a deposit read
+ * as pure profit. The normal deposit→receiver flow is fully covered.
+ */
+export const getEarnCostBasis = async function ({
+  address,
+}: {
+  address: Address
+}): Promise<EarnCostBasis> {
+  const receiver = address.toLowerCase()
+
+  const rows = await paginateHemiEarnSubgraph<CostBasisRow>({
+    async fetchPage({ limit, offset }) {
+      const schema = {
+        query: `query GetEarnCostBasis($limit: Int!, $offset: Int!, $receiver: String!) {
+          Request(
+            where: { receiver: { _eq: $receiver }, processedAt: { _is_null: false } }
+            order_by: [{ processedAt: asc }, { requestId: asc }]
+            limit: $limit
+            offset: $offset
+          ) {
+            amountIn
+            amountOut
+            asset
+            kind
+            stakedAmount
+          }
+        }`,
+        variables: { limit, offset, receiver },
+      }
+
+      const response =
+        await requestHemiEarn<GetEarnCostBasisQueryResponse>(schema)
+      checkGraphQLErrors(response)
+      return response.data.Request
+    },
+  })
+
+  const assets = [...new Set(rows.map(row => row.asset.toLowerCase()))]
+  const shareByAsset: Record<string, Address> = Object.fromEntries(
+    await Promise.all(
+      assets.map(async asset => [asset, await resolveShareByAsset(asset)]),
+    ),
+  )
+
+  const positions = replayCostBasis(rows, shareByAsset)
+
+  return Object.fromEntries(
+    [...positions].map(([share, { costBasis }]) => [
+      share,
+      formatUnits(costBasis, WAD_DECIMALS),
+    ]),
+  )
+}
--- a/portal-backend/api/src/hemi-earn/costBasis.ts
+++ b/portal-backend/api/src/hemi-earn/costBasis.ts
@@ -17,7 +17,12 @@ import {
 } from '../subgraphs/subgraph.ts'
 import type { EarnCostBasis } from '../subgraphs/types/earn.ts'
 
-import { type CostBasisRow, WAD_DECIMALS, replayCostBasis } from './replay.ts'
+import {
+  type CostBasisEvent,
+  type RatePoint,
+  WAD_DECIMALS,
+  replayCostBasis,
+} from './replay.ts'
 
 let hemiEarnRpcClient: PublicClient | undefined
 const getHemiEarnRpcClient = function () {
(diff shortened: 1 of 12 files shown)


portal-backend/api/src/hemi-earn/replay.ts
```diff
@@ -0,0 +1,76 @@
...
+  row: CostBasisRow,
+): Position {
+  const staked = row.stakedAmount ?? null
+  const minted = row.amountOut ?? null
+  if (staked === null || minted === null) return position
+  return {
+    costBasis: position.costBasis + BigInt(staked) * WAD,
+    shares: position.shares + BigInt(minted),
+  }
+}
```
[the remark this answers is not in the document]
ArturDolzan: Good catch — `applyDeposit` now always tracks the minted shares and only skips the cost basis when `stakedAmount` is null, so a later redeem divides by the true share balance. Done in [`bd3c769f`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/bd3c769f).

portal-backend/api/test/hemi-earn/replay.test.ts
```diff
@@ -0,0 +1,122 @@
...
+          amountOut: '50',
+          asset: ASSET_A1,
+          kind: 'DEPOSIT' as const,
+          stakedAmount: null,
+        },
+      ],
+      shareByAsset,
+    )
+    expect(positions.get(SHARE_A)).toEqual({ costBasis: 0n, shares: 0n })
+  })
```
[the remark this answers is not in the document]
ArturDolzan: Updated the test to assert shares are still tracked, and added a mixed known/unknown-deposit + redeem case to lock in the proportional reduction against the full balance. [`bd3c769f`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/bd3c769f)

portal-backend/api/src/subgraphs/types/earn.ts
```diff
@@ -50,6 +50,14 @@ export type SubgraphRequest = EarnRequestCommonFields & {
   status: SubgraphRequestStatus
 }
 
+// Per-vault cost basis for a user, as returned by
+// `GET /subgraphs/:chainId/earn-cost-basis/:address`. Keyed by the Hemi share
+// OFT (lowercased, == the portal's `EarnPosition.shareAddress`); the value is
+// the pegged-token cost basis in base units (decimal string, WAD precision
+// preserved). The portal subtracts it from `convertToAssets(shares)` to derive
+// "Total earned".
+export type EarnCostBasis = Record<Address, string>
```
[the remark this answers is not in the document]
ArturDolzan: Fair — clarified both the type comment and the `costBasis` docstring that it's a decimal string that can be fractional (WAD precision), so not `BigInt`-safe. [`bd3c769f`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/bd3c769f)

portal-backend/api/src/hemi-earn/costBasis.ts
```diff
@@ -0,0 +1,127 @@
...
+  const share = getAssetData(getHemiEarnRpcClient(), {
+    asset: asset as Address,
+  })
+    .then(function (data) {
+      const resolved = data.share.toLowerCase() as Address
+      // A zero share means the asset isn't registered — fail loudly instead of
+      // silently mis-keying the cost basis to the zero address.
+      if (resolved === zeroAddress) {
+        throw new Error(`assetsData returned no share for asset ${asset}`)
+      }
```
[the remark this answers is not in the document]
ArturDolzan: Done — swapped to `getAssetData` for clearer log tracing (that's the wrapper; `assetsData` is the underlying Router method it calls). Heads up, it's the only occurrence — line 68 is just the docstring, no second error string. [`bd3c769f`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/bd3c769f)

portal-backend/api/src/hemi-earn/costBasis.ts
```diff
@@ -0,0 +1,128 @@
...
+  paginateHemiEarnSubgraph,
+  requestHemiEarn,
+} from '../subgraphs/subgraph.ts'
+import type { EarnCostBasis } from '../subgraphs/types/earn.ts'
+
+import { type CostBasisRow, WAD_DECIMALS, replayCostBasis } from './replay.ts'
+
+// Lazily created and reused across cost-basis lookups so the Hemi client is only
+// built when the endpoint is hit. Chain-default RPC with batching, matching how
+// the ve-hemi module consumes its package actions.
```
gndelia: LEt's remove unnecessary comments
```suggestion
```

portal-backend/api/src/hemi-earn/costBasis.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Computes a user's per-vault Hemi Earn cost basis by replaying their processed
+ * requests. Filtered by `receiver` (the holder of the shares) — unlike
+ * `getEarnRequests`, which follows `initiator` — because the earned card values
+ * what the user holds. Keyed by the Hemi share OFT (== the portal shareAddress);
+ * the value is a decimal string on the base-unit scale, which may be fractional
+ * after proportional redeems (WAD precision) — so it is not `BigInt`-safe.
+ *
+ * Known limitation: only Router deposits/redeems are replayed, so peer-to-peer
+ * share-OFT transfers are not tracked — shares acquired outside a deposit read
+ * as pure profit. The normal deposit→receiver flow is fully covered.
```
gndelia: > shares acquired outside a deposit read as pure profit.

I think we don't do that for Vetro; IIRC, we consider the cost at the transfer moment. It may be worth checking (cc @gabmontes  as he implemented that)

portal-backend/api/src/hemi-earn/replay.ts
```diff
@@ -0,0 +1,83 @@
...
+  asset: string
+  kind: 'DEPOSIT' | 'REDEEM'
+  stakedAmount: string | null
+}
+
+type Position = { costBasis: bigint; shares: bigint }
+
+// Fixed precision factor, NOT any token's decimals: the cost basis is scaled up
+// by WAD so the integer proportional-redeem division keeps sub-unit precision.
+// Callers divide back out by the same WAD_DECIMALS.
```
gndelia: probably not necessary
```suggestion

```

portal-backend/api/src/subgraphs/router.ts
```diff
@@ -161,6 +162,13 @@ function parseQueryParams(
 const getStaked = (chainIdStr: string) =>
   getTotalStaked(Number(chainIdStr)).then(staked => ({ staked }))
 
+// `_chainIdStr` is unused but must stay the first param — toJsonMiddleware passes
+// the route params positionally, so `address` needs to land as the second arg.
```
gndelia: Not necessary, that's how function parameter ordering works

```suggestion
```

Btw sometimes you can name the "unused" parameter as `_` (if It's just one)

portal-backend/api/src/subgraphs/types/earn.ts
```diff
@@ -50,6 +50,15 @@ export type SubgraphRequest = EarnRequestCommonFields & {
   status: SubgraphRequestStatus
 }
 
+// Per-vault cost basis for a user, as returned by
+// `GET /subgraphs/:chainId/earn-cost-basis/:address`. Keyed by the Hemi share
+// OFT (lowercased, == the portal's `EarnPosition.shareAddress`); the value is
+// the pegged-token cost basis on the base-unit scale, as a decimal string. It
+// may carry a fractional part from WAD-precision proportional redeems, so it is
+// not necessarily an integer — don't parse it as `BigInt`. The portal subtracts
+// it from `convertToAssets(shares)` to derive "Total earned".
```
gndelia:
```suggestion
```


Probably not needed, or at least it could be widely simplified

portal-backend/api/src/hemi-earn/costBasis.ts
ArturDolzan: Done — dropped the comment. [`ffd973ac`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/ffd973ac)

portal-backend/api/src/hemi-earn/replay.ts
ArturDolzan: Trimmed it to a one-liner — kept a short "not a token's decimals" note since that tripped someone up, but dropped the rest. Shout if you'd rather remove it entirely. [`ffd973ac`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/ffd973ac)

portal-backend/api/src/subgraphs/router.ts
ArturDolzan: Good call — dropped the comment and renamed it to `_`. [`ffd973ac`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/ffd973ac)

portal-backend/api/src/subgraphs/types/earn.ts
ArturDolzan: Simplified it — kept just the key bit (decimal string, not `BigInt`-safe, which Copilot had flagged) and dropped the rest. [`ffd973ac`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/ffd973ac)

portal-backend/api/src/hemi-earn/costBasis.ts
ArturDolzan: You're correct.. checked Vetro's `handleTransfer` and the receiver does get cost basis at FMV (`convertToAssets` at transfer time), not zero. Ours only reads as pure profit because this subgraph indexes the Router deposits/redeems, not the share-OFT transfers, so there's no transfer data to price it. @gabmontes what do you think is best here?

gabmontes: @ArturDolzan pricing the tokens as pure profit did not work for Vetro and is not the most common pricing mechanism in use by other DeFi protocols. How can we get a a FMV in this case?

ArturDolzan: @gabmontes got it, so no pure profit. The tricky part here is it's cross-chain: the share is an OFT on Hemi but `convertToAssets` lives on the Ethereum vault, so we can't price a transfer inline like Vetro does on a single chain.

What do you think on indexing the share transfers and price each at the share -> asset rate nearest that moment, which we can pull from the `staked/shares` snapshot every deposit/redeem already emits (no archive reads needed). It's approximate but close. Would that work for you?

gabmontes: That could be a good first approach. Let's go with that and improve later as needed.

ArturDolzan: Done — went with the FMV approach in [`17595aef`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/17595aef). The subgraph now emits `ShareTransfer` + a share->asset `RateSnapshot` timeline (that's how we get around the cross-chain `convertToAssets`), and the replay prices each received transfer at the rate closest to it — transfers out reduce the basis proportionally, like a redeem. Verified with unit tests + end-to-end against a local Envio + anvil fork: a wallet that only received a transfer now shows a real FMV cost basis instead of zero.

portal-backend/api/src/hemi-earn/costBasis.ts
```diff
@@ -0,0 +1,337 @@
...
+      return response.data.ShareTransfer
+    },
+  })
+
+const getEarnAssets = async function () {
+  const schema = {
+    query: `query GetEarnAssets {
+      RateSnapshot(distinct_on: asset) { asset }
+    }`,
+  }
```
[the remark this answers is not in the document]
ArturDolzan: Good call — added the explicit `order_by: { asset: asc }` in [`41e2d886`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/41e2d886). FWIW `distinct_on` alone already works on our Envio indexer (verified against prod + a local instance), but this is the canonical form and keeps it portable.

portal-backend/api/src/subgraphs/router.ts
```diff
@@ -274,5 +280,17 @@ export function createSubgraphsRouter() {
     getEarnRequestsHandler,
   )
 
+  router.get(
+    '/:chainIdStr/earn-cost-basis/:address',
+    parseChainId,
+    validateAddress,
+    validateChainIsHemiMainnet,
+    toJsonMiddleware(getCostBasis, {
+      maxAge: 30 * 1000,
```
gndelia, line 289: Careful here (bc I've seen the same issue on Earn): If you cache this, then the invalidations on the client side won't work. So in that case you may want to just update optimistically on the client side, and skip the invalidation.

portal-backend/api/src/subgraphs/router.ts
```diff
@@ -161,6 +162,11 @@ function parseQueryParams(
 const getStaked = (chainIdStr: string) =>
   getTotalStaked(Number(chainIdStr)).then(staked => ({ staked }))
 
+const getCostBasis = (_: string, address: string) =>
```
gndelia: Nit: Doesn't work if you type here directly

```suggestion
const getCostBasis = (_: string, address: Address) =>
```

Skipping the cast?

portal-backend/api/src/hemi-earn/replay.ts
```diff
@@ -0,0 +1,86 @@
...
+export const replayCostBasis = function (events: CostBasisEvent[]) {
+  const positions = new Map<Address, Position>()
+  events.forEach(function (event) {
+    const share = event.share as Address
+    positions.set(
+      share,
+      applyEvent(positions.get(share) ?? emptyPosition, event),
+    )
+  })
+  return positions
```
gndelia, line 85: If `applyEvent` calls `dispose` on some positions, this means that positions may end up with zero tokens. Do we need to return these, or is it better to filter them out?

portal-backend/api/src/subgraphs/router.ts
ArturDolzan: Fair concern! Optimistic update is tricky here since the API owns the whole cost-basis replay (subgraph + on-chain), so the client can't predict the new value after an action. But I don't think it needs to: the cost basis only changes once the subgraph indexes the deposit/redeem/transfer, and Hemi's RPC-sync indexing lag already dominates the 30s cache — so the card can't be instant-fresh regardless, and the client picks it up by polling the query (which the cache just batches). If 30s feels like too much on top of that lag, happy to drop it to ~10s to match the poll cadence; removing it entirely wouldn't fix the underlying lag, just add load. If you'd rather rethink the invalidation story, glad to open an issue to dig in calmly.

portal-backend/api/src/subgraphs/router.ts
ArturDolzan: Done — typed the param as `Address` and dropped the cast in [`0111f1d9`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/0111f1d9).

portal-backend/api/src/hemi-earn/replay.ts
ArturDolzan: Good call — now filtering out fully-exited (zero-share) positions in [`0111f1d9`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/0111f1d9), so the response only carries shares the holder still has.

portal-backend/api/src/hemi-earn/costBasis.ts
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// asset -> Hemi share OFT via the Router registry. Memoized because the mapping
+// is immutable and there are only a handful of assets.
+const shareByAssetCache = new Map<string, Promise<Address>>()
+const resolveShareByAsset = function (asset: string) {
+  const key = asset.toLowerCase()
+  const cached = shareByAssetCache.get(key)
+  if (cached) return cached
+  const share = getAssetData(getHemiEarnRpcClient(), {
```
gndelia: I'd make this a bit more clearer because it took me a while to figure out what's being done.

```suggestion
  const sharePromise = getAssetData(getHemiEarnRpcClient(), {
```

portal-backend/api/src/hemi-earn/costBasis.ts
```diff
@@ -0,0 +1,337 @@
...
+  // pricing, so skip that enumeration (and its RPC calls) when there are no
+  // transfers. Resolution is defensive: one de-registered asset must not fail the
+  // whole request.
+  const requestAssets = requestRows.map(row => row.asset.toLowerCase())
+  const globalAssets = transferRows.length > 0 ? await getEarnAssets() : []
+  const resolved = await Promise.all(
+    [...new Set([...requestAssets, ...globalAssets])].map(asset =>
+      resolveShareByAsset(asset).then(
+        share => [asset, share] as const,
+        () => null,
```
gndelia: It's unusual that here you did not use a `.catch()` instead

portal-backend/api/src/hemi-earn/costBasis.ts
ArturDolzan: Good catch — renamed it to `sharePromise` so it's clear the cached value is a promise, in [`e61bd924`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/e61bd924f10839d5f33ea53b83c690477caceaeb).

portal-backend/api/src/hemi-earn/costBasis.ts
ArturDolzan: Fair, reads better that way — switched to `.catch()` in [`e61bd924`](https://github.com/hemilabs/ui-monorepo/pull/2125/commits/e61bd924f10839d5f33ea53b83c690477caceaeb).

gndelia (approved):
uACK

gabmontes (approved):
uACK

PR #2129, opened 2026-07-24
Redesign staking table and unify table chrome

### LINKED ISSUE #1629: Improve the table UI implementation to match the designs in Figma.
It seems that in the implementation of the new table designs, a small padding is being applied to the table. We also need to check and add the overflow-hidden property so the shadows don’t get clipped.

Implementation:

Image
Image

Designs:

Image

### LINKED ISSUE #1998: Improve Governance Staking table width on desktop viewports
We need to iterate the table spacing so that all tabla data items fit the table width (without scrolling) on desktop views.

I suggest:

- Iterate the `<td>` width for the lockup label.
- Iterate the 'time remaining' element width to 80px.

### LINKED ISSUE #2068: Revamp table component
We need to iterate the table component to better match the updated one in figma.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=20328-3651&t=cRKBfQOjHKZkTSEe-1)

### LINKED ISSUE #2070: Revamp the governance staking table UX/UI
We need to update the Governance staking table UI to match the one designed on figma

Designs here

### LINKED ISSUE #2075: Iterate governance staking table for the new updated one
We need to update the current governance staking implemented on production, for the iterated one.

Some of the changes are the following:

- Labels typography styles for the secondary texts where updated fo the `C1`
- Rewards now display the token logos of the rewards and on hover a tooltip appears with the different amounts.
- Now the 'Time Remaining' changed to 'Action' in which we added a badge inside the CTA with the time remaining for withdrawing the position.

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=20346-8921&t=cRKBfQOjHKZkTSEe-1)

arielmergen:
### Description

Redesign the governance staking table per Figma and unify the shared
table chrome so the rest of the tables follow it.

- Governance table: rewards as overlapping token logos + "N Available"
  with an amounts tooltip; the time-remaining column is merged into
  Action (locked rows show a disabled Unlock with a time-remaining
  badge); secondary text uses the caption style; empty state is a clean
  card with no header.
- Shared Table: `h-10` header row, 42px tall as rendered — the `<th>`
  adds 1px of UA padding top and bottom, which matches the 42px band in
  Figma (caption weight/color, 8px top radius, 1px
  border), body card at 8px radius with the `sm` shadow, full-width
  (dropped the inner padding/overflow that clipped the shadow), and a
  reusable `TableCard` for the empty/loading/connect states.
- Call sites updated to the new chrome: Stake (dashboard + strategy) and
  Hemi Earn (transactions + composition). Get Started already opted out
  of the inner padding, so it picks up the new chrome from the shared
  component alone.
- Hemi Earn's composition loading skeleton hardcoded the previous radius
  and shadow, and rendered the header and body flush; it now mirrors the
  shared chrome, including the header/body overlap (`pb-1.5` + `-mt-1.5`),
  the body's `mb-1`, and the 42px header band it was rendering as 40px.
- The rewards cell now has loading and error states. It used to render
  the count straight from `rewardTokens.length`, so an empty list showed
  "0 Available" with an empty tooltip instead of a loading state. It now
  follows the same shape as the neighbouring cells: `-` on error, a
  skeleton until the tokens resolve.

#### Review feedback

- Tunnel → transaction history now uses the shared `Table`. Its own
  `table.tsx`, `transactionHistory.css` and `tunnelHistorySyncStatus.tsx`
  are gone (the sync state lives in `reloadHistory.tsx`), and the columns
  moved to a `columns.tsx` next to the other tables.
- Cell gutters: `first:[&>*]:pl-4 last:[&>*]:pr-4` compiled to
  `> :first-child` / `> :last-child`, so the variant landed on the cell's
  child instead of the first/last cell in the row. Since every cell has a
  single child, all of them carried 16px per side — 160px across five
  columns where the design asks for 32. It is now `first:pl-4 last:pr-4`
  on the `<th>`/`<td>`.
- Column widths recalculated per table against the longest string in
  en/es/pt. Governance 820 → 745, tunnel 1192 → 950. Two columns are not
  driven by English: the tunnel's `Tx Hash` header is widest in Portuguese
  (`Hash da Transação`) and governance's `Action` in Spanish
  (`Dentro de 10 meses`, 100px against 66.5px for `In 10 months`).
- The table no longer drops its minimum width when it fits its container:
  below that floor the columns used to shrink past their content and the
  text overlapped, so it scrolls instead.
- Three dots stay active while the dropdown is open, the external-link
  arrow is gone from the tx hash, and on mobile the Action column leads
  the row left-aligned with the dots before the CTA.
- Boost staking: Protocol / Asset / Staked no longer stretch, the USD
  sublabel uses the caption style, and the Stake tab right-aligns Action.
- Hemi Earn transactions got the missing `Action` header (en/es/pt).
- Full-height states: the `connecting` skeleton sat 3px low because
  `react-loading-skeleton` renders inline-flex inside a wrapper span, so
  it aligned on the baseline; it is block now. The tunnel's `dynamic()`
  placeholder also still used the old chrome (`h-80 md:h-[500px]`,
  `rounded-2xl`) and now matches the table it stands in for.
- Scrollbar detection compares `offsetWidth` with `clientWidth` instead of
  `scrollHeight` with `clientHeight`: with overlay scrollbars (macOS) the
  body keeps its full width, so padding the header pushed it out of
  alignment with the body.

### Screenshots
01-governance-populated-1512-sin-scroll

tunnel-02-reload-skeleton

tunnel-03-es-1280-sin-scroll

tunnel-04-pt-1280-sin-scroll

boost-01-stake-tab-action-derecha

boost-02-dashboard-columnas-juntas-caption

boost-03-mobile-action-izquierda

earn-01-transactions-label-action

earn-02-composition-por-token

earn-03-composition-por-protocolo

gov-02-tres-puntos-activo

gov-03-hash-sin-flecha-externa

gov-04-tab-burned

gov-05-mobile-action-primero-izquierda

gov-06-empty-no-hemi-staked

gov-07-es-sin-solapamiento

gov-08-pt-sin-solapamiento

tabla-01-get-started

tabla-02-loading-skeleton-por-celda

tabla-03-loading-stake-dashboard

tunnel-01-tx-history

### Related issue(s)

Closes #2075
Closes #2068
Closes #2070
Closes #1998
Closes #1629

### Checklist

- [x] Manual testing passed.


portal/messages/en.json
```diff
@@ -582,7 +582,8 @@
       "no-hemi-staked": "No ${symbol} staked",
       "on": "On {date}",
       "position-delegated-to-you": "Position delegated to you",
-      "time-remaining": "Time remaining",
+      "rewards": "Rewards",
+      "rewards-available": "{count} Available",
```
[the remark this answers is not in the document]
arielmergen: The Figma design linked in #2075 (node `20346-8921`) renders this label as "2 Available" with a capital A, so the English string matches the spec. The lowercase form in the PR description was informal prose, not the source of truth — I updated the description to match. `es`/`pt` stay lowercase (`disponibles` / `disponíveis`) since those languages do not capitalize mid-sentence.

portal/stories/table.stories.tsx
```diff
@@ -0,0 +1,252 @@
...
+    <div className="flex -space-x-1">
+      {Array.from({ length: count }).map((_, index) => (
+        <div className="rounded-full ring-2 ring-white" key={index}>
+          <HemiToken className="size-4" />
+        </div>
+      ))}
+    </div>
+    <span className="body-text-caption text-neutral-500">
+      {`${count} Available`}
+    </span>
```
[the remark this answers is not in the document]
arielmergen: Same as the thread on `en.json`: the Figma design (node `20346-8921` in #2075) shows "2 Available". The story mirrors the real usage in the app, so the capital A is intentional.

portal/messages/es.json
```diff
@@ -582,7 +582,8 @@
       "no-hemi-staked": "No hay ${symbol} en stake",
       "on": "En {date}",
       "position-delegated-to-you": "Posición delegada hacia usted",
-      "time-remaining": "Tiempo restante",
+      "rewards": "Recompensas",
+      "rewards-available": "{count} disponibles",
```
gndelia: Both in spanish and portuguese, pluralization needs to be handled here. Otherwise, 1 would read `1 disponibles` instead of `1 disponible`

portal/app/[locale]/staking-dashboard/_components/stakeTable/linearProgress.tsx
```diff
@@ -1,24 +0,0 @@
...
-  unlockTime: number
-}
-
-export function LinearProgress({ percentage, unlockTime }: Props) {
-  const width = `${Math.min(100, Math.max(0, percentage))}%`
-
-  return (
-    <div className="relative flex h-7 min-w-28 items-center justify-center overflow-hidden rounded-md bg-neutral-100 px-2 py-1.5">
-      <div
-        className="absolute inset-y-0 left-0 bg-linear-progress-bar transition-all duration-300 ease-out"
```
Comment (gndelia) [inline portal/app/[locale]/staking-dashboard/_components/stakeTable/linearProgress.tsx:14]: I think `linear-progress-bar` is no longer used anywhere, so we can remove it from Tailwind's theme

portal/messages/es.json
arielmergen: Fixed in 741c0f87. Both locales now use ICU plural, so `count: 1` renders `1 disponible` / `1 disponível`. English keeps `{count} Available`, which reads correctly for any count.

portal/app/[locale]/staking-dashboard/_components/stakeTable/linearProgress.tsx
Comment (arielmergen) [reply portal/app/[locale]/staking-dashboard/_components/stakeTable/linearProgress.tsx:14]: Removed in 20e06bcc. Confirmed `linear-progress-bar` had no other references left once this component was deleted.

arielmergen:
Addressing the suppressed comment on `rewardsDisplay.tsx` (loading state showing "0 Available" with an empty tooltip) — fixed in e6c2d1c4.

Confirmed it by rendering the component in isolation against the real hooks: with an empty token list the readout is `{ hasError: false, isLoading: false, tokenCount: 0 }` and the cell renders "0 Available", with the tooltip opening as an empty overlay on hover.

Worth noting for anyone reading this later: `isLoading` is `false`, not `true`. `useRewardTokensAddresses` is `enabled: !!hemiWalletClient`, and a disabled query in react-query v5 is `isPending` but not `isFetching`, so `isLoading` never flips. Guarding on `isLoading` would not have covered it — the guard is on `tokens.length` instead.

Also added an error branch, since `hasError` was being discarded the same way. Both follow the existing shape in this table: `-` on error, skeleton until the data resolves (`votingPower.tsx:25`, `lockupTime.tsx:33`).

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/composition/compositionTable.tsx
```diff
@@ -33,9 +33,9 @@ export const CompositionTable = function ({
 
   if (isPending) {
     return (
-      <div className="w-full overflow-hidden rounded-xl bg-neutral-100 text-sm font-medium">
-        <div className="h-10" />
-        <div className="rounded-xl bg-white shadow-md">
+      <div className="w-full text-sm font-medium">
+        <div className="h-10 rounded-t-lg bg-neutral-100 shadow-bs" />
+        <div className="overflow-hidden rounded-lg bg-white shadow-sm">
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/composition/compositionTable.tsx]: The loading skeleton renders the header and body flush, without the shared Table's overlap (`pb-1.5` on the header + `-mt-1.5` on the body). When the real `<Table/>` mounts you can get a small corner/shadow seam jump. Consider mirroring the overlap so the transition is seamless. — Gabi 🤖

Comment (arielmergen) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/composition/compositionTable.tsx]: Mirrored in 43e2a720 — the skeleton header now carries `pb-1.5` and the body `-mt-1.5`, matching the shared Table.

Measuring both side by side turned up two more gaps beyond the overlap: the skeleton was missing the body's `mb-1`, and its header was `h-10` (40px) where the rendered one is 42px — the `<th>` picks up 1px of UA padding top and bottom that `ColumnHeader` does not reset. That 2px was the one thing actually shifting content on mount.

On the seam itself: the body lands in the same place either way, since `pb-1.5` and `-mt-1.5` cancel out. What the overlap changes is the strip of `bg-neutral-100` sitting behind the body's rounded top corners, which the skeleton was not painting.

Both now measure identically — header 48px total, 42px visible band, 6px overlap, 4px bottom margin.

gabmontes (dismissed):

gndelia (dismissed):

portal/stories/table.stories.tsx
```diff
@@ -0,0 +1,252 @@
...
+      d="M11.2016 3.7519C11.1372 3.74034 11.0767 3.78272 11.0653 3.84823L10.2895 8.3296H9.71049L8.93468 3.84823C8.92333 3.78272 8.86277 3.74034 8.79844 3.7519C6.04337 4.29521 3.93165 6.68425 3.75757 9.60504C3.75757 9.60889 3.75 9.7322 3.75 9.79385C3.75 9.80156 3.75 9.80926 3.75 9.81312C3.75 9.83624 3.75 9.85935 3.75 9.88247C3.75 9.89018 3.75 9.89789 3.75 9.90945C3.75 9.94027 3.75 9.96725 3.75 9.99807C3.75 13.0961 5.92227 15.674 8.80222 16.2442C8.86656 16.2558 8.92711 16.2134 8.93846 16.1479L9.71427 11.6665H10.2933L11.0653 16.1518C11.0767 16.2173 11.1372 16.2597 11.2016 16.2481C13.9566 15.7009 16.0646 13.3119 16.2424 10.3911C16.2424 10.3873 16.25 10.2639 16.25 10.2023C16.25 10.1946 16.25 10.1869 16.25 10.183C16.25 10.1599 16.25 10.1368 16.25 10.1137C16.25 10.106 16.25 10.0983 16.25 10.0867C16.25 10.0559 16.25 10.0289 16.25 9.99807C16.2538 6.90003 14.0815 4.32218 11.2016 3.7519Z"
+      fill="white"
+    />
+  </svg>
+)
+
+const TokenIcon = () => (
+  <div className="flex h-9 w-9 shrink-0 items-center justify-center rounded-lg bg-neutral-50 p-2">
+    <HemiToken className="size-5" />
+  </div>
```
ArturDolzan, line 37: This could be simplified:

```suggestion
  <div className="flex size-9 shrink-0 items-center justify-center rounded-lg bg-neutral-50 p-2">
    <HemiToken className="size-5" />
  </div>
```

ArturDolzan (commented):
I left a minor comment otherwise LGTM.

dvnahuel:
Here's some points of improvements:

1. We need to upadte the Tunnel -> TX history table

dvnahuel:
**Governance staking**

1. Governance staking table rows needs to fit the table width (at least on desktop views)
Screenshot 2026-08-06 at 11 15 35 AM

2. Three dots CTA should be active while the dropdown container is live.
Screenshot 2026-08-06 at 11 16 58 AM

3. Let's remove the external url arrow interaction
Screenshot 2026-08-06 at 11 18 16 AM

4. On mobile, the action label and values should be left align and in terms of orders, the three dots should be first and then the main action cta (on mobile only)
Screenshot 2026-08-06 at 11 20 28 AM

dvnahuel:
**Boost staking**

1. Protocol - Asset - Staked labels and values should be closer to each other and the action should stay as it is right now
Screenshot 2026-08-06 at 11 20 03 AM

2. On mobile, the action value and label should be left-aligned
Screenshot 2026-08-06 at 11 22 51 AM

3. Sublabel does't seem to be following the same pattern set in governance staking, this sub labels should be 'caption'
Screenshot 2026-08-06 at 11 31 12 AM

4. On the 'stake' tab, the action label and value both should be right-justified
Screenshot 2026-08-06 at 11 33 41 AM

dvnahuel:
**Hemi Earn**

1. 'Action' label is missing
Screenshot 2026-08-06 at 11 26 04 AM

arielmergen:
All the points are in, pushed on top of the branch. Screenshots for each one are in the PR description.

**Tunnel → transaction history.** Migrated to the shared `Table`. Its own `table.tsx`, `transactionHistory.css` and `tunnelHistorySyncStatus.tsx` are gone — the sync state now lives in `reloadHistory.tsx` and the columns moved to a `columns.tsx` next to the other tables.

**Governance staking**

1. Rows fill the table width. Column widths were recalculated per table against the longest string in en/es/pt: governance went from 820px to 745px, tunnel from 1192px to 950px. Two columns are not driven by English — the tunnel's `Tx Hash` header is widest in Portuguese (`Hash da Transação`) and governance's `Action` in Spanish (`Dentro de 10 meses`, 100px against 66.5px for `In 10 months`). While measuring this I also found the cell gutters were wrong: a Tailwind variant landed on the cell's child instead of the first/last cell in the row, so every cell carried 16px per side — 160px across five columns where the design asks for 32.
2. The three-dot CTA stays active while its dropdown is open.
3. The external-link arrow is gone from the tx hash.
4. On mobile the Action column leads the row, left-aligned, with the dots before the main CTA.

**Boost staking**

1. Protocol / Asset / Staked no longer stretch, so the labels and values sit together. Action is unchanged.
2. On mobile the Action column is left-aligned.
3. The USD sublabel uses the caption style, same as governance.
4. On the Stake tab, Action is right-aligned.

**Hemi Earn**

1. The `Action` header is there now, in the three locales.

Two notes on the mobile screenshots: they were taken at 606px, which is the narrowest this browser window goes here, so they show the column order and alignment rather than a real phone viewport. And the table on Hemi Earn needed a fixture to render at all — the share registry is unreachable from this machine — so the rows there are mock data, not production values.

arielmergen:
Going through the four suppressed comments.

**`rewardsDisplay.tsx` — both are real, fixed.**

`useRewardTokens` defaulted the addresses query to `[]` and built `hasError` only from the token metadata queries, so a failure of the addresses query read as "a share with no reward tokens" and the cell showed a skeleton forever. And using `rewardTokens.length === 0` as the loading condition meant a settled-empty response stayed a skeleton too.

The suggested fix — branching on `isLoading` — does not hold here, for the same reason the guard was on `tokens.length` in the first place: `useRewardTokensAddresses` is `enabled: !!hemiWalletClient`, and a disabled query in react-query v5 is pending but not fetching, so `isLoading` never flips. The cut is on `status` instead, which separates pending from success even while the query is disabled. That logic moved to `_utils/rewardTokensStatus.ts` with unit tests over the matrix, and the cell now renders three explicit states: `-` on error, a skeleton while pending, `-` when it settles with no tokens.

**`columns.tsx` — both are valid, both predate this PR.**

The sortable header and the filter trigger are `span`s with `onClick`, so they are mouse-only. That is unchanged behaviour: the same markup was in `table.tsx` before the migration (the sort header at line 147, the type filter at line 175) and moved across verbatim. Turning them into buttons also means reworking the dropdown's `triggerRef`, so it is a change on its own rather than something to fold into this diff. Worth a follow-up issue for the table's keyboard accessibility.

portal/stories/table.stories.tsx
arielmergen: Applied — `TokenIcon` uses `size-9` now. Thanks.

gabmontes (dismissed):

dvnahuel:
Minor thing here, we should iterate the alignment
Screenshot 2026-08-11 at 5 48 28 PM

arielmergen:
Hi @dvnahuel, I've iterated on the alignment as requested

Captura de pantalla 2026-08-12 a la(s) 9 24 17 a  m

dvnahuel (approved):
LGTM!

gndelia (dismissed):

portal/app/[locale]/staking-dashboard/_components/rewardsDisplay.tsx
```diff
@@ -20,20 +30,52 @@ function RewardRow({ token, tokenId }: { token: EvmToken; tokenId: bigint }) {
...
+    return <span className="text-sm text-neutral-950">-</span>
+  }
 
   return (
-    <div className="space-y-1">
-      {rewardTokens.map(item => (
-        <RewardRow key={item.address} token={item} tokenId={tokenId} />
-      ))}
-    </div>
+    <Tooltip
```
Comment (ArturDolzan) [inline portal/app/[locale]/staking-dashboard/_components/rewardsDisplay.tsx:57]: Perhaps it's worth checking whether the rewards queries still run now that `RewardAmount` lives inside the tooltip overlay — `useHasRewards` reads those amounts from the cache, and it looks like it may gate the "Claim rewards" item.


Artur's 🤖

portal/stories/table.stories.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+// Mirrors the dashboard's loading state: a full-height skeleton fills the card
+// while positions are fetched.
+export const Loading: Story = {
+  parameters: {
+    controls: { disable: true },
+  },
+  render: () => (
+    <TableCard>
+      <Skeleton className="h-[calc(100%-3px)] w-full rounded-lg" />
```
ArturDolzan: Might be worth a quick look at the `Loading` story — without `containerClassName` the percentage height may not resolve, so it might not fill the card the way `stakeTable` now does.


Artur's 🤖

portal/components/txLink.tsx
```diff
@@ -24,7 +23,7 @@ export const TxLink = function ({
   const hash = `${txHash.slice(0, 6)}...${txHash.slice(-4)}`
   const href = `${chain?.blockExplorers?.default.url}/tx/${txHash}`
   return (
-    <div className="group/txhash-link flex w-full items-center gap-x-2">
+    <div className="flex w-full items-center gap-x-2">
```
ArturDolzan: Not sure, but maybe `gap-x-2` can go here now that the arrow icon is gone and the flex container has a single child?


Artur's 🤖

portal/app/[locale]/staking-dashboard/_components/stakeTable/index.tsx
```diff
@@ -141,51 +130,54 @@ export function StakeTable({ data, filter = 'active', loading }: Props) {
...
     }
 
     if (status === 'connecting') {
       return (
-        <Container>
-          <Skeleton className="h-[calc(100%-3px)] w-full rounded-xl" />
-        </Container>
+        <TableCard>
+          <Skeleton
+            className="block h-full w-full rounded-lg"
```
Comment (ArturDolzan) [inline portal/app/[locale]/staking-dashboard/_components/stakeTable/index.tsx]: Wonder if `size-full` would read a bit cleaner than `h-full w-full` in these two skeletons — we already use it in `tableBodyContainer` and the new story.


Artur's 🤖

portal/app/[locale]/staking-dashboard/_components/rewardsDisplay.tsx
Comment (arielmergen) [reply portal/app/[locale]/staking-dashboard/_components/rewardsDisplay.tsx:57]: Confirmed — they were not running. The tooltip overlay only mounts when it opens (rc-tooltip with `destroyTooltipOnHide`), so `useCalculateRewards` never ran, and `useHasRewards` reads exactly that cache — "Claim rewards" stayed disabled until you hovered the cell. It used to work because the amounts were rendered inline in the cell.

`useHasRewards` now runs the queries itself with `useQueries` over shared query options, so it no longer depends on another component being mounted. Same query key, so the tooltip adds no extra request.

Fixed in 9431b581.

portal/stories/table.stories.tsx
arielmergen: Right. It also used `h-[calc(100%-3px)]`, which is never generated — the `content` globs in `tailwind.config.ts` cover `components/`, `app/` and `pages/`, not `stories/`, so the skeleton had no height at all. It now mirrors `stakeTable`: `block size-full` plus `containerClassName="block h-full"`.

Fixed in aa3b3241.

portal/components/txLink.tsx
arielmergen: Yes — the gap was only separating the hash from the arrow icon. Removed in 0a65ec2b.

portal/app/[locale]/staking-dashboard/_components/stakeTable/index.tsx
Comment (arielmergen) [reply portal/app/[locale]/staking-dashboard/_components/stakeTable/index.tsx]: Done in 23e7cfad — also in `transactionHistory.tsx`, which had the same skeleton.

portal/app/[locale]/tunnel/transaction-history/_components/columns.tsx
```diff
@@ -0,0 +1,340 @@
...
+          left: Math.min(
+            Math.max(edgeGap, preferred),
+            window.innerWidth - width - edgeGap,
+          ),
+          top: rect.bottom + 4,
+        })
+      }
+      place()
+      const close = () => setIsOpen(false)
+      window.addEventListener('scroll', close, true)
```
Comment (gabmontes) [inline portal/app/[locale]/tunnel/transaction-history/_components/columns.tsx:124]: The portaled menu is `position: fixed` with `left`/`top` computed once from `getBoundingClientRect`. The effect closes it on `scroll` (capture) but nothing handles `resize`, so resizing the window while the menu is open leaves it at stale coordinates until the next open. The old `absolute top-9` menu followed the header naturally here. Minor, but a `resize` listener that closes (or repositions) it would restore parity. — Gabi 🤖

Comment (arielmergen) [reply portal/app/[locale]/tunnel/transaction-history/_components/columns.tsx:124]: Confirmed — `place()` only runs on open and nothing handled `resize`, so the fixed menu kept stale coordinates. It now closes on `resize` too, same as it already did on scroll.

Going back to `absolute` isn't an option though: the header sits inside the `overflow-x-hidden` container that syncs the horizontal scroll, which clipped the menu, and the body card painted over it — that's why it was portaled in the first place.

Fixed in 2242a808.

portal/stories/table.stories.tsx
```diff
@@ -0,0 +1,255 @@
...
+    mode: { control: 'inline-radio', options: ['static', 'virtual'] },
+    onRowClick: { action: 'onRowClick' },
+    placeholder: { control: false },
+  },
+  component: Table,
+  // Replicates the StakeTable shell so stories match the height the table
+  // occupies on the staking dashboard (`h-[56dvh]` / `md:min-h-136`).
+  decorators: [
+    Story => (
+      <div className="w-full max-w-4xl p-2 text-sm font-medium">
```
ArturDolzan: Following the `stories/` glob thing you found on the Loading story, `max-w-4xl` in the decorator looks like it hits the same wall (it's only used here, so it never gets emitted). Wonder if we want an inline max width, or to add `stories/` to the content globs.

arielmergen: Good catch, and it went further than `max-w-4xl` — auditing every class in `stories/` against `app/` and `components/` turned up `size-9` too.

On the globs: `tailwind.config.ts` is the same config the production build uses, so adding `stories/` would emit story-only classes into the app's CSS. Kept the globs as they are.

`size-9` came from the story's token icon being a fixed-size box: the real cell wraps `TokenLogo size="medium"` (`size-6`) in SubLogoTooltip's `rounded-lg bg-neutral-50 p-2`, which measures 40px, while the story had a `size-9` box around a `size-5` logo. The stand-in now uses the same classes and sizes itself from the padding. `max-w-4xl` has no counterpart in the app, so that one is inline.

Fixed in a5309ec0.

ArturDolzan (dismissed):
LGTM!

portal/stories/table.stories.tsx
```diff
@@ -0,0 +1,258 @@
...
+    placeholder: { control: false },
+  },
+  component: Table,
+  // Replicates the StakeTable shell so stories match the height the table
+  // occupies on the staking dashboard (`h-[56dvh]` / `md:min-h-136`).
+  decorators: [
+    Story => (
+      <div
+        className="w-full p-2 text-sm font-medium"
+        style={{ maxWidth: '56rem' }}
```
gabmontes: `maxWidth: '56rem'` is exactly `max-w-4xl` (the portal config doesn't override `theme.maxWidth`), so this inline style can revert to the `w-full max-w-4xl ...` class it replaced — keeps it consistent with the Tailwind-class convention used elsewhere. — Gabi 🤖

arielmergen: The value is right — the config doesn't override `theme.maxWidth`. The problem is emission: the `content` globs are `components/`, `app/` and `pages/`, so the JIT never sees `stories/` and `max-w-4xl` is never generated. Probed it in the Storybook page — applying it to an element computes `max-width: none`, while `max-w-96` and `max-w-72` (both used in the app) compute 384px and 288px. The decorator had been rendering unbounded all along, so the class and the inline style were never equivalent in practice.

Rather than pick another number, it now mirrors the page: the dashboard renders through `PageLayout variant="superWide"`, which sets no max width at all, so the decorator is `px-2 md:px-4 xl:px-6` — classes the app already emits.

Fixed in 7d5bf180.

gabmontes (dismissed):

portal/app/[locale]/staking-dashboard/_hooks/useRewardTokens.ts
```diff
@@ -42,5 +48,5 @@ export const useRewardTokens = function () {
     [tokenQueries],
   )
 
-  return { errors, hasError, isLoading, tokens }
+  return { errors, hasError, isLoading, isPending, tokens }
```
Comment (ArturDolzan) [inline portal/app/[locale]/staking-dashboard/_hooks/useRewardTokens.ts]: Perhaps it's worth checking whether `isLoading` still has a reader here now that `useHasRewards` returns only `hasRewards`. Since we already concluded it does not flip for disabled queries, maybe we can drop it from the return.

Comment (arielmergen) [reply portal/app/[locale]/staking-dashboard/_hooks/useRewardTokens.ts]: Right, useHasRewards was its last reader. Dropped it, and errors turned out to be unread too — the hook now returns hasError, isPending and tokens.

Fixed in [9e47f16](https://github.com/hemilabs/ui-monorepo/commit/9e47f16873c2b2b797409a5415483c63506a7fea)

portal/app/[locale]/staking-dashboard/_components/stakeTable/index.tsx
```diff
@@ -77,42 +77,31 @@ const stakingColumns = ({
...
-    cell: ({ row }) => <TimeRemaining operation={row.original} />,
-    header: () => <Header text={t('table.time-remaining')} />,
-    id: 'time-remaining',
-    meta: { className: 'justify-end', width: 140 },
+    meta: { width: 95 },
   },
   {
     cell: ({ row }) => (
-      <ActionCell openRowId={openRowId} row={row} setOpenRowId={setOpenRowId} />
+      <div className="flex w-full flex-row-reverse items-center justify-end gap-x-2 lg:flex-row lg:justify-end">
```
Comment (ArturDolzan) [inline portal/app/[locale]/staking-dashboard/_components/stakeTable/index.tsx]: Wonder if `lg:justify-end` is doing anything here, since the unprefixed `justify-end` already applies at `lg`. Looks like only `flex-row` needs the breakpoint (it's what flips which side "end" is).

Comment (arielmergen) [reply portal/app/[locale]/staking-dashboard/_components/stakeTable/index.tsx]: Right — `justify-end` already applies at `lg`, so the prefixed one emits the same `justify-content: flex-end`. Only `lg:flex-row` matters, since flipping the axis is what moves which side the "end" lands on.

Dropped it in 695dbb39. It was the only case of that pattern across `app/` and `components/`.

portal/app/[locale]/staking-dashboard/_hooks/useRewardTokens.ts
Comment (arielmergen) [reply portal/app/[locale]/staking-dashboard/_hooks/useRewardTokens.ts]: Right, `useHasRewards` was its last reader. Dropped it, and `errors` turned out to be unread too — the hook now returns `hasError`, `isPending` and `tokens`.

Fixed in 9e47f168.

ArturDolzan (approved):

gndelia (approved):

gabmontes (approved):

PR #2134, opened 2026-07-28
Wire Total earned card to the cost-basis API

### LINKED ISSUE #1965: Hemi Earn: wire 'Total earned' card to real data source

ArturDolzan:
### Description

This PR replaces the mocked "Total earned" card on `/hemi-earn` with a real per-user figure — the last piece of #1965.

For each position, earned = current value (`convertToAssets(shares)`) minus the user's cost basis from the new `/earn-cost-basis` endpoint, converted to USD, summed, and clamped so the card never shows a negative. The cost basis is
treated as opaque, so the calc doesn't care how it's computed (handy if we later switch the backend from the pure-profit fallback to FMV).

It adds `useEarnCostBasis`/`fetchEarnCostBasis`, a pure `earnedAmount` util (per-position USD + aggregate clamp) with unit tests, and invalidates the cost-basis query on settle and on cross-chain delivery so the card refreshes
when a deposit or redeem lands.

**Note: don't merge until portal-backend/api 2.6.0 is deployed**

### Screenshots

Captura de Tela 2026-07-28 às 09 25 37

### Related issue(s)

Closes #1965
Related to #1965 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_fetchers/fetchEarnCostBasis.ts
+++ b/portal/app/[locale]/hemi-earn/_fetchers/fetchEarnCostBasis.ts
@@ -11,7 +11,7 @@ export const fetchEarnCostBasis = async function ({
   account,
 }: {
   account: Address
-}): Promise<Record<string, string>> {
+}) {
   const { costBasis } = (await fetchPlusPlus(getEarnCostBasisUrl(account), {
     method: 'GET',
   })) as { costBasis: Record<string, string> }
--- a/portal/app/[locale]/hemi-earn/_fetchers/fetchEarnCostBasis.ts
+++ b/portal/app/[locale]/hemi-earn/_fetchers/fetchEarnCostBasis.ts
@@ -1,0 +1,19 @@
+import fetchPlusPlus from 'fetch-plus-plus'
+import { hemi } from 'hemi-viem'
+import { type Address } from 'viem'
+
+export const earnCostBasisKeyPrefix = ['hemi-earn', 'cost-basis'] as const
+
+const getEarnCostBasisUrl = (account: Address) =>
+  `${process.env.NEXT_PUBLIC_PORTAL_API_URL}/subgraphs/${hemi.id}/earn-cost-basis/${account}`
+
+export const fetchEarnCostBasis = async function ({
+  account,
+}: {
+  account: Address
+}): Promise<Record<string, string>> {
+  const { costBasis } = (await fetchPlusPlus(getEarnCostBasisUrl(account), {
+    method: 'GET',
+  })) as { costBasis: Record<string, string> }
+  return costBasis
+}
--- a/portal/app/[locale]/hemi-earn/_hooks/useEarnCostBasis.ts
+++ b/portal/app/[locale]/hemi-earn/_hooks/useEarnCostBasis.ts
@@ -1,0 +1,21 @@
+'use client'
+
+import { useQuery } from '@tanstack/react-query'
+import { useNetworkType } from 'hooks/useNetworkType'
+import { useAccount } from 'wagmi'
+
+import {
+  earnCostBasisKeyPrefix,
+  fetchEarnCostBasis,
+} from '../_fetchers/fetchEarnCostBasis'
+
+export const useEarnCostBasis = function () {
+  const [networkType] = useNetworkType()
+  const { address } = useAccount()
+
+  return useQuery({
+    enabled: !!address && networkType === 'mainnet',
+    queryFn: () => fetchEarnCostBasis({ account: address! }),
+    queryKey: [...earnCostBasisKeyPrefix, networkType, address],
+  })
+}
--- a/portal/app/[locale]/hemi-earn/_hooks/useEarnCostBasis.ts
+++ b/portal/app/[locale]/hemi-earn/_hooks/useEarnCostBasis.ts
@@ -8,14 +8,23 @@ import {
   earnCostBasisKeyPrefix,
   fetchEarnCostBasis,
 } from '../_fetchers/fetchEarnCostBasis'
+import { hasInFlightEarnActions } from '../_utils'
+
+import { useEarnTransactionsQuery } from './useEarnTransactionsQuery'
+import { useLocalEarnOperations } from './useLocalEarnOperations'
 
 export const useEarnCostBasis = function () {
   const [networkType] = useNetworkType()
   const { address } = useAccount()
+  const { data: transactions = [] } = useEarnTransactionsQuery()
+  const { localOperations } = useLocalEarnOperations()
 
   return useQuery({
     enabled: !!address && networkType === 'mainnet',
     queryFn: () => fetchEarnCostBasis({ account: address! }),
     queryKey: [...earnCostBasisKeyPrefix, networkType, address],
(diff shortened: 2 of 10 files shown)


portal/app/[locale]/hemi-earn/_hooks/useEarnedAmount.ts
```diff
@@ -1,17 +1,72 @@
...
+  const hasPositions = positions.length > 0
+  const isPending =
+    isPositionsPending ||
+    (hasPositions && (isPricesPending || isCostBasisLoading)) ||
+    peggedAmountQueries.some(q => q.isPending && q.isFetching)
+  const isError =
+    isPositionsError ||
+    (hasPositions && (isPricesError || isCostBasisError)) ||
+    (peggedAmountQueries.length > 0 &&
+      peggedAmountQueries.every(q => q.isError))
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useEarnedAmount.ts]: Good catch — changed the card's error condition to `some`, so any single per-position `sharesToPegged` failure now surfaces an error instead of a silently-incomplete total. [`b8bb08ff`](https://github.com/hemilabs/ui-monorepo/pull/2134/commits/b8bb08ff)

portal/app/[locale]/hemi-earn/_fetchers/fetchEarnCostBasis.ts
```diff
@@ -0,0 +1,19 @@
...
+export const fetchEarnCostBasis = async function ({
+  account,
+}: {
+  account: Address
+}): Promise<Record<string, string>> {
+  const { costBasis } = (await fetchPlusPlus(getEarnCostBasisUrl(account), {
+    method: 'GET',
+  })) as { costBasis: Record<string, string> }
+  return costBasis
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_fetchers/fetchEarnCostBasis.ts:19]:
```suggestion
export const fetchEarnCostBasis = async function ({
  account,
}: {
  account: Address
}) {
  const { costBasis } = await fetchPlusPlus(getEarnCostBasisUrl(account), {
    method: 'GET',
  }) as Promise<Record<string, string>>
  return costBasis
}
```

I think this can be simplified like this? At least let's try to remove the returned value and the extra parenthesis?

portal/app/[locale]/hemi-earn/_hooks/useEarnCostBasis.ts
```diff
@@ -0,0 +1,30 @@
...
+import { hasInFlightEarnActions } from '../_utils'
+
+import { useEarnTransactionsQuery } from './useEarnTransactionsQuery'
+import { useLocalEarnOperations } from './useLocalEarnOperations'
+
+export const useEarnCostBasis = function () {
+  const [networkType] = useNetworkType()
+  const { address } = useAccount()
+  const { data: transactions = [] } = useEarnTransactionsQuery()
+  const { localOperations } = useLocalEarnOperations()
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_hooks/useEarnCostBasis.ts:20]: you can probably acccess these inside `fetchEarnCostBasis` using `queryClient.ensureQueryData`. Better option imho

portal/app/[locale]/hemi-earn/_utils/earnedAmount.ts
```diff
@@ -0,0 +1,25 @@
...
+  price: string
+}) =>
+  Big(currentPegged.toString())
+    .minus(costBasisBaseUnits)
+    .div(Big(10).pow(decimals))
+    .times(price)
+
+// Floor the aggregate earned at 0 — an "earned" card shouldn't read negative
+// (a cooldown-redeem window or peg/transfer edge can make the net slightly < 0).
+export const clampEarnedUsd = (total: Big) => (total.lt(0) ? Big(0) : total)
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_utils/earnedAmount.ts]: do we need to use Big.js ? Can't this be solved with native bigint?

portal/app/[locale]/hemi-earn/_hooks/useEarnCostBasis.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_hooks/useEarnCostBasis.ts:20]: These two reads only feed the `refetchInterval` — they gate the 10s poll while an earn action is still in flight, mirroring `useEarnTransactionsQuery`. They need to stay reactive in the hook so RQ re-evaluates whether to keep polling; `ensureQueryData` in the fetcher wouldn't drive that, and `localOperations` isn't a query anyway (it's the local store). Happy to open an issue if you'd like to explore a cleaner shape though.

portal/app/[locale]/hemi-earn/_utils/earnedAmount.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/earnedAmount.ts]: Big.js is what the portal already uses for USD math (`useTotalDeposits`, `fiatBalance`, `utils/prices`), and here both inputs are fractional decimals — the cost basis carries WAD precision and the price is a fractional string — so native bigint would need manual fixed-point scaling to not lose precision. Leaning toward keeping it consistent, but happy to open an issue if you feel strongly.

portal/app/[locale]/hemi-earn/_fetchers/fetchEarnCostBasis.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchEarnCostBasis.ts:19]: Good catch — dropped the explicit return type in [`50ae6567`](https://github.com/hemilabs/ui-monorepo/pull/2134/commits/50ae65675b196b3dac3ac0e8cbd35732aea3d35c). Kept the `(await …)` parens though, since the cast has to bind to the resolved value rather than the promise.

gndelia (dismissed):

portal/app/[locale]/hemi-earn/_utils/earnedAmount.ts
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_utils/earnedAmount.ts]: Why are we clamping the value and prevent negatives? Total earned could go negative if something 💥 happens. Or is there anything in the design of the contracts, etc. that would prevent that?

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/earnedAmount.ts]: Good question — I checked and there's no contract-level floor. The StakingVault is a standard ERC-4626 yield vault, so `convertToAssets` moves with share price and can drift down — the spec even sets a wider 1% redeem slippage precisely because the cooldown lets the share price drift. So a real loss would make earned negative, and the clamp would hide it. It was really only there to avoid tiny transient negatives (rounding / the FMV nearest-rate approximation) showing as -$0.00x. Given your point I dropped it and now surface the real value in [`968cff16`](https://github.com/hemilabs/ui-monorepo/pull/2134/commits/968cff1659c45f66f765c1159836d104cf5c19b4).

gndelia (approved):

gabmontes (approved):

PR #2135, opened 2026-07-28
Add FMV data layer to the Hemi Earn requests subgraph

ArturDolzan:
### Description

This PR adds the subgraph data layer for fair-market-value cost basis, so the Total earned card can eventually price share tokens received via a peer-to-peer transfer instead of treating them as pure profit.

Reading the share->asset rate inline (the way the single-chain Vetro subgraph does) isn't possible here: the shares are OFTs on Hemi while the vault's convertToAssets lives on Ethereum. So this indexes two new signals the backend can join later — the raw peer-to-peer share transfers, and a global share->asset rate timeline reconstructed from the pegged amounts every processed deposit/redeem already emits. Mints, burns and the Router's in-transit legs are filtered out at the source (the Router address comes from config via the indexer, so nothing is hardcoded twice), so ShareTransfer holds only genuine wallet-to-wallet moves. Nothing consumes these yet; the backend replay lands in a follow-up.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #1965

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/subgraphs/hemi-earn-requests-subgraph/schema.graphql
+++ b/subgraphs/hemi-earn-requests-subgraph/schema.graphql
@@ -45,8 +45,8 @@ type Request {
   status: RequestStatus!
 }
 
-# Share transfers minus mints and burns. Router in-transit legs are NOT excluded
-# here (the Router address isn't available to an Envio handler) — deduped later.
+# Genuine wallet-to-wallet share transfers for FMV pricing: mints, burns, and the
+# Router's in-transit deposit/redeem legs are all excluded by the handler.
 type ShareTransfer {
   from: String! @index
   id: ID!
--- a/subgraphs/hemi-earn-requests-subgraph/schema.graphql
+++ b/subgraphs/hemi-earn-requests-subgraph/schema.graphql
@@ -55,3 +55,15 @@ type ShareTransfer {
   to: String! @index
   value: BigInt!
 }
+
+# Share->asset rate points reconstructed as num/den from each processed
+# deposit/redeem, since the vault's convertToAssets lives on another chain.
+# rateNumerator is the pegged staked/unstaked amount (the cost-basis unit),
+# NOT the underlying asset's unit; `asset` is only the timeline grouping key.
+type RateSnapshot {
+  asset: String! @index
+  id: ID!
+  rateDenominator: BigInt!
+  rateNumerator: BigInt!
+  timestamp: BigInt!
+}
--- a/subgraphs/hemi-earn-requests-subgraph/schema.graphql
+++ b/subgraphs/hemi-earn-requests-subgraph/schema.graphql
@@ -44,3 +44,14 @@ type Request {
   stakedAmount: BigInt
   status: RequestStatus!
 }
+
+# Share transfers minus mints and burns. Router in-transit legs are NOT excluded
+# here (the Router address isn't available to an Envio handler) — deduped later.
+type ShareTransfer {
+  from: String! @index
+  id: ID!
+  share: String!
+  timestamp: BigInt!
+  to: String! @index
+  value: BigInt!
+}
--- a/subgraphs/hemi-earn-requests-subgraph/src/mappings/eventHandlers.ts
+++ b/subgraphs/hemi-earn-requests-subgraph/src/mappings/eventHandlers.ts
@@ -86,6 +86,36 @@ const upsertRequest = async function ({
   context.Request.set({ ...existing, ...patch(existing) })
 }
 
+// Record a global share->asset rate point for a processed request. The asset is
+// read off the Request (set by an earlier event); a missing asset or zero
+// denominator yields no point. `id` is per-log (append-only) so a re-processed
+// request adds a point rather than overwriting the earlier one.
+const recordRate = async function ({
+  context,
+  denominator,
+  id,
+  numerator,
+  requestId,
+  timestamp,
+}: {
+  context: EvmOnEventContext
+  denominator: bigint
+  id: string
+  numerator: bigint
+  requestId: bigint
+  timestamp: bigint
+}): Promise<void> {
+  const request = await context.Request.get(requestId.toString())
+  if (!request?.asset || denominator <= 0n) return
+  context.RateSnapshot.set({
+    asset: request.asset,
+    id,
+    rateDenominator: denominator,
+    rateNumerator: numerator,
+    timestamp,
+  })
+}
+
 // ---------------------------------------------------------------------------
 // Router (Hemi) — request creation and lifecycle.
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


subgraphs/hemi-earn-requests-subgraph/src/mappings/eventHandlers.ts
```diff
@@ -84,6 +86,36 @@ const upsertRequest = async function ({
...
+  context: EvmOnEventContext
+  denominator: bigint
+  id: string
+  numerator: bigint
+  requestId: bigint
+  timestamp: bigint
+}): Promise<void> {
+  const request = await context.Request.get(requestId.toString())
+  if (!request?.asset || denominator <= 0n) return
+  context.RateSnapshot.set({
```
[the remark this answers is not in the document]
ArturDolzan: Good one to flag! I checked the Agent contract to be sure: `*RequestReceived` (which sets `asset`) is emitted right before the `handleDepositRequest` / `handleRedeemRequest` self-call that emits `*RequestProcessed` — same request, same code path — so `asset` is always populated by the time `recordRate` runs. Same-tx it's a lower logIndex; for the cooldown-claim and retry paths the Received is in an earlier block, and Envio processes a chain's events in order. The `!asset` guard is just a safety net for a degenerate case the real flow can't produce (there's a test pinning it), and failed requests emit `RequestFailed` instead of `*Processed` so they don't record a stray point either.

gndelia (commented):
Can the `upsertRequest` and `recordRate` be run in parallel instead of doing so in series?

subgraphs/hemi-earn-requests-subgraph/schema.graphql
```diff
@@ -44,3 +44,26 @@ type Request {
   stakedAmount: BigInt
   status: RequestStatus!
 }
+
+# Share transfers minus mints and burns. Router in-transit legs are NOT excluded
+# here (the Router address isn't available to an Envio handler) — deduped later.
```
gndelia: > the Router address isn't available to an Envio handler

Unsure what does that mean - the address is hardcoded so it's available

subgraphs/hemi-earn-requests-subgraph/schema.graphql
```diff
@@ -44,3 +44,26 @@ type Request {
...
+  to: String! @index
+  value: BigInt!
+}
+
+# Share->asset rate points reconstructed as num/den from each processed
+# deposit/redeem, since the vault's convertToAssets lives on another chain.
+# rateNumerator is the pegged staked/unstaked amount (the cost-basis unit),
+# NOT the underlying asset's unit; `asset` is only the timeline grouping key.
+type RateSnapshot {
+  asset: String! @index
```
gndelia, line 64: Would it be convenient to index by share here instead?

subgraphs/hemi-earn-requests-subgraph/src/mappings/eventHandlers.ts
```diff
@@ -224,6 +256,28 @@ indexer.onEvent(
...
+// ---------------------------------------------------------------------------
+// Share OFT (Hemi) — genuine peer-to-peer transfers for FMV cost basis.
+// ---------------------------------------------------------------------------
+
+indexer.onEvent(
+  { contract: 'ShareToken', event: 'Transfer' },
+  async function ({ context, event }) {
+    const from = lc(event.params.from)
+    const to = lc(event.params.to)
+    // Skip mints and burns. Router in-transit legs pass through and are deduped
```
gabmontes: The rationale "the Router address isn't available to an Envio handler" isn't quite right — the Router address is a compile-time constant already hardcoded in `config.yaml`, so it could be referenced here and used to drop the in-transit legs inline (same way `zeroAddress` is), storing less noise. Is downstream dedup still required because other non-Router legs (e.g. receiver-side delivery) also need filtering? If so, worth adjusting the comment; if not, filtering inline would be simpler.

— Auto Gabi

subgraphs/hemi-earn-requests-subgraph/schema.graphql
```diff
@@ -44,3 +44,26 @@ type Request {
   stakedAmount: BigInt
   status: RequestStatus!
 }
+
+# Share transfers minus mints and burns. Router in-transit legs are NOT excluded
+# here (the Router address isn't available to an Envio handler) — deduped later.
+type ShareTransfer {
```
gabmontes, line 50: Low priority: the subgraph README's "What it indexes" table and "Data sources" list still show only Router + Agent and don't mention the new `ShareToken` source or these `ShareTransfer` / `RateSnapshot` entities. Fine to defer until the backend consumes them, just flagging the doc drift so it isn't forgotten.

— Gabi 🤖

subgraphs/hemi-earn-requests-subgraph/schema.graphql
ArturDolzan: Good catch — the wording was misleading. You're right the address is available (it's the config.yaml literal), so I now filter the Router's in-transit legs inline in the handler and reworded the comment. ShareTransfer holds only genuine wallet-to-wallet transfers now. Done in [`9d5f692d`](https://github.com/hemilabs/ui-monorepo/pull/2135/commits/9d5f692d534eccc4a51316e695480c850ce24dc5).

subgraphs/hemi-earn-requests-subgraph/src/mappings/eventHandlers.ts
ArturDolzan: You're right on both counts. To your question: no other non-Router legs need filtering — deposit delivery is either a mint from 0x0 or a Router→receiver leg, both already covered — so inline is complete. Dropped the Router legs inline; to avoid a second hardcode I read the Router address from config via `indexer.chains[43111].Router.addresses`. Done in [`9d5f692d`](https://github.com/hemilabs/ui-monorepo/pull/2135/commits/9d5f692d534eccc4a51316e695480c850ce24dc5).

subgraphs/hemi-earn-requests-subgraph/schema.graphql
ArturDolzan: The snapshot is keyed by `asset` because that's what the processed events give us — `DepositRequestProcessed` doesn't carry the share address, whereas `asset` is already on the Request (from the Router request / Agent received event). Since each share maps 1:1 to an asset, the backend resolves share→asset and queries the timeline by asset. Indexing by share would mean storing a share we don't reliably have on the deposit path. Open to it if you see a case where per-share is cleaner!

subgraphs/hemi-earn-requests-subgraph/schema.graphql
ArturDolzan: Good call — updated the README in this PR: added the ShareToken data source and the ShareTransfer / RateSnapshot entities to "What it indexes" and "Data sources". Done in [`9d5f692d`](https://github.com/hemilabs/ui-monorepo/pull/2135/commits/9d5f692d534eccc4a51316e695480c850ce24dc5).

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2135#pullrequestreview-4808925845):
> Can the `upsertRequest` and `recordRate` be run in parallel instead of doing so in series?

They both touch the same Request entity — `recordRate` reads it (for `asset`) and `upsertRequest` reads+writes it — so keeping them serial avoids any read/write ordering ambiguity on that entity. They're also just in-memory context ops, so parallelizing wouldn't buy measurable latency. Happy to revisit if it ever shows up in profiling.

gndelia (approved):

gabmontes (approved):
uACK

PR #2146, opened 2026-08-03
Disable the whole Hemi Earn page on testnet

### LINKED ISSUE #2119: Hemi Earn: show a mainnet-only card in the transactions table on testnet (keep pools/actions enabled via ensureConnectedTo)

Hemi Earn is a mainnet-only feature, but the page still renders under the
**testnet** network toggle. We already switch the wallet to the right chain
inside each action hook via `ensureConnectedTo`, so the actions themselves work
regardless of the app toggle. The only genuinely broken surface on testnet is
the **transactions table**: its query is gated to `networkType === 'mainnet'`,
so on testnet it returns empty and renders the misleading **"No transactions"**
state — as if the user had no history, when they may have plenty on mainnet.

Rather than gating the whole page (as `stake` does), we keep the page usable and
only fix the broken surface:

- **Keep everything clickable.** Pools list, the **Manage** button, and the
  deposit/withdraw forms stay as they are. No full-page gate, no disabled
  forms — the wallet chain is handled by `ensureConnectedTo` in the mutation
  hooks.
- **Transactions table only.** On testnet, replace the table content with a
  mainnet-only card (same `InformationBox` shape used by the existing "connect
  wallet" state) with a **Change to Mainnet** action that flips the app's
  `networkType` toggle back to mainnet (via `useNetworkType()`), instead of the
  misleading empty state.
- **Audit the action hooks.** Make sure **every** Earn mutation hook
  (`useDeposit`, `useWithdraw`, `useSettle`, `useClaimUnstake`,
  `useCancelRedeem`, `useRemoteFailedActions`) calls
  `await ensureConnectedTo(chainId)`, so the wallet is always switched to the
  correct chain as part of the action. This is what makes it safe to leave the
  actions enabled on testnet.

> Note on the wallet's chain (per @gndelia): we do **not** need a "switch to
> network" CTA — `ensureConnectedTo` already switches the wallet as part of each
> action. The **Change to Mainnet** card in the transactions table is about the
> app's `networkType` toggle, not the wallet chain.

On testnet the transactions table shows an empty "No transactions" state that
misrepresents the user's mainnet history. Replacing it with a mainnet-only card
(and confirming every action hook switches the wallet) removes the confusing
surface without hiding the rest of the page.

### Proposed implementation

- In `portal/app/[locale]/hemi-earn/_components/transactionsSection/`, detect the
  testnet toggle with `useNetworkType()` and, on testnet, render a mainnet-only
  card with a **Change to Mainnet** CTA (reuse the `InformationBox` pattern; flip
  the toggle via `setNetworkType('mainnet')`).
- Leave pools, Manage, and the deposit/withdraw forms untouched.
- Verify/complete `ensureConnectedTo` coverage across all Earn mutation hooks.

ArturDolzan:
### Description

Hemi Earn is a mainnet-only feature, but on testnet nothing on the page really works — pools show mainnet data, the history is empty, and any action forces the wallet onto mainnet. So instead of leaving a half-broken page, this disables the whole Hemi Earn page on testnet.

It follows the Governance Staking pattern: the page keeps its title and swaps the body for a mainnet-only card ("Only live on Mainnet") with a "Switch to Mainnet" action that flips the app's `networkType` toggle back to mainnet. The pool detail route renders the same state on testnet, so deep links can't start a deposit/withdraw either. The wallet chain itself is still handled by `ensureConnectedTo` in the action hooks, so no "switch network" CTA is needed there.

It also removes the pool page's `redirectOnNetworkChange` effect, which stripped the `networkType` query param when switching and bounced the user back to mainnet.

### Screenshots

https://github.com/user-attachments/assets/da682f05-ccce-43dd-9996-f82c824e24bb

### Related issue(s)

Closes #2119 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/switchToMainnet.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/switchToMainnet.tsx
@@ -1,0 +1,25 @@
+'use client'
+
+import { ButtonLink } from 'components/button'
+import { LiveIcon } from 'components/icons/liveIcon'
+import { usePathname } from 'i18n/navigation'
+import { useTranslations } from 'next-intl'
+
+export const SwitchToMainnet = function () {
+  const pathname = usePathname()
+  const t = useTranslations('hemi-earn.transactions')
+  const tCommon = useTranslations('common')
+
+  return (
+    <div className="flex size-full flex-col items-center justify-center gap-y-1">
+      <LiveIcon />
+      <h4>{tCommon('only-live-on-mainnet')}</h4>
+      <p className="mb-3 max-w-44 text-center font-medium text-neutral-500 sm:max-w-64 md:max-w-72 lg:max-w-72 xl:max-w-full">
+        {t('switch-to-view')}
+      </p>
+      <ButtonLink href={{ pathname, query: { networkType: 'mainnet' } }}>
+        {tCommon('switch-to-mainnet')}
+      </ButtonLink>
+    </div>
+  )
+}
--- a/portal/app/[locale]/hemi-earn/_components/transactionsSection/switchToMainnet.tsx
+++ b/portal/app/[locale]/hemi-earn/_components/transactionsSection/switchToMainnet.tsx
@@ -2,13 +2,21 @@
 
 import { ButtonLink } from 'components/button'
 import { LiveIcon } from 'components/icons/liveIcon'
+import { useNetworkType } from 'hooks/useNetworkType'
 import { usePathname } from 'i18n/navigation'
 import { useTranslations } from 'next-intl'
+import { MouseEvent } from 'react'
 
 export const SwitchToMainnet = function () {
+  const [, setNetworkType] = useNetworkType()
   const pathname = usePathname()
   const t = useTranslations('hemi-earn.transactions')
   const tCommon = useTranslations('common')
+
+  const onClick = function (e: MouseEvent<HTMLAnchorElement>) {
+    e.preventDefault()
+    setNetworkType('mainnet')
+  }
 
   return (
     <div className="flex size-full flex-col items-center justify-center gap-y-1">
@@ -17,7 +25,10 @@ export const SwitchToMainnet = function () {
       <p className="mb-3 max-w-44 text-center font-medium text-neutral-500 sm:max-w-64 md:max-w-72 lg:max-w-72 xl:max-w-full">
         {t('switch-to-view')}
       </p>
-      <ButtonLink href={{ pathname, query: { networkType: 'mainnet' } }}>
+      <ButtonLink
+        href={{ pathname, query: { networkType: 'mainnet' } }}
+        onClick={onClick}
+      >
         {tCommon('switch-to-mainnet')}
       </ButtonLink>
     </div>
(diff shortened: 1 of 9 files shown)


portal/app/[locale]/hemi-earn/_components/transactionsSection/switchToMainnet.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <h4>{tCommon('only-live-on-mainnet')}</h4>
+      <p className="mb-3 max-w-44 text-center font-medium text-neutral-500 sm:max-w-64 md:max-w-72 lg:max-w-72 xl:max-w-full">
+        {t('switch-to-view')}
+      </p>
+      <ButtonLink href={{ pathname, query: { networkType: 'mainnet' } }}>
+        {tCommon('switch-to-mainnet')}
+      </ButtonLink>
+    </div>
+  )
+}
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/switchToMainnet.tsx:36]: Good catch — now flipping the toggle via `setNetworkType()` on click and keeping the `href` just for link semantics, so other query params are preserved. Done in [`291eb976`](https://github.com/hemilabs/ui-monorepo/pull/2146/commits/291eb976).

gndelia:
Rather than the Table, I think the whole page should be disabled, as nothing really works on testnet

See how we do it on Governance Staking

image

dvnahuel:
> Rather than the Table, I think the whole page should be disabled, as nothing really works on testnet
> 
> See how we do it on Governance Staking
> 
> image

@ArturDolzan Let me work on designs on this, following the pattern for the governance staking page.

dvnahuel:
LGTM!

ArturDolzan:
> [@dvnahuel commented](https://github.com/hemilabs/ui-monorepo/pull/2146#issuecomment-5169728625):
> @ArturDolzan Let me work on designs on this, following the pattern for the governance staking page.

Thanks @dvnahuel! As we discussed, the whole page is now disabled on testnet following the Governance Staking pattern — done in [`27e48c63`](https://github.com/hemilabs/ui-monorepo/pull/2146/commits/27e48c63). cc @gndelia

portal/app/[locale]/hemi-earn/page.tsx
```diff
@@ -36,14 +38,21 @@ const TokensGate = function ({ children }: { children: ReactNode }) {
 }
 
 export default function Page() {
+  const [networkType] = useNetworkType()
+  const isEnabled = networkType !== 'testnet'
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/page.tsx]: should this go into layout? https://github.com/hemilabs/ui-monorepo/blob/main/portal/app/%5Blocale%5D/hemi-earn/layout.tsx

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/page.tsx]: Good call — moved the gate into the layout so it covers the list and the pool detail in one spot. `page.tsx` goes back to untouched, and the now-dead `redirectOnNetworkChange` in `poolPageContent` is dropped too. Done in [`24d611a4`](https://github.com/hemilabs/ui-monorepo/pull/2146/commits/24d611a4).

gndelia (approved):

gndelia:
Note that this card already exists in 2 extra places, so it may be worth extracting in a component "TestnetDisabled" or something like that (not a blocker for this PR though)

ArturDolzan:
> Note that this card already exists in 2 extra places, so it may be worth extracting in a component "TestnetDisabled" or something like that (not a blocker for this PR though)

Sounds good. Opened an issue for it. Thanks!

gabmontes (approved):

PR #2153, opened 2026-08-04
Add `mint` subcommand to hemi-earn sandbox

ArturDolzan:
### Description

This PR ports the `mint` sandbox subcommand that calls `mint(address,uint256)` on any ERC20-mock running under the local sandbox.

Before this, two flows still needed raw `cast send`: topping up an EOA mid-session (past what `setup` funds on init) and injecting yield into the staking vault (mint the pegged token straight to the vault so `totalAssets()` inflates without new shares — that's what makes the portal's "Total earned" card start showing profit). Both now go through `pnpm --filter portal sandbox:hemi-earn -- mint`.

Implementation follows the same shape as `fail-gateway`: inline `parseAbiItem`, no artifact dependency, `writeContract` + receipt-status check, and registered in the top-level dispatcher. All existing mocks (`LabeledERC20Mock`, `PeggedTokenMock`, aliased Vetro contracts) share the same OZ-style `mint` signature so a single subcommand covers them all.

Broken down into:
- `b886c7fd` — Add mint sandbox subcommand
- `ba24c8ae` — Document mint in sandbox README

### Screenshots

Captura de Tela 2026-08-04 às 10 41 31

### Related issue(s)

Related to #2092 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/scripts/hemi-earn/README.md
+++ b/portal/scripts/hemi-earn/README.md
@@ -70,7 +70,7 @@ node portal/scripts/hemi-earn/fundAccount.ts [flags]
 Flags are parsed by the handler of each subcommand.
 
 - `setup` — `--address` / `-a` (required), `--port` / `-p` (default `8545`), `--upstream-rpc` / `-u` (default `https://rpc.hemi.network/rpc`), `--fork-url` / `-f` (skips auto-start), `--deployer-pk` (default is Anvil's well-known account #0).
-- `mint` — `--token` / `-t` (required), `--to` (required), `--amount` / `-n` (default `10`, parsed as 18-dec units). Optional: `--fork-url` / `-f`, `--deployer-pk`.
+- `mint` — `--token` / `-t` (required), `--to` (required), `--amount` / `-n` (default `10`, parsed via `parseEther` — sandbox mocks are all 18-decimal by design, no `decimals()` lookup). Optional: `--fork-url` / `-f`, `--deployer-pk`.
 - `mining` — `--seconds` / `-s` (default `6`, `0` returns to instant mining), `--fork-url` / `-f` (default `http://127.0.0.1:8545`).
 - `relayer` — `--router` / `-r` (required), `--agent` / `-a` (required) — both come from the address banner `setup` prints; `--fork-url` / `-f`, `--deployer-pk`, `--poll` (seconds between ticks, default `1`), `--from-block N` (first block to backfill from, default `0` — full history), `--disable-autoclaim` (observe events but skip the claim; simulates a downed keeper).
 - `fail-gateway` — either `--status` (read-only, prints the current state) or `--kind` / `-k` (`deposit` | `redeem`) + `--mode` / `-m` (`off` | `on` | `slippage` | `fee` | `unknown`). Optional: `--fork-url` / `-f`, `--deployer-pk`.
@@ -89,7 +89,7 @@ Every sandbox token exposes the OpenZeppelin `mint(address,uint256)` shape, so a
 ```bash
 # Top up an EOA with 5 hemiBTC (copy the token address from the setup banner)
 pnpm --filter portal sandbox:hemi-earn -- mint \
-  --token 0x5FbDB2315678afecb367f032d93F642f64180aa3 \
+  --token 0xYourHemiBTC \
   --to    0xYourEOA \
   --amount 5
 
--- a/portal/scripts/hemi-earn/README.md
+++ b/portal/scripts/hemi-earn/README.md
@@ -53,6 +53,7 @@ pnpm --filter portal sandbox:hemi-earn -- <subcommand> [flags]
 | Subcommand     | Purpose                                                                                                                  |
 | -------------- | ------------------------------------------------------------------------------------------------------------------------ |
 | `setup`        | Start Anvil + deploy mocks + fund the test account.                                                                      |
+| `mint`         | Mint from any ERC20-mock — top up an EOA or inject yield into the vault (see [Minting](#minting)).                       |
 | `mining`       | Toggle Anvil's interval mining at runtime (see [Slow mining](#slow-mining)).                                             |
 | `relayer`      | Emulate the production keeper: claim mature cooldown redeems and bridge cancellation requests (see [Relayer](#relayer)). |
 | `fail-gateway` | Toggle `PreviewableGatewayMock` into a failure mode (see [Failure simulation](#failure-simulation)).                     |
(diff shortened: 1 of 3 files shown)


portal/scripts/hemi-earn/mint.ts
```diff
@@ -0,0 +1,114 @@
...
+const mintAbi = parseAbiItem('function mint(address,uint256)')
+
+function printUsage() {
+  console.error(
+    'Usage: pnpm --filter portal sandbox:hemi-earn -- mint --token 0x... --to 0x... [--amount 10] [flags]',
+  )
+  console.error('  -t, --token TOKEN       ERC20-mock address (required)')
+  console.error('      --to RECIPIENT      recipient address (required)')
+  console.error('  -n, --amount AMOUNT     amount in ether units (default 10)')
+  console.error(
```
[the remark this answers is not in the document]
ArturDolzan: Good catch — done in [`1b29d1c6`](https://github.com/hemilabs/ui-monorepo/pull/2153/commits/1b29d1c6641e582db3eb1bef5b9d1da8b0d8752c).

portal/scripts/hemi-earn/README.md
```diff
@@ -77,6 +79,27 @@ Flags are parsed by the handler of each subcommand.
...
+
+Every sandbox token exposes the OpenZeppelin `mint(address,uint256)` shape, so a single subcommand covers both common flows:
+
+- **Top up an EOA** — mint deposit assets to a wallet you're testing with. `setup` already funds the test account with 10 of each on init; use this to bump balances mid-session or fund a second wallet.
+- **Inject yield** — mint the pegged token directly into the staking vault. That inflates `totalAssets()` without issuing new shares, so `convertToAssets(userShares)` grows proportionally and the portal's "Total earned" card starts showing profit.
+
+```bash
+# Top up an EOA with 5 hemiBTC (copy the token address from the setup banner)
+pnpm --filter portal sandbox:hemi-earn -- mint \
+  --token 0x5FbDB2315678afecb367f032d93F642f64180aa3 \
```
gndelia: Where does this address `0x5FbDB2315678afecb367f032d93F642f64180aa3` come from? It appears to be tagged as a phishing address on Etherscan 🤔 - see [etherscan](https://etherscan.io/address/0x5FbDB2315678afecb367f032d93F642f64180aa3)

portal/scripts/hemi-earn/README.md
```diff
@@ -69,6 +70,7 @@ node portal/scripts/hemi-earn/fundAccount.ts [flags]
 Flags are parsed by the handler of each subcommand.
 
 - `setup` — `--address` / `-a` (required), `--port` / `-p` (default `8545`), `--upstream-rpc` / `-u` (default `https://rpc.hemi.network/rpc`), `--fork-url` / `-f` (skips auto-start), `--deployer-pk` (default is Anvil's well-known account #0).
+- `mint` — `--token` / `-t` (required), `--to` (required), `--amount` / `-n` (default `10`, parsed as 18-dec units). Optional: `--fork-url` / `-f`, `--deployer-pk`.
```
gndelia: `parsed as 18-dec units`

I think it should be parsed according to token's decimals. Not all tokens use `18` units - `hemiBTC` for example uses 8

portal/scripts/hemi-earn/mint.ts
```diff
@@ -0,0 +1,116 @@
...
+  if (!/^0x[0-9a-fA-F]{64}$/.test(deployerPk)) {
+    console.error(
+      '✗ --deployer-pk must be a 32-byte hex string starting with 0x',
+    )
+    printUsage()
+    process.exit(1)
+  }
+
+  return {
+    amount: parseEther(values.amount ?? '10'),
```
gndelia: `--amount` is the only input that isn't validated. `--token`, `--to` and `--deployer-pk` all fail with a friendly `printUsage()` + exit, but a few amounts slip past this line:

- `--amount ''` → `parseEther('')` returns `0n`, so it mints nothing and still prints `✓ minted  to 0x…` — a silent no-op reported as success.
- `--amount -5` → `-5000000000000000000n`, which then fails deep inside viem's ABI encoder instead of the usage block.
- `--amount abc` / `1e3` → `InvalidDecimalNumberError`; handled by the dispatcher's catch, but off-shape vs the other flags.

A positive-decimal check next to the address checks would cover all four cases.

gabmontes: `--amount` skips the friendly validation the other flags here get: `--token`, `--to` and `--deployer-pk` all `printUsage()` + `exit(1)` on bad input, and the sibling subcommands (`mining` → `--seconds`, `relayer` → `--poll`) guard their numeric flags too. Here an invalid value (e.g. `--amount abc`) reaches `parseEther` and surfaces a raw viem error instead of the usage banner. Consider validating before parsing for consistency. — Gabi 🤖

portal/scripts/hemi-earn/README.md
ArturDolzan: Fair — that address is anvil's deterministic first-deploy slot from the setup, unrelated to the mainnet phishing entry, but confusing enough for anyone clicking through. Swapped it for a `0xYourHemiBTC` placeholder in [`e4beeb74`](https://github.com/hemilabs/ui-monorepo/pull/2153/commits/e4beeb74).

portal/scripts/hemi-earn/README.md
ArturDolzan: True — reworded to say `parseEther` + a note that sandbox mocks are all 18-decimal by design (unlike real hemiBTC's 8-dec on Ethereum) in [`e4beeb74`](https://github.com/hemilabs/ui-monorepo/pull/2153/commits/e4beeb74).

portal/scripts/hemi-earn/mint.ts
ArturDolzan: Good catch — added a positive-decimal regex + a `> 0` check in [`e4beeb74`](https://github.com/hemilabs/ui-monorepo/pull/2153/commits/e4beeb74). Covers `''` / `-5` / `abc` / `1e3` and the silent-zero case too.

ArturDolzan: Good catch — validated `--amount` with a positive-decimal regex + `> 0` check in [`e4beeb74`](https://github.com/hemilabs/ui-monorepo/pull/2153/commits/e4beeb74) so it fails with the same `printUsage()+exit(1)` shape as the sibling numeric flags.

gabmontes (approved):

gndelia (approved):

PR #2157, opened 2026-08-04
Extract shared TestnetDisabled component

### LINKED ISSUE #2152: Extract the testnet-disabled (mainnet-only) card into a shared component

Extract the "disabled on testnet / mainnet-only" screen into a single reusable
component (e.g. `TestnetDisabled` / `MainnetOnlyState`) under
`portal/components/`, and have the mainnet-only features render it instead of
each shipping their own copy.

The same card — `LiveIcon` + the `common.only-live-on-mainnet` heading + a short
subtitle + a "Switch to Mainnet" CTA (`common.switch-to-mainnet`) — is now
duplicated across three features:

- `portal/app/[locale]/hemi-earn/_components/earnDisabledTestnet.tsx`
- `portal/app/[locale]/staking-dashboard/_components/stakingDashboardDisabledTestnet.tsx`
- `portal/app/[locale]/genesis-drop/_components/claimRewardsDisabledTestnet.tsx`

They've already started to drift: Hemi Earn flips the toggle via
`setNetworkType()` on click while keeping `href` only for link semantics (so
other query params are preserved), whereas the other two navigate with a plain
`href`. Keeping three copies means every future tweak has to be applied in three
places, and they'll keep diverging.

Came up in the review of #2146 (non-blocking).

### Proposed implementation

- Add a shared component in `portal/components/` that renders the `LiveIcon`, the
  `common.only-live-on-mainnet` heading, a `subtitle` (prop), and the "Switch to
  Mainnet" CTA (`common.switch-to-mainnet`).
- Standardize the CTA on the query-param-preserving approach:
  `setNetworkType('mainnet')` on click + `href` for link semantics.
- Each call site passes its feature-specific subtitle and drops its local copy:
  - `hemi-earn.switch-to-start-earning`
  - `staking-dashboard.switch-to-start-staking`
  - `genesis-drop.switch-to-start-claiming`

ArturDolzan:
### Description

This PR extracts the "disabled on testnet / mainnet-only" card into a single shared component. The same card — LiveIcon + "Only live on Mainnet" + a "Switch to Mainnet" CTA — lived in three near-identical copies across Hemi Earn, Governance Staking and Genesis Drop, and had already started to diverge, so every future tweak meant touching three places.

The new `components/testnetDisabled.tsx` takes the feature-specific subtitle as a prop and a `variant` (`card` for Earn/Staking, `overlay` for Genesis' absolute-positioned layout); the heading, CTA and toggle flip live inside it. The three call sites render it directly and the three per-feature wrappers are deleted.

It also standardizes the CTA on the query-param-preserving approach: Staking and Genesis now flip `networkType` via `setNetworkType()` on click (keeping other query params) like Earn already did, rather than a plain link navigation.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Closes #2152 

### Checklist

- [x] Manual testing passed.


portal/components/testnetDisabled.tsx
```diff
@@ -7,29 +7,43 @@ import { usePathname } from 'i18n/navigation'
 import { useTranslations } from 'next-intl'
 import { MouseEvent } from 'react'
 
-export const EarnDisabledTestnet = function () {
+const variants = {
+  card: 'mt-8 rounded-xl bg-white py-48 shadow-md',
+  overlay: 'absolute inset-0',
+}
```
gndelia:
```suggestion
const variants = {
  card: 'mt-8 rounded-xl bg-white py-48 shadow-md',
  overlay: 'absolute inset-0',
} as const
```

should it be with `as const` ?

gndelia (dismissed):
I left a minor question, otherwise, LGTM

portal/components/testnetDisabled.tsx
ArturDolzan: Good call — matches how `ButtonLink` types its maps. Done in [`ed53efa0`](https://github.com/hemilabs/ui-monorepo/pull/2157/commits/ed53efa0).

gabmontes (approved):

gndelia (approved):

PR #2170, opened 2026-08-05
Align Hemi Earn cooldown countdown across surfaces

### LINKED ISSUE #2156: Hemi Earn: cooldown in the transactions table drops hours ("6d" vs drawer's "6d 23h")
### Context

On `/hemi-earn`, the cooldown countdown for a withdrawal is shown at two granularities. The **drawer** reads `Cooldown ends in 6d 23h` (days + hours), while the **transactions table** row reads `Cooldown · ready in 6d` (days only) for the same request. They should match.

### Problem

The two surfaces format the remaining time differently:

- **Table** — `formatCooldownText` in `portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx` shows only the **largest** whole unit, floored: `≥ 1 day → cooldown-ready-in-days` (`6d`), dropping the trailing hours.
- **Drawer** — `cooldownPostAction.ts` (`.../pool/[shareAddress]/_components/poolReview/`) breaks it into days **and** hours:

  ```ts
  const remainingDays = Math.floor(secondsToDays(cooldownRemainingSec))
  const remainingHours = Math.floor(
    secondsToHours(cooldownRemainingSec - remainingDays * 86400),
  )
  t('wait-cooldown-countdown', { days: remainingDays, hours: remainingHours }) // "6d 23h"
  ```

So a request with 6d 23h left reads `6d` in the table but `6d 23h` in the drawer.

### Proposed fix

Align the table with the drawer's day + hour granularity so both read `6d 23h`. Best done by reusing the same breakdown (`secondsToDays` / `secondsToHours` + a shared `#d #h` string) rather than duplicating it, so the two can't drift again.

Image

ArturDolzan:
### Description

This PR fixes the Hemi Earn cooldown countdown showing at different granularities depending on where you look. The withdraw drawers already rendered "6d 23h", but the transactions table row dropped the hours and showed just "6d".

The root of the drift was that the day/hour breakdown lived inline inside the drawers' deriveCooldownPostAction, while the table rolled its own single-largest-unit logic. So the first commit pulls that breakdown into a shared secondsToDaysAndHours helper in utils/time (now backing both drawers), and the second routes the table through it so every surface reads the same "6d 23h". While there, I also aligned the sub-hour tier: the drawers deliberately say "less than an hour" below 3600s (the countdown only ticks every 60s, so a per-minute number would lie), so the table now shows "< 1h" instead of "2m"/"< 1m" to match.

  Broken down into:
  - `a24af88b` — Extract cooldown day/hour split into utils/time
  - `457a1979` — Align cooldown table with drawer granularity
 
### Screenshots

https://github.com/user-attachments/assets/aa74b655-874f-4295-9def-527d47f4ad99

### Related issue(s)

Closes #2156 

### Checklist

- [x] Manual testing passed.


portal/test/utils/time.test.ts
```diff
@@ -9,6 +13,27 @@ describe('utils/time', function () {
     })
   })
 
+  describe('secondsToDaysAndHours', function () {
```
gabmontes, line 16: These four cases all use exact hour multiples, so none exercise a sub-hour remainder — they don't prove the breakdown floors rather than rounds. Consider adding a dirty-remainder case to pin both floors at once:

```ts
it('should floor extra minutes and seconds instead of rounding up', function () {
  expect(secondsToDaysAndHours(6 * 86400 + 23 * 3600 + 3599)).toEqual({
    days: 6,
    hours: 23,
  })
})
```

`3599s` of extra remainder must stay `{ days: 6, hours: 23 }`; this guards against a future refactor swapping `Math.floor` for `Math.round`.

— Auto Gabi

portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx
```diff
@@ -14,28 +15,21 @@ import { StatusBadge } from './statusBadge'
 
 const SECONDS_PER_DAY = 86_400
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx]: These duplicate magnitudes that already live in `utils/time` (`secondsPerDay`, plus the divisor behind `secondsToHours`). Since this PR is centralizing the day/hour math, consider exporting `secondsPerDay` / `secondsPerHour` from `utils/time` and importing them here instead of redefining `SECONDS_PER_DAY` / `SECONDS_PER_HOUR` — a single source of truth for the boundaries too. Bonus: it also drops these `SCREAMING_SNAKE_CASE` locals in favor of the repo's camelCase convention.

— Gabi 🤖

portal/test/utils/time.test.ts
ArturDolzan: Good call — added that dirty-remainder case in [`5500d0a4`](https://github.com/hemilabs/ui-monorepo/pull/2170/commits/5500d0a4); the extra 3599s stays `{ days: 6, hours: 23 }`, so it now pins the floor against a future round swap.

portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx]: Agreed — exported `secondsPerDay`/`secondsPerHour` from `utils/time` and dropped the local SCREAMING_SNAKE constants for those in [`5500d0a4`](https://github.com/hemilabs/ui-monorepo/pull/2170/commits/5500d0a4). Single source for the boundaries now too.

gabmontes (dismissed):

portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx
```diff
@@ -12,30 +17,20 @@ import { type EarnPool, type EarnTransaction } from '../../types'
...
   seconds: number,
   t: ReturnType<typeof useTranslations<'hemi-earn.transactions'>>,
 ) {
-  if (seconds >= SECONDS_PER_DAY) {
-    return t('status.cooldown-ready-in-days', {
-      value: Math.floor(seconds / SECONDS_PER_DAY),
-    })
+  if (seconds >= secondsPerDay) {
+    const { days, hours } = secondsToDaysAndHours(seconds)
+    return t('status.cooldown-ready-in-days-hours', { days, hours })
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx:27]: When the row has no `claimableAt` yet (redeem submitted, subgraph hasn't indexed it), `displaySec` falls back to `cooldownDurationSec` at line 93:

```ts
const displaySec = hasClaimableAt ? (remainingSec ?? 0) : cooldownDurationSec
```

That value is the static, whole-day vault duration, so this branch now renders `Cooldown · ready in 7d 0h` where it used to render `7d`. The trailing `0h` carries no information, and more importantly it isn't a countdown at all — it's a fixed duration being presented as one.

The equivalent drawer state uses `wait-cooldown-pending` ("Cooldown: 7 days"), so this ends up being the one state the PR's "every surface reads the same" goal doesn't cover. Worth keeping the day-only key for the `!hasClaimableAt` branch.

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/withdrawStatusCell.tsx:27]: Nice catch — you're right, that pre-index state was rendering the static vault duration as a countdown. Restored the day-only key for the `!hasClaimableAt` branch so it reads `7d` again (no phantom `0h`), and kept days+hours for the actual countdown. Done in [`5fbcaa3b`](https://github.com/hemilabs/ui-monorepo/pull/2170/commits/5fbcaa3b).

gndelia (approved):

gabmontes (approved):

Issue #2008, opened 2026-06-16
Allow different vaults for btc deposits and withdrawals


gndelia:
### Feature or change description

Currently, a vault is configured for both deposits and withdrawals. We want to allow, through configuration, so the btc deposit vault can be different from the withdrawal one.

Note that this doesn't mean the user is able to select vaults - it just means that we are able to go through different wallets depending on the flow

History resync should also resync the full history keeping this in mind - note that up to before this PR, a vault would work for both deposits and withdrawals, but after this, the sync process would go split

gndelia:
Implementation notes

## Findings

Today, a single vault index per chain drives both flows:

- `portal/utils/hemiClientExtraActions.ts`
  - `defaultBitcoinVaults` ← `NEXT_PUBLIC_DEFAULT_BITCOIN_VAULT_{MAINNET,SEPOLIA}` (defaults to `0`)
  - `pastBitcoinVaults` ← `NEXT_PUBLIC_BITCOIN_PAST_VAULTS_{MAINNET,SEPOLIA}`
  - `getVaultChildIndex(client)` → active vault, used by both flows
  - `getVaultHistoricVaultIndexes(client)` → `[...past, current]`, used only by the deposit sync

The two flows already reach the chain through separate call sites, so splitting them is mostly
mechanical:

| Flow       | Call sites                                                                                                                                                                                                                                         |
| ---------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| deposit    | `utils/hemi.ts` `initiateBtcDeposit`, `utils/hemi.ts` `getBitcoinDepositFee`, `utils/hemiMemoized.ts` `getVaultIndexByBTCAddress`, `_hooks/useMinDepositSats.ts`                                                                                   |
| withdrawal | `utils/hemi.ts` `getHemiStatusOfBtcWithdrawal`, `utils/hemi.ts` `initiateBtcWithdrawal`, `utils/hemi.ts` `getBitcoinWithdrawalFee`, `_hooks/useMinWithdrawalSats.ts`, `_hooks/useEstimateBtcWithdrawFees.ts`, `hooks/useSimpleVaultGracePeriod.ts` |

`useEstimateBtcDepositFees` resolves its vault through `getVaultIndexByBTCAddress`, so it needs no
direct change.

### The blocking problem

`getHemiStatusOfBtcWithdrawal` resolves the vault from **configuration**, not from the withdrawal
[...]

PR #2171, opened 2026-08-06
Return the vault of BTC withdrawals in the API

gndelia:
### Description

The BTC withdrawals subgraph now indexes the vault each withdrawal was initiated against. This exposes it through the API: `GET /subgraphs/:chain-id/withdrawals/:address/btc` now returns a `vault` field per withdrawal.

Until now consumers had to assume the vault from configuration. That assumption only holds while a single vault serves both deposits and withdrawals — once they diverge, a withdrawal initiated against a previous vault gets checked against the wrong vault state, and the portal would offer to challenge a withdrawal that was already fulfilled.

### Screenshots

N/A — no UI changes.

### Related issue(s)

Related to #2008

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal-backend/api/src/subgraphs/types/tunnel.ts
+++ b/portal-backend/api/src/subgraphs/types/tunnel.ts
@@ -1,6 +1,6 @@
 // Copied from https://github.com/hemilabs/ui-monorepo/blob/853f366d/webapp/types/tunnel.ts and modified to remove unused types
 
-import { type Chain, type Hash } from 'viem'
+import { type Address, type Chain, type Hash } from 'viem'
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const MessageDirection = {
@@ -63,6 +63,7 @@ export type ToBtcWithdrawOperation = CommonOperation &
     challengeTxHash?: Hash
     uuid?: string // bigint can't be serialized into local storage
     netSatsAfterFee: string
+    vault: Address
   }
 
 export type BtcDepositOperation = {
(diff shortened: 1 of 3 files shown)


portal-backend/api/src/subgraphs/types/tunnel.ts
```diff
@@ -63,6 +63,7 @@ export type ToBtcWithdrawOperation = CommonOperation &
     challengeTxHash?: Hash
     uuid?: string // bigint can't be serialized into local storage
     netSatsAfterFee: string
+    vault: Address
```
gabmontes, line 66: `vault: Address` is inconsistent with every other address field here — `from`, `l1Token`, `to`, and the deposit's own `vault` are all typed `string` — and it's also inaccurate: this type backs `GetBtcWithdrawalsQueryResponse`, where `d.vault` is the raw subgraph value, which is lowercased (not checksummed). That's why the sibling fields carry `// @ts-expect-error OP-SDK does not properly type addresses as Address` before `toChecksum(...)`; typing `vault` as `Address` just hides that same mismatch.

Suggest `vault: string`, drop the new `Address` import, and add the `@ts-expect-error` line before `vault: toChecksum(d.vault)` — matching how the BTC deposit path already handles its `vault`.

— Gabi 🤖

gndelia: To be fair, the correct type is Address. As the "expect" error says, `OP-SDK does not properly type addresses`  - but this type does not come from the OP-SDK - it's ours. So we can just use `Address` directly, which is correct (Even if not checksummed)

ArturDolzan (approved):

gabmontes (approved):

PR #2182, opened 2026-08-10
Add slippage and approval settings to Hemi Earn

### LINKED ISSUE #1981: Implement Slippage Control and Advanced Settings on Hemi Earn.

ArturDolzan:
### Description

This PR implements the slippage control and Advanced settings panel for Hemi Earn.

Slippage was hardcoded at 0.5% for deposits and 1% for redeems, which meant that when a min-out failure happened there was no way to widen it and retry — the user just kept hitting the same wall. Now there's a gear in the form header that opens a panel with a max slippage control and the extra-approval toggle.

Auto keeps the existing per-operation defaults rather than collapsing them into one: a redeem freezes its min-out across the ~7 day cooldown, so it genuinely needs the wider value. On the other end, values are floored at 0.1% — a zero tolerance demands the exact quote, and any rate drift before the remote fulfillment would revert. Above 5% the panel warns inline, and committing anything past that asks for an explicit risk acknowledgement.

Broken down into:
- `0345ed6a` — Extract shared toggle and checkbox primitives
- `993b29c4` — Support a custom approval amount in earn actions
- `310e2d51` — Add slippage and approval settings to Hemi Earn

The first two are prep: the toggle and checkbox already existed but were private to the tunnel and to the custom token drawer, and the actions had no way to request a larger allowance. Both consumers pick up minor visual changes (a hover state on the checkbox, a keyboard-only focus ring) — worth a look when reviewing.

### Screenshots

https://github.com/user-attachments/assets/58fca1ac-bb31-46b3-8697-aecc69df0ac5

**Storybook:**

https://github.com/user-attachments/assets/d60942f5-7db9-41e5-b42c-74bb60668668

### Related issue(s)

Closes #1981 

### Checklist

- [x] Manual testing passed.


portal/components/toggle.tsx
```diff
@@ -0,0 +1,37 @@
+type Props = {
+  checked: boolean
+  disabled?: boolean
+  id: string
+  onCheckedChange: (checked: boolean) => void
```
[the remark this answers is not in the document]
ArturDolzan: Good catch — the toggle now takes an `ariaLabel` and both callers pass their translated label, in [`ca63b5b9`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/ca63b5b9).

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/highSlippageModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+  slippage,
+}: Props) {
+  const [accepted, setAccepted] = useState(false)
+  const t = useTranslations('hemi-earn.pool.settings')
+  const tCommon = useTranslations('common')
+
+  const isVeryHigh = getSlippageLevel(slippage) === 'veryHigh'
+
+  return (
+    <Modal onClose={onClose}>
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/highSlippageModal.tsx:54]: Good point, and it is reachable through the blur-commit path. Focus now moves to the acknowledgement checkbox on mount and is restored on close in [`ca63b5b9`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/ca63b5b9) — full trapping would need to live in the shared `Modal`, which the other four dialogs use too, so I left that out of this PR.

packages/hemi-earn-actions/src/actions/wallet/requestDeposit.ts
```diff
@@ -49,6 +51,9 @@ const runRequestDeposit = ({
 }: {
   account: Address
   amount: bigint
+  // Allowance to request when approving; defaults to amount. Larger values let the
+  // user pre-approve future deposits and skip the approval step next time.
```
gndelia: I don't think this adds value - it's something we've used in several packages widely

```suggestion
```

packages/hemi-earn-actions/src/actions/wallet/requestRedeem.ts
```diff
@@ -57,6 +59,9 @@ const runRequestRedeem = ({
   walletClient,
 }: {
   account: Address
+  // Allowance to request when approving; defaults to the adjusted shares. Larger values
+  // let the user pre-approve future redeems and skip the approval step next time.
```
gndelia: same here

packages/hemi-earn-actions/src/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+export const resolveApprovalAmount = ({
+  approvalAmount,
+  required,
+}: {
+  approvalAmount: bigint | undefined
+  required: bigint
+}) =>
+  approvalAmount !== undefined && approvalAmount > required
+    ? approvalAmount
+    : required
```
gndelia: To me, this is too verbose. ApprovalAmount doesn't come from a user input, so there's no way `approvalAmount` ends up being less than `required`. To me, this should be simply `approvalAmount ?? required`. If anything, just pass both to the package, which should already do this validation (of using `approvalAmount ?? required` and failing if incorrect), which makes this function to add no value

Edit: SORRY, I MISREAD. I thought this file was on the `portal` project. Here it's ok. However, I would still use `approvalAmount ?? required`  and throw if not. Let the users know they're sending invalid things!

portal/app/[locale]/hemi-earn/_constants/slippage.ts
```diff
@@ -1,7 +1,17 @@
...
-export const REDEEM_SLIPPAGE_BPS = BigInt(100)
+export const defaultDepositSlippage = 0.5
+export const defaultRedeemSlippage = 1
+
+export const highSlippageThreshold = 5
+export const veryHighSlippageThreshold = 10
+
+// A zero tolerance makes minOut equal the quote exactly, which only holds if the
+// rate doesn't move between the request and the remote fulfillment — and for a
+// redeem that window is the whole ~7d cooldown. Floor it instead of allowing 0.
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_constants/slippage.ts]: 🤔 

I'm wondering if overriding user input is correct. Should we here warn to the user instead? Maybe they just want slippage (which, under normal circumstances, may revert). To me, it's similar to the scenario of allowing 100% slippage. This is probably more of a UX question

gndelia (commented):
The `Checkbox` component is missing stories - we could add them (given now Storybook is a real thing).

I think the same applies to `Toggle`, though that one technically existed before. But since we're moving them, it'd be nice to add t hem 😄 

This also makes me thing that the PR could've been split in a few smaller PRs. This one has more than 1k lines updated - the components could've been added in one PR, packages change in a 2nd one, and then changes for the portal in a 3rd one. It makes it easier to review

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx
```diff
@@ -56,7 +57,7 @@ export const RetryFailedDeposit = function ({
   })
 
   const sharesOutMin = shares
-    ? applySlippage(shares, DEPOSIT_SLIPPAGE_BPS)
+    ? applySlippage(shares, percentToBps(defaultDepositSlippage))
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx:60]: Say the user enters 0% slippage (or something different than the default). The tx fails (for whatever reason), so the user must retry (which I understand is what this drawer shows). Is the Retry using a different slippage (the default one) than the user entered?

gndelia:
The gear icon looks different from Figma


| Figma | This PR |
| ------- | --------|
| image | image |
| image | image  |

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/panel.tsx
```diff
@@ -0,0 +1,134 @@
...
+  )
+
+  return (
+    <div className="flex w-80 max-w-[calc(100vw-2rem)] flex-col gap-y-1 rounded-lg bg-white p-1 shadow-xl">
+      <div className="px-3 py-1.5 text-sm font-medium text-neutral-500">
+        {t('advanced-settings')}
+      </div>
+      <div className="flex items-center justify-between gap-x-2 px-3 py-1">
+        <label
+          className="text-sm font-medium text-neutral-900"
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/panel.tsx]:
```suggestion
          className="body-text-medium text-neutral-900"
```

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/trigger.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <div className="flex shrink-0 items-center gap-x-2">
+      {slippage !== undefined && (
+        <span
+          className={`whitespace-nowrap text-xs font-medium ${labelStyles[level]}`}
+        >
+          {t(labelKeys[level], { value: slippage })}
+        </span>
+      )}
+      <button
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/trigger.tsx]: Don't we have a `ButtonIcon` component for this kind of usage?

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippageBps.ts
```diff
@@ -0,0 +1,15 @@
...
+import { usePoolForm } from '../_context/poolFormContext'
+import { type PoolOperation } from '../_types/operations'
+
+export const getDefaultSlippage = (operation: PoolOperation) =>
+  operation === 'deposit' ? defaultDepositSlippage : defaultRedeemSlippage
+
+export const useSlippageBps = function (operation: PoolOperation) {
+  const { slippage } = usePoolForm()
+  return percentToBps(clampSlippage(slippage ?? getDefaultSlippage(operation)))
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippageBps.ts:15]: This is creating an entire PoolForm state just to hold the slippage. A regular useState should be enough.

portal/app/[locale]/tunnel/_components/erc20TokenApproval.tsx
```diff
@@ -48,7 +19,7 @@ export const Erc20TokenApproval = function ({
     <div
       className={`text-sm ${
         disabled ? 'cursor-not-allowed' : ''
-      } flex items-center gap-x-2 font-medium text-neutral-950`}
+      } flex items-center justify-between gap-x-2 font-medium text-neutral-950`}
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/erc20TokenApproval.tsx:22]: Just to confirm - is this changing the UI of the tunnel where it was used? IF so, let's post a screenshot and verify it's actually changing. If it "looks the same", then let's confirm why are we adding this change 😅

portal/components/customTokenDrawer/acknowledge.tsx
```diff
@@ -11,33 +12,14 @@ export const Acknowledge = function ({ acknowledged, onChange }: Props) {
...
-            viewBox="0 0 12 12"
-          >
-            <path
-              d="M2 6.5l2.5 2.5L10 3"
-              strokeLinecap="round"
-              strokeLinejoin="round"
-            />
-          </svg>
-        </div>
-      )}
```
gndelia, line 40: Can we double check this looks still correct?

portal/messages/en.json
```diff
@@ -294,6 +294,22 @@
...
+        "approve-extra-withdrawal": "Approve {multiplier}x withdrawal",
+        "auto": "Auto",
+        "continue-anyway": "Continue anyway",
+        "high-slippage": "High slippage",
+        "high-slippage-value": "High slippage {value}%",
+        "max-slippage": "Max slippage",
+        "slippage-value": "Slippage {value}%",
+        "very-high-slippage": "Very high slippage",
+        "very-high-slippage-value": "Very high slippage {value}%",
+        "warning-description": "<highlight>Your slippage is set above {threshold}%.</highlight> This increases your exposure to price impact. You may receive significantly fewer tokens than quoted.",
```
gndelia: I wonder if we should use `shares` and `tokens` depending on whether the operation is a deposit or a withdrawal. Not really important though, just throwing it out 

```suggestion
        "warning-description": "<highlight>Your slippage is set above {threshold}%.</highlight> This increases your exposure to price impact. You may receive significantly fewer shares than quoted.",
```

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx:60]: This retry runs from the transactions history drawer on the earn home page (`TransactionsSection` in `page.tsx`), which sits outside `PoolFormProvider`, so it always falls back to the default slippage with no way to widen it. The PR's motivation is that "when a min-out failure happened there was no way to widen it and retry" — but a failure the user comes back to via the history drawer is exactly the case that still can't widen. Is carrying the custom slippage (or exposing the gear) here intended to be out of scope for this PR? Same applies to the mirrored path in `retryFailedWithdraw.tsx` (line 58).

— Gabi 🤖

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/trigger.tsx
```diff
@@ -0,0 +1,77 @@
...
+        aria-label={t('advanced-settings')}
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/trigger.tsx:61]: The gear toggles a custom popover but only exposes `aria-label`. Passing `isOpen` down and setting `aria-expanded={isOpen}` (plus `aria-haspopup`) would let assistive tech announce the open/closed state.

— Auto Gabi

packages/hemi-earn-actions/src/actions/wallet/requestDeposit.ts
ArturDolzan: Fair — dropped in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de).

packages/hemi-earn-actions/src/actions/wallet/requestRedeem.ts
ArturDolzan: Same, dropped in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de).

packages/hemi-earn-actions/src/utils.ts
ArturDolzan: Good call on throwing — done in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de). Worth noting the portal never hits it: it passes `undefined` when the toggle is off, and 10x the amount when it's on. It's there for the dust-snap case, where a redeem transfers more shares than were requested.

portal/app/[locale]/hemi-earn/_constants/slippage.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_constants/slippage.ts]: You're right, and it's the same question as allowing 100%. Removed the clamp in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de) — 0% is accepted now and gets an inline "Slippage too low" warning instead. No dialog for that one though: too little tolerance reverts the request, while too much is what silently costs funds, so only the upper end asks for an acknowledgement.

Writing it also exposed a bug in my own threshold — the input mask only allowed one decimal, so the only value below 0.1% anyone could type was zero and the whole band was unreachable. It takes two decimals now, which `percentToBps` maps exactly and which makes 0.25% expressible too.

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx:60]: Yes, it is — good catch. Those two components live outside `PoolFormProvider`, so they can't read the setting; the retries inside the pool page do honor it. Fixing it means moving the preference somewhere both reach, so I opened #2183 rather than growing this PR further.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/panel.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/panel.tsx]: Didn't know about that one — applied in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de), and to the other four spots where I'd hardcoded the same pair.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/trigger.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/trigger.tsx]: We do, and it fits now — see the gear comment. I'd hand-rolled it because I was colouring the whole icon per severity and `ButtonProps` omits `className`; once the gear went back to neutral that reason disappeared. It's a plain `ButtonIcon` in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de), with the badge as a sibling so the design system's `opacity-70` on the icon doesn't mute it.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippageBps.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippageBps.ts:15]: It has to be shared, unfortunately — besides the panel, six other places read it: both preview hooks, both review drawers and both in-pool retries. With local state the value wouldn't leave the panel and the signed min-out would still be the default.

portal/messages/en.json
ArturDolzan: Nice one — split the copy per operation in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de), shares for deposits and tokens for withdrawals, in all three locales.

portal/app/[locale]/tunnel/_components/erc20TokenApproval.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/tunnel/_components/erc20TokenApproval.tsx:22]: Good instinct asking — it does touch it. Extracting the toggle also surfaced a latent bug: the knob was centred with a top margin inside an inline `<label>`, so it drifted with whatever line-height the consumer inherited. The tunnel's `text-sm` happened to make it land right; Storybook, with no text class, showed it off-centre. It's flex-centred in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de) and the inner `<label>` (which wrapped only the input and named nothing) is gone.

Net effect in the tunnel should be none — screenshot below. One invisible change though: Enter with the toggle focused no longer submits the form, because in Hemi Earn the panel sits inside the pool form and that would fire the deposit.

Captura de Tela 2026-08-11 às 10 55 30

portal/components/customTokenDrawer/acknowledge.tsx
ArturDolzan: Checked it. The only visual change is a hover state on the checkbox, from the Hemi Earn Figma — and it lands on `CallToActionContainer`, which is already `bg-neutral-50`, the exact colour the hover paints. So it's invisible in this drawer; you only see it where the background is white, like the settings panel. Screenshot below.

I'd also swapped `focus:ring` for `focus-visible:ring` at one point — no design backing, so that's reverted and the focus behaviour is byte-identical to main.

Captura de Tela 2026-08-11 às 11 05 03
Captura de Tela 2026-08-11 às 11 06 14

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2182#issuecomment-5247430243):
> **The gear icon looks different from Figma**

Good catch — I was colouring the whole icon per severity, but Figma keeps the gear at `#737373` in every state and lets the badge carry it. Fixed in [`4b0258de`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/4b0258de): the gear stays neutral and the badge sits outside the icon so the design system's `opacity-70` on icon buttons doesn't mute it.

That also removed my reason for hand-rolling the button, so it's a plain `ButtonIcon` now — answering your other comment too.

New video linked in the description.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/trigger.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_components/advancedSettings/trigger.tsx:61]: Good catch — added both in [`75b70806`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/75b70806). Went with `aria-haspopup="dialog"` rather than `"true"`, since the latter maps to menu semantics and the panel is a popover with form controls.

portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/transactionsSection/transactionDrawer/retryFailedDeposit.tsx:60]: Right, and out of scope on purpose. Both drawers sit outside `PoolFormProvider`, so they have no way to read the setting — carrying it there means moving the preference to a scope both paths reach, which is more than this PR should grow. Tracked in #2183, including the mirrored path in `retryFailedWithdraw.tsx`.

@gndelia raised the same thing in [the other thread](https://github.com/hemilabs/ui-monorepo/pull/2182#discussion_r3754307249), for reference.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippageBps.ts
Comment (gndelia) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippageBps.ts:15]: Ah, I see. I confused `usePoolForm` with `usePoolFormState` - I did not notice it was a context.

portal/app/[locale]/hemi-earn/_utils/approval.ts
```diff
@@ -0,0 +1,13 @@
...
+
+// undefined lets the action approve exactly what it pulls, which for a redeem can
+// exceed the requested amount once the dust snap rounds it up to the full balance.
+export const getExtraApprovalAmount = (amount: bigint, enabled: boolean) =>
+  enabled ? amount * BigInt(extraApprovalMultiplier) : undefined
+
+// The allowance the approval tx will request, for gas estimation — which needs a
+// concrete number rather than the action's "approve whatever you pull" default.
+export const getApprovalAmount = (amount: bigint, enabled: boolean) =>
+  getExtraApprovalAmount(amount, enabled) ?? amount
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_utils/approval.ts:15]: Nit: `getExtraApprovalAmount` returns `BigInt(0)` — not `undefined` — when the toggle is on and the amount is `0n` (there's even a test asserting that), so this `??` won't fall back, since `0n` isn't nullish. The gas estimate would then be built against an allowance of `0`.

It's unreachable today because `useEstimateApproveErc20Fees` is gated on `needsApproval`, which is false at amount 0, but `??` on a bigint is a trap if that gate ever loosens. `|| amount` would be safe here regardless.

gabmontes (dismissed):

portal/app/[locale]/hemi-earn/_utils/approval.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/approval.ts:15]: Good catch, and the `??`-on-bigint framing is the right one — I'd waved this away as harmless earlier precisely because the values coincide at 0.

Fixed at the source rather than the call site in [`2488c12c`](https://github.com/hemilabs/ui-monorepo/pull/2182/commits/2488c12c): `getExtraApprovalAmount` now returns `undefined` when there's no extra allowance to request, whether that's because the toggle is off or the amount is zero. That clears the same trap in `deposit.tsx` and `withdraw.tsx`, which both do `extraApprovalAmount ?? amount` too.

gabmontes (approved):

portal/app/[locale]/hemi-earn/_utils/approval.ts
```diff
@@ -0,0 +1,15 @@
...
+// exceed the requested amount once the dust snap rounds it up to the full balance.
+export const getExtraApprovalAmount = (amount: bigint, enabled: boolean) =>
+  enabled && amount > BigInt(0)
+    ? amount * BigInt(extraApprovalMultiplier)
+    : undefined
+
+// The allowance the approval tx will request, for gas estimation — which needs a
+// concrete number rather than the action's "approve whatever you pull" default.
+export const getApprovalAmount = (amount: bigint, enabled: boolean) =>
+  getExtraApprovalAmount(amount, enabled) ?? amount
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_utils/approval.ts:15]: This functionality also works on the Tunnel: after all they are both the same scenario: They let the user 10x approve (or not!) a spender for a given token

So far it's only enabled on the tunnel (Besides here), but perhaps we could add the 10x control to all approvals in the app: Staking / staking governance could use them as well.

Perhaps we could create an issue to unify these snippets of code for the 10x approval

gndelia (approved):

portal/app/[locale]/hemi-earn/_utils/approval.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_utils/approval.ts:15]: Good idea — thanks, opened #2185 for it.

Scoped it to the tunnel and Hemi Earn, since the interesting part turned out not to be the shared arithmetic but the four places the two have drifted: tooltip or no tooltip, hardcoded `10x` versus an interpolated multiplier, whether the toggle disables when no approval is needed, and deposit-only versus per-operation copy. Wrote those up as the decisions to settle.

Left staking out on purpose — `useStake` already passes `approvalAmount`, but there is no UI for it and no product call that those screens should have one, so it felt like a feature rather than a refactor. Worth revisiting once a single component exists.

Comment (gndelia) [reply portal/app/[locale]/hemi-earn/_utils/approval.ts:15]: > but there is no UI for it and no product call that those screens should have one

Right, I am making the product call that those pages should have it lol.

PR #2186, opened 2026-08-12
Fix earn retry ignoring the configured slippage

### LINKED ISSUE #2183: Hemi Earn: retry from the transactions drawer ignores the slippage the user set
Spotted while reviewing #2182, and narrowed down after looking at which failures actually reach each button.

## What

`retryFailedDeposit.tsx` and `retryFailedWithdraw.tsx` — the "Try again" in the transactions drawer on the Hemi Earn home page — recompute the min-out from `defaultDepositSlippage` / `defaultRedeemSlippage` instead of the tolerance the user configured. The retries inside the pool page (`poolReview/retryDeposit.tsx`, `poolReview/retryWithdraw.tsx`) do honor it, so the same button label behaves differently depending on where it is clicked.

Those are the only two call sites: everything else that derives a min-out goes through `useSlippageBps`, and the Agent retry/cancel path (`useRemoteFailedActions`) does not touch slippage at all — it acts on the request already on chain.

## What this is NOT

**We do not need to expose the settings gear in the drawer.** There is no failure a user can reach from there where changing the slippage would help:

- A slippage failure is remote (the Agent reverted on Ethereum). `RemoteFailedCta` deliberately offers **only Cancel** for `category === slippage`, because retrying just slips again against the frozen min-out. The user is sent back to the form, where the gear already lives.
- The "Try again" in question is gated by `canRetryRow`, i.e. `FAILED && isLocalEarnTransactionRow` — the Hemi-side transaction reverted or the signature was rejected. The request never reached the Agent, so the min-out was never evaluated and slippage cannot have been the cause (gas, rejection, balance and nonce can).

So the fix is to **carry the configured value through**, not to let it be edited there.

## Why it still matters

Consistency: the user signed at 8%, so a retry should re-send at 8%. Today it silently drops to 0.5%/1% with nothing on screen saying so. It will not unblock a failure — it just stops the retry from being a different transaction than the one it claims to repeat.

## Why it was left out of #2182

Both components sit outside `PoolFormProvider`, which is where the preference lives, so they cannot read it. Carrying it means moving the preference to a scope both paths reach — a context at the `hemi-earn` level, or persisting it — and that PR was already large enough that the reviewer flagged its size.

## Correction to the first version of this issue

It claimed this contradicts the `subheading-slippage` copy ("try the withdrawal again with a higher slippage tolerance"). That was wrong: that copy belongs to the remote-failed path, which has no Retry button. The two are unrelated.

ArturDolzan:
### Description

This PR fixes the retry in the Hemi Earn transactions drawer ignoring the slippage the user set, split into two commits.

The slippage control we shipped in #2182 lives in the pool form provider, which is only mounted on the pool page. The two "Try again" buttons in the home transactions drawer sit outside it, so they recomputed the min-out from the default constants. Same button, same label, different tolerance depending on where you clicked it — sign at 8%, retry at 0.5%.

Rather than moving the preference to a wider scope, the value is now recorded on the local entry at sign time. A shared preference would replay the *current* value, not what was signed: `resetSettings` fires on every tab switch, and the user may well have touched the form since the failure. Storing it on the request replays exactly what went out, and survives a reload. It follows the same path `approvalTxHash` already takes — a field only the local store knows about, exposed on `EarnTransaction`, read by the retry components.

Worth being explicit that **this unblocks nothing on its own**. Slippage is enforced remotely by the Agent, and that path deliberately offers only Cancel (a slippage retry just slips again against the frozen min-out). The "Try again" in question is gated by `canRetryRow`, so it only ever reaches local failures where the min-out was never evaluated. This is about a retry repeating the transaction it says it repeats.

Broken down into:
- `df7631de` — Record the slippage earn requests are signed with (plumbing only, no behavior change)
- `9a66faa9` — Reuse the recorded slippage when retrying (the fix)

### Screenshots

N/A - No UI changes.

### Related issue(s)

Closes #2183 

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippage.ts
```diff
@@ -9,7 +9,10 @@ import { type PoolOperation } from '../_types/operations'
...
   operation === 'deposit' ? defaultDepositSlippage : defaultRedeemSlippage
 
-export const useSlippageBps = function (operation: PoolOperation) {
+export const useSlippage = function (operation: PoolOperation) {
   const { slippage } = usePoolForm()
-  return percentToBps(clampSlippage(slippage ?? getDefaultSlippage(operation)))
+  return clampSlippage(slippage ?? getDefaultSlippage(operation))
 }
+
+export const useSlippageBps = (operation: PoolOperation) =>
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippage.ts:17]: Minor / non-blocking: after this change `useSlippage` is the primary hook here (used by deposit/withdraw and both pool retries), and `useSlippageBps` is a thin wrapper used in only two places. The file is still named `useSlippageBps.ts`, which now points at the secondary export. Consider renaming to `useSlippage.ts` for discoverability — feel free to ignore if you'd rather avoid the import churn.

— Gabi 🤖

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/_hooks/useSlippage.ts:17]: Good point, the name was pointing at the wrong export after this change. Renamed to `useSlippage.ts` in [`a1f13cdf`](https://github.com/hemilabs/ui-monorepo/pull/2186/commits/a1f13cdf), covering the 4 imports this PR already touched plus the 3 that were untouched. The module itself is byte for byte the same, so git tracked it as a 100% rename. Thanks!

gabmontes (approved):

gndelia (approved):

PR #2187, opened 2026-08-12
Sort earn composition items by share

### LINKED ISSUE #2158: Hemi Earn: Update the composition table order on the pool detail from higgest % to lowest %
As the titles says, we need to set the default order for this table (both for token and protocol) from higgest % to lowest %

Image
Image

ArturDolzan:
### Description

This PR sorts the Hemi Earn composition breakdown from the largest slice to the smallest.

Until now items came out in fetch order, per token and per strategy within each token, so the biggest slice could show up anywhere in the legend. Two pools with the same composition could read completely differently depending on how the data happened to arrive. Sorting descending is what a breakdown is expected to do.

The reserve buffer stays pinned at the end, below the sorted slices. It isn't a position, it's leftover idle funds, so ranking it by size could drop it between two strategies and made the grouping shift around as the buffer grew (thanks @gabmontes for catching it). This only affects the protocol view, since the token view folds idle funds into each token and has no buffer row.

Test fixtures were reworked so the ordering is actually observable (the old amounts happened to already be descending), plus two new cases: one covering both view modes, one checking the buffer stays last even when it outweighs every strategy.

### Screenshots

Captura de Tela 2026-08-12 às 13 47 05
Captura de Tela 2026-08-12 às 13 47 13

### Related issue(s)

Closes #2158 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/hemi-earn/_fetchers/fetchComposition.ts
+++ b/portal/app/[locale]/hemi-earn/_fetchers/fetchComposition.ts
@@ -68,7 +68,7 @@ const toCompositionData = function (
   return data
 }
 
-// 'protocol' groups by strategy and appends idle funds (withdrawable − totalDebt)
+// 'protocol' groups by strategy and adds idle funds (withdrawable − totalDebt)
 // as a reserve buffer; 'token' groups by token (idle already included).
 export const toCompositionItems = function ({
   data,
@@ -104,7 +104,9 @@ export const toCompositionItems = function ({
     }
   }
 
-  const visible = items.filter(item => item.amount > 0)
+  const visible = items
+    .filter(item => item.amount > 0)
+    .sort((a, b) => b.amount - a.amount)
   const total = visible.reduce((sum, item) => sum + item.amount, 0)
   return visible.map(item => ({
     amount: item.amount,
--- a/portal/app/[locale]/hemi-earn/_fetchers/fetchComposition.ts
+++ b/portal/app/[locale]/hemi-earn/_fetchers/fetchComposition.ts
@@ -68,7 +68,7 @@ const toCompositionData = function (
   return data
 }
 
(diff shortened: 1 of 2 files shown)


gndelia (dismissed):

portal/app/[locale]/hemi-earn/_fetchers/fetchComposition.ts
```diff
@@ -104,7 +104,9 @@ export const toCompositionItems = function ({
     }
   }
 
-  const visible = items.filter(item => item.amount > 0)
+  const visible = items
+    .filter(item => item.amount > 0)
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_fetchers/fetchComposition.ts]: The reserve buffer isn't a real position in the breakdown — it's leftover idle funds — so ranking it by size alongside strategies can push it into the middle of the legend, which reads oddly (a "buffer" sitting between two strategies). Suggest keeping the buffer pinned at the end and sorting only the actual slices, e.g. sort the strategy/token items descending before appending the reserve buffer, rather than sorting the combined list. That keeps the visual grouping stable regardless of buffer size.

— Gabi 🤖

gabmontes (changes requested):

portal/app/[locale]/hemi-earn/_fetchers/fetchComposition.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchComposition.ts]: Makes sense, the buffer isn't a position so grouping it apart reads better. Sorted only the real slices and pinned it at the end in [`6a9810d2`](https://github.com/hemilabs/ui-monorepo/pull/2187/commits/6a9810d2), and flipped the test so it now covers the case you described (a buffer bigger than every strategy still staying last). Updated the PR description too, since it argued the opposite. Thanks!

gndelia (approved):

gabmontes (approved):

PR #2188, opened 2026-08-12
Add keeper subcommand to hemi-earn sandbox

ArturDolzan:
### Description

This PR adds a `keeper` subcommand to the hemi-earn sandbox scripts so the "keeper wins the race" branch of the portal UI becomes testable locally.

The portal has two paths that only fire when a production keeper resolves a request before the user clicks the recovery CTA: a REMOTE_FAILED request whose `failedRequests[id].tokenIn` has gone to zero, and a mid-cooldown redeem cancelled cross-chain via `Agent.cancel`. There was no way to trigger those Agent-side actions from the ui-monorepo sandbox, so the corresponding UI branches were only exercised on real deployments.

The new subcommand dispatches `Agent.cancel` or `Agent.retry` by request id. For cancel, it probes `failedRequests` then `unstakeRequests` (matching the production `Agent.cancel` order) and picks the branch automatically, or bails out with a diagnostic if the request is in neither map. For retry, it requires the entry to exist in `failedRequests` and reads it back after the tx to report whether the entry cleared or the underlying failure still holds. Both actions guard on `receipt.status` so a reverting tx surfaces as an error rather than a false success.

Wired into the dispatcher alongside the other subcommands and documented under a new "Keeper actions" section in the sandbox README, with invocation examples for the failed branch, unstake branch, and retry.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2092 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/scripts/hemi-earn/README.md
+++ b/portal/scripts/hemi-earn/README.md
@@ -50,13 +50,14 @@ All sandbox actions are dispatched through a single pnpm script that forwards it
 pnpm --filter portal sandbox:hemi-earn -- <subcommand> [flags]
 ```
 
-| Subcommand     | Purpose                                                                                                                  |
-| -------------- | ------------------------------------------------------------------------------------------------------------------------ |
-| `setup`        | Start Anvil + deploy mocks + fund the test account.                                                                      |
-| `mint`         | Mint from any ERC20-mock — top up an EOA or inject yield into the vault (see [Minting](#minting)).                       |
-| `mining`       | Toggle Anvil's interval mining at runtime (see [Slow mining](#slow-mining)).                                             |
-| `relayer`      | Emulate the production keeper: claim mature cooldown redeems and bridge cancellation requests (see [Relayer](#relayer)). |
-| `fail-gateway` | Toggle `PreviewableGatewayMock` into a failure mode (see [Failure simulation](#failure-simulation)).                     |
+| Subcommand     | Purpose                                                                                                                      |
+| -------------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| `setup`        | Start Anvil + deploy mocks + fund the test account.                                                                          |
+| `mint`         | Mint from any ERC20-mock — top up an EOA or inject yield into the vault (see [Minting](#minting)).                           |
+| `mining`       | Toggle Anvil's interval mining at runtime (see [Slow mining](#slow-mining)).                                                 |
+| `relayer`      | Emulate the production keeper: claim mature cooldown redeems and bridge cancellation requests (see [Relayer](#relayer)).     |
+| `fail-gateway` | Toggle `PreviewableGatewayMock` into a failure mode (see [Failure simulation](#failure-simulation)).                         |
+| `keeper`       | Fire `Agent.cancel` / `Agent.retry` to simulate the "keeper wins the race" scenario (see [Keeper actions](#keeper-actions)). |
 
 Building blocks used by `setup` (`deployMocks.ts`, `fundAccount.ts`) are still invocable directly for advanced cases:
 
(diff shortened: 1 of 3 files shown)


portal/scripts/hemi-earn/README.md
```diff
@@ -170,6 +172,37 @@ pnpm --filter portal sandbox:hemi-earn -- fail-gateway --kind deposit --mode off
...
 ```
 
+## Keeper actions
+
+The production keeper can resolve a REMOTE_FAILED request (or a mid-cooldown redeem) before the user clicks the recovery CTA in the portal — the "keeper wins the race" path. When that happens on-chain, `Agent.failedRequests[id].tokenIn` (or `Agent.unstakeRequests[id].share`) goes to zero and the portal must hide its recovery CTAs on the next refetch. The `keeper` subcommand fires those Agent-side actions manually so that branch of the UI is testable end-to-end.
+
+Two actions, one dispatch:
+
+- `--action cancel` — pre-flight probes `failedRequests` then `unstakeRequests` (matches the production `Agent.cancel` order) and picks the branch automatically. Errors out with a diagnostic if the request is in neither map.
+- `--action retry` — pre-flight requires the request to be in `failedRequests`; post-flight re-reads and reports whether the entry cleared (success) or is still populated (underlying failure still active — clear it via `fail-gateway --mode off` first).
```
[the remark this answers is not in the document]
ArturDolzan: Good catch, fixed in [`7e4d4ce6`](https://github.com/hemilabs/ui-monorepo/pull/2188/commits/7e4d4ce6).

gndelia (dismissed):
uACK

portal/scripts/hemi-earn/keeper.ts
```diff
@@ -0,0 +1,319 @@
...
+        '  Expected either a REMOTE_FAILED request (via `fail-gateway` + submit) or a mid-cooldown redeem.',
+    )
+  }
+
+  const hash = await walletClient.writeContract({
+    abi: [agentCancelAbi],
+    account,
+    address: agent,
+    args: [requestId],
+    chain: walletClient.chain,
```
gabmontes, line 202: `Agent.cancel` is `payable` and the sandbox's own `ToggleableAgent` gates it behind `MOCK_CANCEL_FEE` when `strictFeeCheck` is on (reverts `InsufficientFee` on a `MSG_REQUEST_CANCEL` with `nativeFee < MOCK_CANCEL_FEE`). But `--value` is parsed and then dropped for `cancel` (README says "retry only"), so under `strictFeeCheck = true` — the REMOTE_FAILED path this command targets — `keeper --action cancel` always reverts with no way to fund it. `retry` already forwards `--value` for the symmetric `MOCK_CALLBACK_FEE`; suggest forwarding `value` to the cancel tx too and updating the docs. Was cancel's fee intentionally left unfundable, or is this an oversight?

— Gabi 🤖

gabmontes (approved):

gabmontes (commented):
Dismiss the prev approval 😓

gabmontes (commented):
Y said... dismiss!

gabmontes (changes requested):
Ok. It didn't work. Let's see if requesting changes dismiss the prev approval.

portal/scripts/hemi-earn/keeper.ts
ArturDolzan: Oversight, thanks for catching. Forwarded `--value` to the cancel tx too (symmetric to retry) and reworded the flag hint / README in [`7755243d`](https://github.com/hemilabs/ui-monorepo/pull/2188/commits/7755243d).

gndelia (approved):

gabmontes (approved):

Issue #2194, opened 2026-08-13
Migrate the portal from Next.js to Vite and deploy it on Cloudflare


ArturDolzan:
### Feature or change description

Migrate the portal from Next.js (static export) to a plain React SPA built with Vite, served from Cloudflare Workers instead of Hostinger — mirroring the setup already running in [`vetro-monorepo/web`](https://github.com/hemilabs/vetro-monorepo/tree/master/web).

The end state is the same shape as `web/`, minus the `src/` nesting: the portal keeps its flat layout, so these files sit at the package root.

```
portal/
  index.html
  vite.config.ts
  wrangler.jsonc
  main.tsx          # entry: providers + createRoot
  app.tsx           # react-router route tree
  worker.ts         # Cloudflare Worker: security headers + per-request CSP nonce
  pages/            # one file per route
  components/ context/ hooks/ i18n/ networks/ utils/ ...
```

### Problem or use case addressed

**We are paying for a framework we do not use.** The portal is a fully static export (`output: 'export'`) with no server rendering, no server action and no route handler. Nothing survives `next build` other than static HTML, and in exchange we carry Next's build/config surface, a Next-specific Storybook framework, a Next-specific Sentry SDK, and an i18n layer built around request-scoped server config we never run on a server.

**The deploy path is the most fragile part of the repo.** `hostinger-deployment.yml` rsyncs `out/` over SSH to a shared host, and every security header is emitted by `portal/scripts/generateServerConfig.js` into an `.htaccess`. That script hardcodes the CSP allow-list because it could not import from viem/hemi-viem (ESM vs commonjs), so it silently drifts from `portal/networks/*`. On Cloudflare the Worker is bundled by Vite and can import `networks` directly, the way `web/src/index.ts` does.

**No preview deployments.** #1563 asked for exactly this and proposed Cloudflare as the answer. `wrangler.jsonc` gives it via `preview_urls`.

### Proposed implementation

Sequenced so each phase is reviewable on its own, as separate PRs against the migration branch — see [Rollout](#rollout) for how that branch is run. The app does not boot until phase 3 or so is done; that is expected, not a sign the split is wrong.

#### 1. Build tooling

- [x] `next@16` → `vite@8` + `@vitejs/plugin-react`; scripts become `dev` / `build` (`vite build`) / `preview`. `tsc` is deliberately **not** part of `build`: it lives in its own `typecheck` script, so a type error cannot stop the branch from producing a `dist/`.
[...]

PR #2197, opened 2026-08-14
Decouple evmApi from the wallet context

ArturDolzan:
### Description

This PR removes the dependency between `utils/evmApi` and the EVM wallet context, which is the first step of the portal's Vite migration.

Reading a block or a transaction receipt only needs a chain and a transport, but `evmApi` was pulling the wagmi `Config` out of `context/evmWalletContext`. That dragged RainbowKit, wagmi connectors and `react-device-detect` into every web worker that imports it, and those touch `window` at module scope. Turbopack happens to tree-shake all of it away today (I checked the current export, and there is no wallet code anywhere in the worker chunks), so production is fine, but Rollup does not, and under Vite three of the five tunnel workers die on startup.

The replacement is `getPublicClient` from `utils/chainClients`, which already builds a plain viem client with no wallet dependency and is already used by `utils/watch/bitcoinWithdrawals.ts`. The transport is identical either way, since wagmi's config was built from the same `buildTransport` per chain. Clients are memoized per chain because `buildTransport` wires an `eth-rpc-cache` whose state lives inside the transport, so creating a fresh client on every call would never hit the cache. `allEvmNetworksWalletConfig` is no longer exported, since `evmApi` was its only consumer outside the context.

The three existing `utils/watch` test files mock `utils/evmApi` wholesale, so the module had no coverage of its own. The new tests cover the parts that actually hold logic, and I verified by mutation that they fail when the memoization or the memo key is broken.

Broken down into:
- `1086a98` - Use a wallet-free public client in evmApi
- `a242c52` - Add tests for evmApi

### Screenshots

N/A - No UI changes

### Related issue(s)

Related to #2194 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/utils/evmApi.ts
+++ b/portal/utils/evmApi.ts
@@ -1,17 +1,24 @@
-import {
-  getBlock,
-  getTransactionReceipt as wagmiGetTransactionReceipt,
-} from '@wagmi/core'
-import { allEvmNetworksWalletConfig } from 'context/evmWalletContext'
 import pMemoize from 'promise-mem'
+import { getPublicClient } from 'utils/chainClients'
 import { type Chain, type Hash, type TransactionReceipt } from 'viem'
+
+const clients = new Map<Chain['id'], ReturnType<typeof getPublicClient>>()
+
+// buildTransport() wires an eth-rpc-cache whose state lives inside the
+// transport, so a fresh client per call would never hit the cache.
+const getClient = function (chainId: Chain['id']) {
+  const existing = clients.get(chainId)
+  if (existing) {
+    return existing
+  }
+  const client = getPublicClient(chainId)
+  clients.set(chainId, client)
+  return client
+}
 
 export const getEvmBlock = pMemoize(
   (blockNumber: bigint | number, chainId: Chain['id']) =>
-    getBlock(allEvmNetworksWalletConfig, {
-      blockNumber: BigInt(blockNumber),
-      chainId,
-    }),
+    getClient(chainId).getBlock({ blockNumber: BigInt(blockNumber) }),
   { resolver: (blockNumber, chainId) => `${blockNumber}-${chainId}` },
 )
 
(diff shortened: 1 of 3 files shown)


gndelia (commented):
I think this change can land on `main` as it's not VITE specific - right? Perhaps doing so is better to reduce the amount of changes that will be on hold, simplifying future rebases

portal/utils/evmApi.ts
```diff
@@ -1,31 +1,37 @@
...
+  return client
+}
+
 export const getEvmBlock = pMemoize(
   (blockNumber: bigint | number, chainId: Chain['id']) =>
-    getBlock(allEvmNetworksWalletConfig, {
-      blockNumber: BigInt(blockNumber),
-      chainId,
-    }),
+    getClient(chainId).getBlock({ blockNumber: BigInt(blockNumber) }),
```
gndelia: I think, in general, I prefer using the form


```suggestion
    getBlock(getClient(chainId), { blockNumber: BigInt(blockNumber) })
```

Importing `getBlock` from `viem/actions`. I thought it was part of CLAUDE.md, but it seems it's not.

Reasoning: Better typings (methods coming out from `client` sometimes are oddly typed), and this one allows tree-shaking. Same reasoning applies to `getTransactionReceipt` below

ArturDolzan: Makes sense, switched both to the `viem/actions` form in [`b879e91`](https://github.com/hemilabs/ui-monorepo/pull/2197/commits/b879e91da4249568ccfa3a48040c97bcc4e2a373).

One note on the tree-shaking part: `getPublicClient` builds the client with `createPublicClient`, which does `client.extend(publicActions)`, so all the public actions land in the bundle either way. The typing win is real though, it actually bit me while writing the test for this.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2197#pullrequestreview-4940815693):
> I think this change can land on `main` as it's not VITE specific - right? Perhaps doing so is better to reduce the amount of changes that will be on hold, simplifying future rebases

You're right that it isn't Vite specific. I'd lean toward keeping it on `vite-migration` anyway, it's a small change and I don't think it moves the needle much on the rebase side. It does change how the RPC reads work at runtime, so I'd rather it ride along with the rest and get validated on the staging env together. If you feel strongly about it, happy to retarget to `main`, no problem at all.

gndelia (approved):

gabmontes (approved):

PR #2202, opened 2026-08-18
Build the portal with Vite

ArturDolzan:
### Description

This PR swaps the portal's build from Next to Vite, and deliberately stops there.
 
The portal has been a static export for a while now (179 files marked `use client`, nothing server-side), so Next was mostly build surface we paid for without using. The swap itself is small. What took the thinking was deciding what to leave out.

**The app is a placeholder on purpose.** `main.tsx` renders a single div, so `pnpm build` transforms 17 modules. Nothing under `app/` is reachable from the Vite entry yet, and wiring it up is the routing step that comes next. The goal here is narrower: produce a real `dist/` so the Cloudflare environment can exist and start receiving every push, and we validate the migration there as it lands instead of all at once at the end. `portal/wrangler.jsonc` is assets only, no Worker, which keeps the output shape stable so that environment does not need reconfiguring when the headers Worker arrives later.

**Next stays installed.** It builds nothing now, but it still feeds types to the ~800 files that import it. Removing it today would turn `tsc` and `knip` red across the whole portal for weeks and cost us both signals during the migration. It goes away with the last import.

A few changes came along because the toolchain forced them: the ESLint preset swap (`bloq/next` is what pulls in react-hooks and jsx-a11y today, so `bloq/react` replaces it), a `tsconfig.build.json` so the type gate covers app code without breaking path resolution in tests, and some pre-existing violations the new presets started seeing. One of those was a real bug, `clip-rule` and `fill-rule` in an SVG that React was silently ignoring.

  ## Notes

Two things are switched off on purpose, both with the reason written next to them in the code. `verbatimModuleSyntax` would mean marking type-only imports across ~640 files, unrelated to the bundler and better done once those files are in their final shape. Seven `jsx-a11y` rules that `eslint-config-next` never enabled are off too, so accessibility gets its own pass when there is an app to test against.

Worth flagging for whoever writes the headers Worker: vetro's CSP sets `worker-src` to `'none'` because it has no web workers. The portal has five, so it needs `'self'` or tunnel history stops syncing with no visible error.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2194

### Checklist

- [x] Manual testing passed.


gndelia:
> Removing it today would turn tsc and knip red across the whole portal for weeks and cost us both signals during the migration.

FYI I recommend ignoring the CI checks for this migration. Keeping the CI green in each PR will make your life harder. As you're using a feature branch, you are not constrained by those rules - you can merge with CI broken. After all, that's not getting into staging

portal/package.json
```diff
@@ -85,19 +82,21 @@
...
     "storybook-addon-pseudo-states": "10.4.4",
     "tailwindcss": "3.4.14",
     "typescript": "5.8.3",
-    "vite-tsconfig-paths": "6.1.1",
+    "vite": "8.0.13",
+    "vite-plugin-node-polyfills": "0.28.0",
     "vitest": "4.1.6"
   },
   "engines": {
     "node": ">=20.9.0"
```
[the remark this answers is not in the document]
ArturDolzan: Good catch, done in [`315ebd47`](https://github.com/hemilabs/ui-monorepo/pull/2202/commits/315ebd47). Went with a plain `>=22.12.0` rather than Vite's exact `^20.19.0 || >=22.12.0`, to match how the other workspaces here declare engines. Worth noting `>=20.9.0` was next's own value, so it was already stale before this PR.

portal/Environment Variables.md
```diff
@@ -16,7 +16,7 @@ Any environment variable that can is shared across all environments and is not a
 
 Changes related to security headers - for example, when adding a new external url to fetch from, or download images from.
 
-- portal/scripts/generateServerConfig.js
+The `portal/scripts/generateServerConfig.js` script that emitted the `.htaccess` was removed with the Next build. Its replacement is the Cloudflare Worker, which is added in the deploy step of the migration ([#2194](https://github.com/hemilabs/ui-monorepo/issues/2194)). Until then this branch ships no security headers, which is why it is not deployed anywhere.
```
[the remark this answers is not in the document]
ArturDolzan: You're right, that sentence was written before this branch had a preview to deploy to. Reworded in [`315ebd47`](https://github.com/hemilabs/ui-monorepo/pull/2202/commits/315ebd47) to say it only goes to the Cloudflare migration preview, never to official staging or production.

packages/ve-hemi-actions/actions/wallet/createLock.ts
```diff
@@ -1,7 +1,11 @@
 import { EventEmitter } from 'events'
 import { toPromiseEvent } from 'to-promise-event'
-import type { Address, TransactionReceipt, WalletClient } from 'viem'
-import { encodeFunctionData } from 'viem'
+import {
+  type Address,
+  type TransactionReceipt,
+  type WalletClient,
+  encodeFunctionData,
+} from 'viem'
```
gndelia, line 8: It's not incorrect, but not really needed for the migration. Or is Vite giving you errors for this?

gndelia:
> Worth flagging for whoever writes the headers Worker: vetro's CSP sets

You! 😆

portal/nextImages.d.ts
```diff
@@ -0,0 +1,9 @@
+// Types static image imports as next/image's StaticImageData, which is what
+// the ~60 `<Image src={logo} />` call sites still expect.
+//
+// WARNING: this is the Next shape, not the Vite one. Vite resolves an image
+// import to a plain URL string, so once these files enter the Vite graph the
+// declared type will be a lie: `src.width` is undefined at runtime and next/image
+// throws "missing required width property", with no type error to warn you.
+// Delete this file in the same change that replaces next/image with <img>.
+/// <reference types="next/image-types/global" />
```
gndelia: I'd just drop this. Let the CI fail. It doesn't matter. You wil fix it once Vite is set

portal/vite.config.ts
```diff
@@ -0,0 +1,15 @@
+import react from '@vitejs/plugin-react'
+import { defineConfig } from 'vite'
+import { nodePolyfills } from 'vite-plugin-node-polyfills'
+
+const polyfills = () => nodePolyfills({ include: ['http', 'https', 'util'] })
+
+export default defineConfig({
+  define: { global: 'globalThis' },
```
gndelia, line 12: Do we use `globalThis` anywhere? 🤔

portal/package.json
```diff
@@ -2,18 +2,16 @@
...
   "scripts": {
-    "build": "next build --turbopack",
-    "postbuild": "pnpm run generate:htaccess",
+    "build": "tsc -p tsconfig.build.json && vite build",
     "deploy": "pnpm run build",
-    "dev": "next dev --turbopack",
-    "dev:wifi": "WIFI=true LOCAL_IP=$(ipconfig getifaddr en0) PORT=3000 pnpm run dev",
-    "generate:htaccess": "node ./scripts/generateServerConfig.js",
+    "dev": "vite",
+    "dev:wifi": "vite --host",
```
gndelia, line 8: If we update this script, it means `qrcode-terminal` and  `@types/qrcode-terminal` can be dropped.

After doing so, the script [instrumentation.ts](https://github.com/hemilabs/ui-monorepo/blob/main/portal/instrumentation.ts) can be updated to remove it... unless you want to keep the QR code to be shown. IF that's the case, then that script needs to be reverted to set those 2 variables so the QR appears.

I'm ok with either option - just letting you know so we don't forget about it

portal/tailwind.config.ts
```diff
@@ -5,6 +5,9 @@ const config: Config = {
     './components/**/*.{js,ts,jsx,tsx,mdx}',
     './app/**/*.{js,ts,jsx,tsx,mdx}',
     './pages/**/*.{js,ts,jsx,tsx,mdx}',
+    // Vite's entry files live at the package root
+    './index.html',
```
gndelia: I don't think any style will be included in that file. Is this correct?

```suggestion
```

What do the Tailwind v3 docs recommend doing with that file?

packages/ve-hemi-actions/actions/wallet/createLock.ts
ArturDolzan: Not Vite, it's the lint preset swap in this PR. `no-duplicate-imports` lives in `bloq/esm`, which replaced `bloq/next` in the override that already covered `packages/**`, and lint-staged runs eslint with `--fix --max-warnings 0`, so it rewrites these on any commit that touches the file. Happy to carve them out if you'd rather keep the diff tight, just say the word.

portal/nextImages.d.ts
ArturDolzan: Done in [`ac8aba91`](https://github.com/hemilabs/ui-monorepo/pull/2202/commits/ac8aba91), dropped it. You were more right than I realized: Vite's own types already say `string`, and that file was overriding them with the Next shape, so it was hiding 42 real errors rather than deferring them. I also took `tsc` out of the `build` script, otherwise the build fails and the branch stops producing a dist for the Cloudflare env. `typecheck` is still there as its own script and is red now, which is fine since CI doesn't run it.

portal/vite.config.ts
ArturDolzan: Not our code, and my fault for shipping that line with no comment. It's stream-http and readable-stream (pulled in by the http/https polyfills) reading the bare `global` at module scope, and the plugin injects its shim into the main bundle but not into worker ones. I forced a worker into the graph to measure it: without the define the worker bundle has 12 bare `global` references and zero declarations, including `e.fetch=i(global.fetch)` at module scope. Added the explanation in [`ac8aba91`](https://github.com/hemilabs/ui-monorepo/pull/2202/commits/ac8aba91).

portal/package.json
ArturDolzan: Dropped both, done in [`ac8aba91`](https://github.com/hemilabs/ui-monorepo/pull/2202/commits/ac8aba91). The QR was already dead twice over: `dev:wifi` no longer sets those env vars, and `instrumentation.ts` is a Next hook that Vite never calls. `vite --host` prints the LAN URL anyway. Thanks for catching it before it rotted there.

portal/tailwind.config.ts
ArturDolzan: Removed in [`ac8aba91`](https://github.com/hemilabs/ui-monorepo/pull/2202/commits/ac8aba91). To answer the docs question, the Tailwind v3 Vite guide does list `./index.html` in `content`, but you're right that it buys us nothing here: our index.html has no classes, and the built CSS came out with the identical hash after removing it. Vetro isn't a reference for this one since it's on v4, which has no content array.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2202#issuecomment-5330969270):
> FYI I recommend ignoring the CI checks for this migration. Keeping the CI green in each PR will make your life harder.

Fair, and I took that to heart on the nextImages.d.ts thread: dropped the file and let the 42 type errors show instead of hiding them behind a declaration that was lying about the shape.

One thing worth knowing for the rest of the migration: CI runs format:check, lint, deps:check, build and test, but not typecheck. So the type errors we're choosing to live with don't turn anything red on their own, and `tsc` stays usable as a to-do list for the routing step.

gabmontes (approved):
LGT Auto Gabi and me

gndelia (approved):

PR #2203, opened 2026-08-19
Custom bitcoin receiving address

### LINKED ISSUE #2132: Iterate the 'Receiving Bitcoin Address' container adding the 'Custom address' feature.
We need to implement the newly iterated ''Receiving Bitcoin Address' ' flow, adding the 'Custom address' feature for withdrawals [ From Hemi to Bitcoin ] 

Designs [here](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=20656-8181&t=ZRT2tEzbB6ujJ5jH-1)

Some considerations:

1. Custom address switch only appears on Bitcoin withdrawals.
2. All type of errors from validations regarding the address inputed will we rendered as 'Invalid Address'
3. The default address will always be from the connected Bitcoin wallet, unless a custom address is inputed.
4. CTA validations will remain the same, but we will adding the 'Input a correct custom address' as a last step validation and if the custom address inputed is invalid.
5. When 'Custom Address' switch is active we should focus to the 'Custom Address' input.

arielmergen:
### Description

Adds the "Custom address" toggle to the Bitcoin withdrawal form, so funds can be sent to an address other than the one of the connected Bitcoin wallet.

Notes:

- retryBtcWithdraw.tsx and reviewBtcWithdrawal.tsx read the address from the connected wallet instead of from the withdrawal. Equivalent until now, wrong once the address can be custom.
- The validation checks the network because the contract doesn't: BitcoinKit.isAddressValid only verifies the address converts to a script, so a testnet tb1… is accepted on mainnet. Verified with eth_call.
- bitcoinjs-lib was already in the repo via packages/btc-wallet; it is now declared in the portal because utils/bitcoin.ts imports it directly.

No real withdrawal was run (no Bitcoin wallet available), so the retryBtcWithdraw.tsx change is verified by reading only, and there is no screenshot of the CTA error state.
### Screenshots

01-toggle-off-full

01-toggle-off

03-empty-after-blur

05-testnet-address-on-mainnet

06-broken-checksum

07-garbage

08-while-validating

09-valid-taproot-full

**Testnet valid Address**

10-testnet-valid-address

11-testnet-valid-base58

**Mainnet Address on testnet**

12-mainnet-address-on-testnet-full

12-mainnet-address-on-testnet

**Mainet**

https://github.com/user-attachments/assets/576606fa-2e4d-46d5-a1c2-e99171dee112

**TestNet**

https://github.com/user-attachments/assets/0ea67798-9377-40e8-ba94-b81b44bc5cd2

### Related issue(s)

Closes #2132 
Fixes #
Related to #

### Checklist

- [x] Manual testing passed.


portal/utils/bitcoin.ts
```diff
@@ -36,3 +39,63 @@ export const getBitcoinTimestamp = function (timestamp: number) {
...
+    // Witness v0 is either P2WPKH (20 bytes) or P2WSH (32 bytes), and v1 is
+    // P2TR (32 bytes). Later versions are not defined yet, so for those only
+    // the size range BIP141 allows is checked.
+    if (version === 0) {
+      return data.length === 20 || data.length === 32
+    }
+    if (version === 1) {
+      return data.length === 32
+    }
+    return data.length >= 2 && data.length <= 40
```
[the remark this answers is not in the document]
arielmergen: Fixed in 53da8830. Confirmed `fromBech32` accepts versions 17-31 while `toOutputScript` rejects them. The fallback now caps at 16, with tests for v16, v17 and v31.

portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts
```diff
@@ -0,0 +1,49 @@
...
+
+type Props = {
+  amountValidation: ReturnType<typeof validateSubmit>
+  network: BtcChain['id']
+  walletAddress: string | undefined
+}
+
+/**
+ * Owns the address a Bitcoin withdrawal pays out to: the one of the connected
+ * wallet by default, or a custom one entered by the user.
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts]: Fixed in 3ae277f5.

portal/app/[locale]/tunnel/_components/retryBtcWithdraw.tsx
```diff
@@ -39,10 +39,13 @@ export const RetryBtcWithdraw = function ({ withdrawal }: Props) {
     // once the user confirms the withdraw, the withdrawal will change its state
     // and this component gets unmounted
     setOperationStatus('withdrawing')
-    const { amount, l1ChainId, l2ChainId } = withdrawal
+    const { amount, l1ChainId, l2ChainId, to } = withdrawal
 
     withdrawBitcoin({
       amount: BigInt(amount),
+      // retry the withdrawal to the address it originally targeted, which may
+      // not be the one of the connected wallet
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/retryBtcWithdraw.tsx]: Fixed in 3ae277f5.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,85 @@
...
+            />
+            {customAddress !== '' &&
+              (isCustomAddressValid ? (
+                <GreenCheckIcon />
+              ) : (
+                <div className="flex shrink-0 items-center gap-x-1 text-rose-600">
+                  <WarningIcon />
+                  <span>{t('invalid-address')}</span>
+                </div>
+              ))}
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Fixed in bf12b318. The input now has an `aria-label`, plus `aria-invalid` and `aria-describedby` pointing at the error.

portal/app/[locale]/tunnel/_components/reviewOperation/reviewBtcWithdrawal.tsx
```diff
@@ -74,8 +71,10 @@ const ReviewContent = function ({
     isError: isBitcoinWithdrawalEstimateFeesError,
   } = useEstimateBtcWithdrawFees({
     amount: parseTokenUnits(withdrawal.amount, fromToken),
-    btcAddress,
-    enabled: !!btcAddress && showWithdrawalStepFees,
+    // fees depend on the address the withdrawal targets, which is not
+    // necessarily the one of the currently connected wallet
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/reviewOperation/reviewBtcWithdrawal.tsx]: Fixed in 3ae277f5.

portal/app/[locale]/tunnel/_components/withdraw.tsx
```diff
@@ -206,15 +223,20 @@ const BtcWithdraw = function ({ state }: BtcWithdrawProps) {
     <TunnelForm
       belowForm={
         <div className="relative -translate-y-7">
-          <ReceivingAddress
+          <BitcoinReceivingAddress
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/withdraw.tsx:268]: Fixed in bf12b318. The toggle and the input are now disabled while withdrawing, like the amount and network controls.

portal/utils/bitcoin.ts
```diff
@@ -36,3 +39,58 @@ export const getBitcoinTimestamp = function (timestamp: number) {
...
+    const { data, prefix, version } = bitcoin.address.fromBech32(address)
+    if (prefix !== network.bech32) {
+      return false
+    }
+    // bech32 encodes witness versions up to 31, but Bitcoin only defines 0 to 16
+    if (version === 0) {
+      return data.length === 20 || data.length === 32
+    }
+    if (version === 1) {
+      return data.length === 32
```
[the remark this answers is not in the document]
arielmergen: Fixed in 05482695. Taproot programs are now checked with `schnorr.utils.lift_x` from `@noble/curves`, which is already in the bundle through viem.

Two notes on the report. The `tb1pqqqq…` fixture is not all-zero and its program is a valid curve point, so it was passing correctly. And `initEccLib` cannot be used here: it asserts the full `TinySecp256k1Interface`, so a partial shim is rejected.

The gap itself was real. I built a bech32m address with x = 0 and confirmed both the validator and `BitcoinKit.isAddressValid` accepted it, so nothing downstream was catching it.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,92 @@
...
+          <>
+            <input
+              aria-describedby={isCustomAddressValid ? undefined : errorId}
+              aria-invalid={!isCustomAddressValid}
+              aria-label={receivingText}
+              // mounting only happens when the user turns the toggle on
+              autoFocus
+              className="min-w-0 flex-1 bg-transparent text-neutral-950 placeholder:text-neutral-500 focus:outline-none disabled:cursor-not-allowed"
+              disabled={disabled}
+              onChange={e => onCustomAddressChange(e.target.value)}
```
Comment (ArturDolzan) [inline portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Perhaps it's worth checking what happens when an address is pasted with a trailing space or newline (fairly common from chat/email). It looks like it would show "Invalid address". Maybe a `.trim()` on change, plus `autoCapitalize="none"` for mobile typing?

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,92 @@
...
+          />
+        </div>
+      </div>
+      <div className="mt-4 h-px w-full bg-neutral-200" />
+      <div className="mt-4 flex h-[18px] items-center gap-x-2">
+        {useCustomAddress ? (
+          <>
+            <input
+              aria-describedby={isCustomAddressValid ? undefined : errorId}
+              aria-invalid={!isCustomAddressValid}
```
Comment (ArturDolzan) [inline portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Following up on the a11y thread fixed in `bf12b318`. Since `isValidBtcAddress('')` is false, the empty field is announced as invalid right after the toggle opens, and `aria-describedby` points at an error span that isn't rendered yet. Wonder if gating both on `customAddress !== ''` would read better?

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Fixed in 86121456. Confirmed whitespace in any position makes `fromBech32` reject the address, so `onChange` now trims. Added `autoCapitalize="none"` too: an all-uppercase address is valid, but the mixed case a mobile keyboard produces is not.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Fixed in 86121456. You're right, the empty field was announced as invalid with `aria-describedby` pointing at an unrendered span. Both are now gated on a single `showError` flag, which also drives the error markup.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,95 @@
...
+            ariaLabel={t('custom-address')}
+            checked={useCustomAddress}
+            disabled={disabled}
+            id={toggleId}
+            onCheckedChange={onUseCustomAddressChange}
+          />
+        </div>
+      </div>
+      <div className="mt-4 h-px w-full bg-neutral-200" />
+      <div className="mt-4 flex h-[18px] items-center gap-x-2">
```
Comment (ArturDolzan) [inline portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: I think you could use h-4.5 here.

```suggestion
      <div className="mt-4 flex h-4.5 items-center gap-x-2">
```

ArturDolzan (commented):
I left a minor improvement otherwise LGTM.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Good catch! Thank you! Added in 41c3528

ArturDolzan (dismissed):

gabmontes (dismissed):

portal/app/[locale]/tunnel/_components/reviewOperation/reviewBtcWithdrawal.tsx
```diff
@@ -74,8 +71,10 @@ const ReviewContent = function ({
     isError: isBitcoinWithdrawalEstimateFeesError,
   } = useEstimateBtcWithdrawFees({
     amount: parseTokenUnits(withdrawal.amount, fromToken),
-    btcAddress,
-    enabled: !!btcAddress && showWithdrawalStepFees,
+    // fees depend on the address the withdrawal targets, not necessarily the
+    // connected wallet's
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/reviewOperation/reviewBtcWithdrawal.tsx]: I think this comment adds no value

```suggestion
```

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,95 @@
...
+  isCustomAddressValid,
+  onCustomAddressChange,
+  onUseCustomAddressChange,
+  receivingText,
+  tooltipText,
+  useCustomAddress,
+}: Props) {
+  const t = useTranslations('tunnel-page.form')
+  // an empty field is not an error yet, it is just not filled in
+  const showError = customAddress !== '' && !isCustomAddressValid
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: We can confirm with Nahuel, but usually this is _true_ unless the field's been touched/focused at least once. That's the general UX pattern.

The flow would go like this:

- Field is empty. No error
- The user focuses on the field. Then it focuses on something else (in the middle, they either typed nothing or typed something and they deleted it).
- After focusing out, a validation like "field cannot be empty" or similar can be displayed. So far we've been showing these errors on the CTA, so I assume we should show it there.

portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts
```diff
@@ -0,0 +1,46 @@
...
+ * address by default, or a custom one entered by the user. The address check is
+ * merged into the amount validation, as the submit button reports it last.
+ */
+export const useReceivingBitcoinAddress = function ({
+  amountValidation,
+  network,
+  walletAddress,
+}: Props) {
+  const [customAddress, setCustomAddress] = useState('')
+  const [useCustomAddress, setUseCustomAddress] = useState(false)
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts]: `useCustomAddress` may be confusing because hooks are generally prefixed with `use`. I'd think a different name

gndelia:
> The validation checks the network because the contract doesn't: BitcoinKit.isAddressValid only verifies the address converts to a script, so a testnet tb1… is accepted on mainnet. Verified with eth_call.

I think BitcoinKit's only supports a subset of Bitcoin addresses. I think we should match what the contract accepts, and, on top of that, we should prevent `testnet` addresses from being mixed with `mainnet`. Perhaps we may need to discuss with Engineering which kind of addresses are accepted by the contract. For example, I am not sure if Taproot is supported

gndelia (commented):
For @dvnahuel  - I'll think if the user has typed a custom address, we'll have to show it in the Drawer when reviewing the operation

dvnahuel:
> For @dvnahuel - I'll think if the user has typed a custom address, we'll have to show it in the Drawer when reviewing the operation

Good catch, let's me think where to add that to the drawer

portal/utils/bitcoin.ts
```diff
@@ -36,3 +38,16 @@ export const getBitcoinTimestamp = function (timestamp: number) {
...
+  livenet: { base58: ['1', '3'], bech32: 'bc1' },
+  testnet: { base58: ['2', 'm', 'n'], bech32: 'tb1' },
+}
+
+export const isAddressOfBitcoinNetwork = function (
+  address: string,
+  network: BtcChain['id'],
+) {
+  const { base58, bech32 } = networkAddressPrefixes[network]
+  return address.toLowerCase().startsWith(bech32) || base58.includes(address[0])
```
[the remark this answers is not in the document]
arielmergen: Measured against mainnet BitcoinKit, with bech32m addresses generated for each witness version (20-byte program): v2, v15, v16, v17, v20 and v31 all revert, and a revert is reported as invalid, so versions above 1 do not pass. v0 with 21 bytes and v1 with 40 bytes revert as well.

The curve point is real, though: `bc1pqqqq…qqenm` (x = 0) returns true. So does v1 with a 20-byte program, which is not a taproot output either.

Both were given up on purpose in 17ff8049, when the validation moved to the contract to match exactly what it accepts. Recovering them means declaring `@scure/base` and `@noble/curves` as direct dependencies again.

portal/app/[locale]/tunnel/_hooks/useIsValidBtcAddress.ts
```diff
@@ -0,0 +1,29 @@
...
+  const hemi = useHemi()
+  const hemiClient = useHemiClient()
+
+  return useQuery({
+    enabled: enabled && address !== '',
+    queryFn: () =>
+      isValidBtcAddress(hemiClient, address).then(
+        isValid => isValid && isAddressOfBitcoinNetwork(address, network),
+      ),
+    queryKey: ['is-valid-btc-address', hemi.id, network, address],
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_hooks/useIsValidBtcAddress.ts:26]: Fixed in 001666b0. `useDebounce` at 300ms, and while the debounce lags behind the input the validity is reported as pending, so a cached answer for a previous address cannot enable the submit.

Typing a 62-character address: 10 `eth_call` to BitcoinKit before, 1 after.

portal/utils/hemi.ts
```diff
@@ -416,3 +417,18 @@ export const getBitcoinWithdrawalFee = ({
     .then(vaultStateAddress =>
       calculateBtcWithdrawalFee(hemiClient, { amount, vaultStateAddress }),
     )
+
+export const isValidBtcAddress = (
+  hemiClient: PublicClient,
+  btcAddress: string,
+) =>
+  getBitcoinKitAddress(hemiClient)
```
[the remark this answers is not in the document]
arielmergen: Fixed in 001666b0. Three cases in `test/utils/hemi.test.ts`, which already mocks `hemi-viem/actions`: the contract answer, a revert returning false, and any other failure being rethrown.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Fixed in 26f0c891 and 5145be05. `touched` now lives in the hook, so the CTA waits for the blur too - turning the toggle on no longer shows an error on an untouched field.

An empty field after the blur got its own message, `Enter a custom address`, mirroring `Enter an amount`. `Input a correct custom address` is left for an address that is actually invalid.

arielmergen:
> I think BitcoinKit's only supports a subset of Bitcoin addresses. (...) For example, I am not sure if Taproot is supported

Checked with `eth_call` on both chains. Taproot is supported: `bc1p5cyx…drcr` returns true on Hemi and on Hemi Sepolia. The contract takes witness versions 0 and 1 only - v2 to v31 revert, and so do wrong program lengths (v0 with 21 bytes, v1 with 40).

It does not check the network: `tb1…` passes on mainnet and `bc1…` passes on Sepolia. Base58 of the wrong network reverts instead of returning false.

So, as you asked, the validation now calls `BitcoinKit.isAddressValid` and checks the network on top of it, in 17ff8049. `bitcoinjs-lib` and `@noble/curves` are gone from the portal.

The custom address in the review Drawer is not part of this PR.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <div className="mt-4 h-px w-full bg-neutral-200" />
+      <div className="mt-4 flex h-4.5 items-center gap-x-2">
+        {customAddressEnabled ? (
+          <>
+            <input
+              aria-describedby={showError ? errorId : undefined}
+              aria-invalid={showError}
+              aria-label={receivingText}
+              // addresses are never capitalized, and mobile keyboards would
+              // otherwise produce a mixed case one, which is invalid
```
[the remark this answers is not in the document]
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Right, base58 addresses are mixed case - the testnet fixture `mipcBbFg9gMiCh81Kj8tqqdgoZub1ZJRfn` is one. Removed the comment in b745620e; `autoCapitalize="none"` stays.

pnpm-lock.yaml
```diff
@@ -26667,7 +26667,7 @@ snapshots:
   ox@0.6.9(typescript@5.8.3)(zod@3.25.76):
     dependencies:
       '@adraffy/ens-normalize': 1.11.1
-      '@noble/curves': 1.8.2
+      '@noble/curves': 1.9.7
```
ArturDolzan: Perhaps it's worth checking whether this lockfile line should still be here. No `package.json` changed in the PR anymore, so the `@noble/curves` bump under `ox@0.6.9` looks like leftover from the `bitcoinjs-lib`/`@noble/curves` iteration we reverted.

portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+    if (isError) {
+      return t('try-again')
+    }
+    return data === false ? t('input-a-correct-custom-address') : undefined
+  }
+
+  const canSubmit =
+    amountValidation.canSubmit &&
+    (!customAddressEnabled || (isChecked && data === true))
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts]: I don't think `amountValidation` belongs here. Validating a Bitcoin address depends only on the address being valid (checked against the contract, against the network), but is not related to the amount the user has entered. That's related to submission, and therefore should be done on the caller side (Withdraw.tsx I think)

If anything, this hook can return if the address is valid, and then the consumer can use it as part of its "canSubmit" calc.

pnpm-lock.yaml
gndelia: There are no changes in `package.json` so I don't think this file should be updated - let's revert it

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,100 @@
...
+      <div className="mt-4 flex h-4.5 items-center gap-x-2">
+        {customAddressEnabled ? (
+          <>
+            <input
+              aria-describedby={showError ? errorId : undefined}
+              aria-invalid={showError}
+              aria-label={receivingText}
+              autoCapitalize="none"
+              // mounting only happens when the user turns the toggle on
+              autoFocus
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx:71]: The input is autoFocused so onBlur never fires; customAddressTouched stays false, and the CTA sits disabled with no stated reason.

gndelia:
What are we showing while the address is pasted and validation is taking place? This is async, so in a slow network may take a few seconds - are we showing a spinner or something? I don't see that from the PR screenshots. Can we add a video or something showing the process of pasting the address and getting the green check?

pnpm-lock.yaml
arielmergen: Reverted in 46c75dde - `pnpm-lock.yaml` is back to `main`. Leftover from the `bitcoinjs-lib` iteration that was reverted; no `package.json` change is left in the PR.

arielmergen: Reverted in 46c75dde - `pnpm-lock.yaml` is back to `main`. Leftover from the `bitcoinjs-lib` iteration that was reverted; no `package.json` change is left in the PR.

portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts]: Fixed in f88a05ff. The hook no longer takes `amountValidation`: it returns `addressError` and `canSubmitAddress`, and `withdraw.tsx` combines them into `canWithdraw` and the submit button's `validationError`.

One note on `canWithdraw`: it is written as `[...].every(Boolean)` instead of a chain of `&&`. The extra `&&` plus the `??` pushed `BtcWithdraw` from a complexity of 10 to 12, over the eslint `complexity` max, and the repo has no such warning today.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx:71]: Fixed in 2df799ed by dropping `customAddressTouched` altogether. It gated two things, not one: the submit button's error and `showError` in the component, which drives the inline warning, `aria-invalid` and `aria-describedby`. The blur that never arrives hid both.

The debounce already covers not reporting while the user is still typing (`isChecked`), so the touched flag was only adding a gate that this form never reaches. Now, with the toggle on and the field empty, the button reads "Enter a custom address", and an address that fails the check shows the inline error without needing a blur.

portal/app/[locale]/tunnel/_hooks/useIsValidBtcAddress.ts
```diff
@@ -0,0 +1,29 @@
...
+}) {
+  const hemi = useHemi()
+  const hemiClient = useHemiClient()
+
+  return useQuery({
+    enabled: enabled && address !== '',
+    queryFn: () =>
+      isValidBtcAddress(hemiClient, address).then(
+        isValid => isValid && isAddressOfBitcoinNetwork(address, network),
+      ),
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_hooks/useIsValidBtcAddress.ts]: I'm wondering if here we should revert the order - make the sync and cheap call first, and then validate with the contract. Thoughts?

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <div className="mt-4 h-px w-full bg-neutral-200" />
+      <div className="mt-4 flex h-4.5 items-center gap-x-2">
+        {customAddressEnabled ? (
+          <>
+            <input
+              aria-describedby={showError ? errorId : undefined}
+              aria-invalid={showError}
+              aria-label={receivingText}
+              autoCapitalize="none"
+              // mounting only happens when the user turns the toggle on
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: I don't think this one adds value

```suggestion
```

gndelia:
> For @dvnahuel - I'll think if the user has typed a custom address, we'll have to show it in the Drawer when reviewing the operation

@arielmergen let's create an issue for this so we don't forget

gndelia:
IT seems there's an error in the validation because if my BTC wallet is disconnected, I can't submit and I get the "Connect your BTC wallet"


image

Additionally, if the address pasted is incorrect, the subsection does detect the "Invalid Address" (Ok), but the CTA still is hanging on "Connect your wallet"

image

The goal of pasting a custom address is to be able to initiate a withdrawal without having a BTC wallet connected - so let's fully test that user is able to fully use the new feature

gndelia:
I left some suggestions. Let's work on those - the UI implementation looks slick 🚀 🎉

arielmergen:
> What are we showing while the address is pasted and validation is taking place?

A spinner, added in 2a2e7ca7. It shows from the moment the field changes until the check settles - it covers both the 300ms debounce and the `BitcoinKit.isAddressValid` call - and it is replaced by the green check when the address is valid, or by the "Invalid Address" message when it is not.

portal/app/[locale]/tunnel/_components/retryBtcWithdraw.tsx
```diff
@@ -39,10 +39,11 @@ export const RetryBtcWithdraw = function ({ withdrawal }: Props) {
     // once the user confirms the withdraw, the withdrawal will change its state
     // and this component gets unmounted
     setOperationStatus('withdrawing')
-    const { amount, l1ChainId, l2ChainId } = withdrawal
+    const { amount, l1ChainId, l2ChainId, to } = withdrawal
 
     withdrawBitcoin({
       amount: BigInt(amount),
+      btcAddress: to,
```
Comment (gabmontes) [inline portal/app/[locale]/tunnel/_components/retryBtcWithdraw.tsx:46]: `to` isn't guaranteed here even though it's typed as a required `string` — a withdrawal synced from the subgraph can end up without it (see the `to`-recovery fallback in `sync-history/bitcoin.ts`), and a local record missing it is never corrected afterwards: `mergeContent` keeps the old entry on a duplicate `transactionHash`, so a stale broken record permanently wins over a later resync that gets `to` right.

The fee-estimate call just above already accounts for this with `enabled: !!withdrawal.to`, but this path doesn't — retrying such a withdrawal calls `withdrawBitcoin` with `btcAddress: undefined`.

Suggest bailing out (or disabling the retry button) when `!to`, the same way the fee estimate does.

portal/app/[locale]/tunnel/_hooks/useIsValidBtcAddress.ts
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_hooks/useIsValidBtcAddress.ts]: Done in `0b0bc23c`.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Removed in `a1dd9feb`.

portal/app/[locale]/tunnel/_components/retryBtcWithdraw.tsx
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/retryBtcWithdraw.tsx:46]: Fixed in 1b1ad903. The retry button is now disabled when the address is missing, the same as the fee estimate.

gabmontes (dismissed):
LGTM

ArturDolzan (dismissed):

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
```diff
@@ -0,0 +1,95 @@
+import { GreenCheckIcon } from 'components/icons/greenCheckIcon'
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: This file was removed in https://github.com/hemilabs/ui-monorepo/pull/2229 - so merging this PR will break the build. Let's rebase and fix

gndelia (changes requested):

portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (!isChecked) {
+      return undefined
+    }
+    if (isError) {
+      return t('try-again')
+    }
+    return data === false ? t('input-a-correct-custom-address') : undefined
+  }
+
+  const canSubmitAddress = !customAddressEnabled || (isChecked && data === true)
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts]: Should we consider `walletAddress` here? With the current implementation, if `customAddressEnabled` is false, the user won't use a custom address. But if false, then `canSubmitAddress` is true, as we're not checking that the wallet is connected here (basically, `walletAddress` should be defined).

If `walletAddress` is `undefined`, we can get into a state in the return a few lines below there `receivingAddress` is undefined but `canSubmitAddress` is true.

Maybe worth taking a look

portal/messages/pt.json
```diff
@@ -721,7 +724,9 @@
       "connect-both-wallets": "Conectar as suas carteiras BTC e EVM",
       "connect-btc-wallet": "Conectar a sua carteira BTC",
       "deposit": "Depositar",
+      "enter-a-custom-address": "Introduzir um endereço personalizado",
```
ArturDolzan: Just a minor suggestion here:
```suggestion
      "enter-a-custom-address": "Insira um endereço personalizado",
```

ArturDolzan (commented):
I left a minor suggestion of improvement otherwise LGTM.

portal/messages/pt.json
arielmergen: Applied in bedd538b — it also lines up with input-a-correct-custom-address right below, which already used "Insira".

portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_hooks/useReceivingBitcoinAddress.ts]: Fixed in 090339df — canSubmitAddress now requires walletAddress when the toggle is off, so it can't be true while receivingAddress is undefined.

That also turned the !!receivingAddress in feeEstimationEnabled into the same check twice, so it's gone in e0889add.

portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx
Comment (arielmergen) [reply portal/app/[locale]/tunnel/_components/bitcoinReceivingAddress.tsx]: Rebased on main, fixed in e864783b.

gndelia (approved):

gabmontes (approved):

ArturDolzan (approved):

PR #2205, opened 2026-08-19
Render images with the img element

ArturDolzan:
### Description

This PR replaces `next/image` with plain `` across the portal, and takes `pnpm typecheck` from 42 errors to 0.

Those 42 were all the same thing: `vite/client` types an image import as `string`, which is the truth, while the code was still typed against `StaticImageData`. They have been sitting there since the toolchain swap, drowning out anything new. This is phase 4 work pulled forward because it does not depend on routing or i18n, and it does not conflict with the env vars PR (16 files here, 30 there, no overlap).

Nothing is lost in the swap. The old `next.config.ts` had `images: { unoptimized: true }`, so `next/image` was not resizing, converting formats or proxying anything, and its `remotePatterns` entry for the token list host stops being needed. **What it did still provide were defaults**, and those had to come back by hand: it read `width`/`height` from the static import (which is how the browser reserves the box before the asset loads), and it lazy-loaded with async decoding. `components/image.tsx` restores the loading defaults in one place and requires `alt`, so a decorative image has to opt out with an empty string rather than by omission.

The dimensions could not be centralized, so they are declared next to each image. That is a maintenance trap, since a swapped asset would silently distort with no type or lint error, so `test/imageDimensions.test.ts` reads every file (PNG header, SVG viewBox) and asserts the declared numbers still match. It covers 27 assets and I checked that it actually fails by flipping one.

**This is not fully tested yet.** The four `fill` usages needed a CSS equivalent, and those are the only places where a visual regression is possible. I confirmed the Tailwind classes are emitted and produce what `fill` did (`size-full` is `width:100%;height:100%`, `object-cover` replaces the inline `objectFit`), but the components themselves cannot be rendered on this branch: the Vite entry is still a placeholder and `next dev` is gone. Three places need eyes once routing lands: `/ecosystem` (the demo card background and icon), the custom token drawer (the preview background SVG) and a review drawer mid operation (the `subStep` spinner, which should stay spinning and centered).

Tested with `pnpm lint`, `pnpm format:check`, `pnpm deps:check`, `pnpm typecheck`, `pnpm test` (878 tests) and `pnpm build`.

## Notes

`priority` in `next/image` meant eager plus high fetch priority plus a preload, and only three call sites had it: the two `demoCard` images and the `subStep` spinner. Those became `loading="eager"`. Nothing else did, so nothing else got it. The 404 and 500 artwork in particular stays lazy, which matters more than it looks: those SVGs are 1.7 MB and 1.4 MB.

The 11 DEX icons got `size-full` rather than explicit dimensions. They are all square (40x40 or 48x48) inside a square container, so the class both fits them and resolves the box before load, with no magic numbers.

`docs/add-new-token-to-stake.md` was documenting the old `StaticImageData` type and an example whose shape was already wrong. It now matches the registry and explains what `height`/`width` are for.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2194 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/not-found.tsx
+++ b/portal/app/[locale]/not-found.tsx
@@ -15,9 +15,7 @@ export default function NotFound() {
       <Image
         alt="404"
         className="absolute inset-0 -top-64 m-auto w-72 md:w-fit"
-        height={error404.height}
-        src={error404.src}
-        width={error404.width}
+        {...error404}
       />
       <div className="z-10 m-auto flex flex-col items-center gap-4">
         <ExclamationMark />
--- a/portal/app/[locale]/not-found.tsx
+++ b/portal/app/[locale]/not-found.tsx
@@ -5,7 +5,7 @@ import { ExclamationMark } from 'components/icons/exclamationMark'
 import { Image } from 'components/image'
 import { useTranslations } from 'next-intl'
 
-import Error404 from '../_images/404.svg'
+import { error404 } from '../_images/error404'
 
 export default function NotFound() {
   const t = useTranslations('error-pages')
@@ -15,11 +15,9 @@ export default function NotFound() {
       <Image
         alt="404"
         className="absolute inset-0 -top-64 m-auto w-72 md:w-fit"
-        fetchPriority="high"
-        height={402}
-        loading="eager"
-        src={Error404}
-        width={973}
+        height={error404.height}
+        src={error404.src}
+        width={error404.width}
       />
       <div className="z-10 m-auto flex flex-col items-center gap-4">
         <ExclamationMark />
--- a/portal/app/[locale]/not-found.tsx
+++ b/portal/app/[locale]/not-found.tsx
@@ -2,7 +2,7 @@
 
 import 'styles/globals.css'
 import { ExclamationMark } from 'components/icons/exclamationMark'
-import Image from 'next/image'
+import { Image } from 'components/image'
 import { useTranslations } from 'next-intl'
 
 import Error404 from '../_images/404.svg'
@@ -15,7 +15,11 @@ export default function NotFound() {
       <Image
         alt="404"
         className="absolute inset-0 -top-64 m-auto w-72 md:w-fit"
+        fetchPriority="high"
+        height={402}
+        loading="eager"
         src={Error404}
+        width={973}
       />
       <div className="z-10 m-auto flex flex-col items-center gap-4">
         <ExclamationMark />
--- a/portal/app/[locale]/stake/protocols/protocolImages.ts
+++ b/portal/app/[locale]/stake/protocols/protocolImages.ts
@@ -1,4 +1,3 @@
-import { StaticImageData } from 'next/image'
 import { StakeProtocols } from 'types/stake'
 
 import babypie from './images/babypie.png'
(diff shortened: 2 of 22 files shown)


portal/components/reviewOperation/subStep.tsx
```diff
@@ -30,9 +30,7 @@ export function SubStep({ description, status }: Props) {
           {isProgress && (
             <Image
               alt="Loading indicator"
-              className="animate-spin object-contain"
-              fill
-              priority
+              className="absolute inset-0 size-full animate-spin object-contain"
               src={gradientLoadingImg}
```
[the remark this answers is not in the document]
ArturDolzan: Good catch, this one slipped through. subStep was the only `priority` image of that pair (positionStatus never had it) and I dropped it without putting the eager translation back. Fixed in [`74b457a1`](https://github.com/hemilabs/ui-monorepo/pull/2205/commits/74b457a1).

Chasing it turned up a worse one right next door: I had *added* `loading="eager"` and `fetchPriority="high"` to the 404 and 500 artwork, which never had `priority` either. Those SVGs are 1.7 MB and 1.4 MB, so that was a real regression on the error pages. Reverted to lazy in the same commit, and `loading="eager"` now sits only where the original had `priority`.

portal/test/imageDimensions.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        expect(intrinsicSize(src)).toEqual({ height, width })
+      },
+    )
+  })
+
+  it.each([
+    { height: 402, name: '404.svg', src: error404, width: 973 },
+    { height: 349, name: '500.svg', src: svg500, width: 941 },
+    {
+      height: 20,
```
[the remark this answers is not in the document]
ArturDolzan: Fair, and it was a real gap: the test caught the asset changing but not the declared numbers changing. The three one-offs now live in modules next to their asset (`app/_images/error404.ts`, `components/error500/errorArtwork.ts`, `components/reviewOperation/_images/gradientLoading.ts`) in the same shape `protocolImages` already used, and the components render from those. Confirmed it fails now by flipping 402 to 500, which used to stay green. It also removed the duplicated 973x402 between the two not-found pages. Done in [`74b457a1`](https://github.com/hemilabs/ui-monorepo/pull/2205/commits/74b457a1).

Same commit hardened the parsing behind it, since reviewing this turned up that a bare `width=` also matches `stroke-width=` on a child (the hyphen opens a word boundary), so an SVG with only a viewBox reported 2x22 instead of 24x24. It reads the root tag only now, and handles comma separated viewBoxes and the base64 data URI Vite emits for SVGs containing `<text>`.

portal/components/image.tsx
```diff
@@ -0,0 +1,14 @@
+import { type ComponentProps } from 'react'
+
+// A bare <img> loads eagerly, so dropping next/image regressed every list of
+// logos into fetching all of them on mount. These restore the defaults it set.
+// `alt` is required so decorative images have to opt out with an empty string
+// rather than by omission.
```
gndelia: I'd drop this

```suggestion

```

This may be a temporary comment that makes sense, but in the future (once migration is complete) it will make no sense to reference `next/image`. Unless we're sure we'll revisit everything, I'd drop these kinds of comments

portal/app/[locale]/stake/protocols/protocolImages.ts
```diff
@@ -27,30 +26,30 @@ import yieldNest from './images/yieldNest.svg'
...
+  obeliskNodeDao: { height: 18, src: obeliskNodeDao, width: 64 },
+  pumpBtc: { className: 'w-24', height: 72, src: pumpBtc, width: 282 },
+  river: { height: 18, src: river, width: 52 },
+  solv: { height: 16, src: solv, width: 56 },
+  sumer: { className: 'w-15', height: 40, src: sumer, width: 131 },
+  tether: { height: 14, src: tether, width: 63 },
+  tetherGold: { height: 20, src: tetherGold, width: 64 },
+  threshold: { height: 10, src: threshold, width: 108 },
+  uniBtc: { height: 24, src: bedRock, width: 113 },
+  uniRouter: { className: 'w-21', height: 48, src: uniRouter, width: 166 },
```
Comment (gndelia) [inline portal/app/[locale]/stake/protocols/protocolImages.ts:52]: if `width` and `height` are set - is `w-21` relevant?

portal/app/[locale]/not-found.tsx
```diff
@@ -15,7 +15,9 @@ export default function NotFound() {
       <Image
         alt="404"
         className="absolute inset-0 -top-64 m-auto w-72 md:w-fit"
-        src={Error404}
+        height={error404.height}
+        src={error404.src}
+        width={error404.width}
```
Comment (gndelia) [inline portal/app/[locale]/not-found.tsx]: Minor, but using these objects lets you do


```suggestion
       {...error404}
```

portal/components/image.tsx
ArturDolzan: Fair, that one was written for the migration and would read as noise once it is over. Dropped the next/image part in [`20f9ce3e`](https://github.com/hemilabs/ui-monorepo/pull/2205/commits/20f9ce3e).

I kept the line about `alt` being required, since that one is intent rather than history (it is what stops someone from quietly making it optional again). Happy to drop that too if you'd rather have none there.

portal/app/[locale]/stake/protocols/protocolImages.ts
Comment (ArturDolzan) [reply portal/app/[locale]/stake/protocols/protocolImages.ts:52]: Yes, they do different jobs. The `width`/`height` attributes only give the browser the aspect ratio so it can reserve the box before the asset loads, and the class still wins for the rendered size (preflight's `height: auto` scales the height along with it).

These six are all quite a bit wider than their class, so dropping it would roughly double them: makerDao is 166px native rendered at `w-21` (84px), merlinChain 232px at `w-29`, pumpBtc 282px at `w-24`.

portal/app/[locale]/not-found.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/not-found.tsx]: Much nicer, done in [`20f9ce3e`](https://github.com/hemilabs/ui-monorepo/pull/2205/commits/20f9ce3e). I took it to the other four call sites too, and the best one was `protocolImage.tsx`, which was destructuring four fields just to hand them straight back and is now a one liner.

gndelia (approved):

gabmontes (approved):

PR #2207, opened 2026-08-20
Fix uneven card heights on boost stake page

### LINKED ISSUE #2151: The cards on "Boost stake" are not of the same size
### Description

See the image

Image

the 2 cards should have the same height

### Steps to reproduce

Just navigate to `/stake/dashboard/` and you'll see them

Godbrand0:
## Summary

Closes #2151.

On `/stake/dashboard`, the "Total Hemi points earned" card was taller than the "Your Stake" / "Total staked" cards.

- The shared `Container` in `stakePointsCards/index.tsx` wraps `Card` in a fixed `h-24` box, but `Card` itself has no height constraint, so it grows to fit its own content instead of the container's height.
- "Total Hemi points earned" has more content (a heading + description inside a bordered box) than the other cards (a heading + single value), so it rendered taller despite the `h-24` wrapper.
- Added `[&>div]:h-full` alongside the existing `[&>div]:overflow-hidden` so `Card` actually fills the `h-24` container, keeping all three cards the same height.

## Test plan

- [x] Verified locally on `/stake/dashboard` that all three cards now render at the same height

image

### DIFF
--- a/portal/app/[locale]/stake/dashboard/_components/stakePointsCards/index.tsx
+++ b/portal/app/[locale]/stake/dashboard/_components/stakePointsCards/index.tsx
@@ -33,8 +33,13 @@ const Points = ({
   'points'
 >) => <p className={`text-xl font-semibold ${color}`}>{points}</p>
 
+// No fixed height here on purpose: the parent row is a flex container, so
+// by default (align-items: stretch) every card is stretched to match the
+// tallest one. EarnedPoints, which is naturally shorter, fills that
+// stretched height itself (see its h-full chain) instead of the other
+// cards being clipped down to a fixed size.
 const Container = ({ children }: { children: ReactNode }) => (
-  <div className="h-24 w-full [&>div]:h-full [&>div]:overflow-hidden">
+  <div className="w-full [&>div]:h-full">
     <Card>
       <div className="relative h-full">{children}</div>
     </Card>
--- a/portal/app/[locale]/stake/dashboard/_components/stakePointsCards/index.tsx
+++ b/portal/app/[locale]/stake/dashboard/_components/stakePointsCards/index.tsx
@@ -34,7 +34,7 @@ const Points = ({
 >) => <p className={`text-xl font-semibold ${color}`}>{points}</p>
 
 const Container = ({ children }: { children: ReactNode }) => (
-  <div className="h-24 w-full [&>div]:overflow-hidden">
+  <div className="h-24 w-full [&>div]:h-full [&>div]:overflow-hidden">
     <Card>
       <div className="relative">{children}</div>
     </Card>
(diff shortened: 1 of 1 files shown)


gndelia:
I think this fix causes the 2 cards not to be vertically symmetric 

image

What I think we should do (imho) is that all the cards should have the section text with the same size. This is approx. 26px (given by the font size on the first 2 cards). As the last card uses a smaller `font-size`, this causes the card to have less height than the others. I'd increase the min-height of the "Season ended" plus vertically centering it, without changing the font-size. A similar option is for that last card, increasing the `gap-y` so it compensates for the smaller font size.

Godbrand0:
Fixed the orange card's alignment/spacing:
- The bordered box was sized by its content instead of a fixed height, so the inset from the card border shifted depending on the text. Fixed the height chain (`Container` → `Card` → `.relative` → the padding wrapper → the bordered box) so it now has a definite height throughout and stays vertically centered with a fixed inset regardless of content.
- Reduced the gap between the heading and "Season ended." from `gap-y-3` to `gap-y-2` so the two lines sit closer together.

Pushed in 38356843.

gndelia (changes requested):
I think the fix is still wrong. This causes for the small padding below not to match the padding at the top of the box. See

image

Compare with the container before the fix 

image

Note that the fix needs to be done on the card that say "Season ended", matching the height of the others. We don't need to shrink the 2 cards that actually work - we need to enlarge the one that's shorter

Godbrand0:
You're right, apologies — reworked the fix per your guidance: enlarge the shorter card instead of clipping the ones that work.

- Removed the fixed `h-24` + `overflow-hidden` forcing from `Container`. The parent row (`page.tsx`) is a flex container, so with the default `align-items: stretch` every card is naturally stretched to match the tallest one — no forced pixel height needed, and `TotalStaked`/`YourStake` are completely untouched.
- `EarnedPoints` now fills that stretched height itself (`h-full` chain from `Card` down to the bordered box, with `justify-center` to vertically center its content), so it grows to match the other two instead of them shrinking to it.

Verified against a production build (the dev server had gone stale from switching branches) — all three cards render at equal height with symmetric padding on the orange card. Pushed in 47638aeb.

Godbrand0:
image
Screenshot from 2026-08-21 11-36-58
Screenshot from 2026-08-21 11-36-43


the heights are the same

portal/app/[locale]/stake/dashboard/_components/stakePointsCards/index.tsx
```diff
@@ -33,10 +33,15 @@ const Points = ({
   'points'
 >) => <p className={`text-xl font-semibold ${color}`}>{points}</p>
 
+// No fixed height here on purpose: the parent row is a flex container, so
+// by default (align-items: stretch) every card is stretched to match the
+// tallest one. EarnedPoints, which is naturally shorter, fills that
+// stretched height itself (see its h-full chain) instead of the other
+// cards being clipped down to a fixed size.
```
Comment (gndelia) [inline portal/app/[locale]/stake/dashboard/_components/stakePointsCards/index.tsx]: I think I would drop this IA generated comment


```suggestion
```

It's just explaining implementation 🤷🏽

gndelia (approved):

arielmergen (approved):

gndelia:
Thanks @Godbrand0  for the fix! Merged

PR #2210, opened 2026-08-20
Load components with lazy instead of next/dynamic

ArturDolzan:
### Description

This PR replaces `next/dynamic` with `React.lazy` and `Suspense` across the portal, 40 call sites in 35 files.

It is the first of two for the routing phase. It comes first because it does not depend on the router at all: `lazy` and `Suspense` are plain React and behave the same under Next and under Vite, so it can be reviewed and merged on its own instead of riding along with the route tree. It also clears 35 `next/*` imports ahead of the big one, which shrinks the conflict surface with `main` while the migration branch is open. `components/tooltip/index.tsx` already used this exact shape, so the pattern is the repo's own.

**No behavior change is intended**, and that is the requirement rather than a happy side effect. `next/dynamic` did two things: split the chunk, and through its `loading` option decide what shows meanwhile. `lazy` keeps the first, and the second moves to a `Suspense` boundary at the point of use. Which means the whole risk of this PR is where that boundary lands.

The 17 calls that had a `loading` got their fallback transplanted literally. **The 23 without one mattered just as much**: they used to render nothing on their own, and a `lazy` component with no boundary of its own suspends up to the nearest ancestor, which here is the last-resort wrapper in the root layout. Left implicit, something that used to disappear quietly would blank the page instead.

Two of them are not rendered where they are declared, they are passed by reference (`balanceComponent={StakedBalance}`, `balanceComponent ?? Balance`). Wrapping them at the render site would apply their skeleton to whatever a caller passes instead, so the boundary travels with the component. Those are the only two comments the PR adds.

`ssr: false` is the only option dropped outright, and it costs nothing: it told Next not to render on the server, and there is no server.

## Notes

**This is not visually verified.** The Vite entry is still a placeholder and `next dev` is gone, so nothing here can be rendered on this branch. Lint, types and tests do not catch a misplaced boundary either, which makes file by file review the real check on this one. It becomes observable when the routing PR lands.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2194 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/get-started/_components/addChainAutomatically.tsx
+++ b/portal/app/[locale]/get-started/_components/addChainAutomatically.tsx
@@ -1,4 +1,3 @@
-import dynamic from 'next/dynamic'
 import { lazy, Suspense } from 'react'
 import { type Chain } from 'viem'
 
@@ -14,12 +13,10 @@ const AddChain = lazy(() =>
   import('./addChain').then(mod => ({ default: mod.AddChain })),
 )
 
-const AddChainButton = dynamic(
-  () => import('./addChain/addChainButton').then(mod => mod.AddChainButton),
-  {
-    loading: () => <span aria-hidden="true" className="block h-7" />,
-    ssr: false,
-  },
+const AddChainButton = lazy(() =>
+  import('./addChain/addChainButton').then(mod => ({
+    default: mod.AddChainButton,
+  })),
 )
 
 export const AddChainAutomatically = function ({ chain, layer }: Props) {
@@ -27,7 +24,11 @@ export const AddChainAutomatically = function ({ chain, layer }: Props) {
     <ChainIdentityRow
       chain={chain}
       layer={layer}
-      trailing={<AddChainButton chain={chain} />}
+      trailing={
+        <Suspense fallback={<span aria-hidden="true" className="block h-7" />}>
+          <AddChainButton chain={chain} />
+        </Suspense>
+      }
     />
   )
 
--- a/portal/app/[locale]/get-started/_components/addChainAutomatically.tsx
+++ b/portal/app/[locale]/get-started/_components/addChainAutomatically.tsx
@@ -1,6 +1,7 @@
-import { lazy, Suspense } from 'react'
+import { lazy, type ReactNode, Suspense } from 'react'
 import { type Chain } from 'viem'
 
+import { AddChainButton } from './addChain/addChainButton'
 import { ChainIdentityRow } from './addChain/chainIdentityRow'
 import { Container } from './addChain/container'
 
@@ -13,28 +14,20 @@ const AddChain = lazy(() =>
   import('./addChain').then(mod => ({ default: mod.AddChain })),
 )
 
-const AddChainButton = lazy(() =>
-  import('./addChain/addChainButton').then(mod => ({
-    default: mod.AddChainButton,
-  })),
-)
-
 export const AddChainAutomatically = function ({ chain, layer }: Props) {
-  const content = (
-    <ChainIdentityRow
-      chain={chain}
-      layer={layer}
-      trailing={
-        <Suspense fallback={<span aria-hidden="true" className="block h-7" />}>
-          <AddChainButton chain={chain} />
-        </Suspense>
-      }
-    />
+  const row = (trailing: ReactNode) => (
+    <ChainIdentityRow chain={chain} layer={layer} trailing={trailing} />
   )
 
   return (
-    <Suspense fallback={<Container>{content}</Container>}>
-      <AddChain chain={chain}>{content}</AddChain>
+    <Suspense
+      fallback={
+        <Container>
+          {row(<span aria-hidden="true" className="block h-7" />)}
+        </Container>
+      }
+    >
+      <AddChain chain={chain}>{row(<AddChainButton chain={chain} />)}</AddChain>
     </Suspense>
   )
 }
--- a/portal/app/[locale]/staking-dashboard/_components/stakeForm.tsx
+++ b/portal/app/[locale]/staking-dashboard/_components/stakeForm.tsx
@@ -1,9 +1,10 @@
 'use client'
 
+import { lazyWithFallback } from 'components/lazyWithFallback'
 import { ToastLoader } from 'components/toast/toastLoader'
 import { useHemiToken } from 'hooks/useHemiToken'
 import { useTranslations } from 'next-intl'
-import { lazy, Suspense } from 'react'
+import { Suspense } from 'react'
 import Skeleton from 'react-loading-skeleton'
 import {
   CollectAllRewardsDashboardStatus,
@@ -17,8 +18,9 @@ import { useDrawerStakingQueryString } from '../_hooks/useDrawerStakingQueryStri
 import { Stake } from './stake'
 import { StakeReview } from './stakeReview'
 
-const StakeToast = lazy(() =>
-  import('./stakeToast').then(mod => ({ default: mod.StakeToast })),
+const StakeToast = lazyWithFallback(
+  () => import('./stakeToast').then(mod => ({ default: mod.StakeToast })),
+  <ToastLoader />,
 )
 
 const SideDrawer = function () {
@@ -77,28 +79,22 @@ export const StakeForm = function () {
   return (
     <>
       {showStakeToast && (
-        <Suspense fallback={<ToastLoader />}>
-          <StakeToast
-            title={t('staking-dashboard.stake-successful')}
-            transactionHash={stakingDashboardOperation.transactionHash!}
-          />
-        </Suspense>
+        <StakeToast
+          title={t('staking-dashboard.stake-successful')}
+          transactionHash={stakingDashboardOperation.transactionHash!}
+        />
       )}
       {showUnlockToast && (
-        <Suspense fallback={<ToastLoader />}>
-          <StakeToast
-            title={t('staking-dashboard.unlock-successful')}
-            transactionHash={unlockingDashboardOperation.transactionHash!}
-          />
-        </Suspense>
+        <StakeToast
+          title={t('staking-dashboard.unlock-successful')}
+          transactionHash={unlockingDashboardOperation.transactionHash!}
+        />
       )}
       {showCollectRewardsToast && (
-        <Suspense fallback={<ToastLoader />}>
-          <StakeToast
-            title={t('staking-dashboard.claim-rewards-successful')}
-            transactionHash={collectRewardsDashboardOperation.transactionHash!}
-          />
-        </Suspense>
+        <StakeToast
+          title={t('staking-dashboard.claim-rewards-successful')}
+          transactionHash={collectRewardsDashboardOperation.transactionHash!}
+        />
       )}
       <Stake />
       <Suspense>
--- a/portal/app/[locale]/staking-dashboard/_components/stakeForm.tsx
+++ b/portal/app/[locale]/staking-dashboard/_components/stakeForm.tsx
@@ -2,9 +2,8 @@
 
 import { ToastLoader } from 'components/toast/toastLoader'
 import { useHemiToken } from 'hooks/useHemiToken'
-import dynamic from 'next/dynamic'
 import { useTranslations } from 'next-intl'
-import { Suspense } from 'react'
+import { lazy, Suspense } from 'react'
 import Skeleton from 'react-loading-skeleton'
 import {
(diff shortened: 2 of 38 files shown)


portal/app/[locale]/staking-dashboard/_components/stakeForm.tsx
```diff
@@ -82,22 +77,28 @@ export const StakeForm = function () {
...
-        <StakeToast
-          title={t('staking-dashboard.claim-rewards-successful')}
-          transactionHash={collectRewardsDashboardOperation.transactionHash!}
-        />
+        <Suspense fallback={<ToastLoader />}>
+          <StakeToast
+            title={t('staking-dashboard.claim-rewards-successful')}
+            transactionHash={collectRewardsDashboardOperation.transactionHash!}
+          />
+        </Suspense>
```
Comment (gndelia) [inline portal/app/[locale]/staking-dashboard/_components/stakeForm.tsx]: With a bit of reorg, you can here use just one `Suspense` as a parent, and then inside render the component accordingly. After all, it's the same component lazy loaded with different props - no need for 3 Suspense components!

portal/app/demos/page.tsx
```diff
@@ -1,13 +1,17 @@
...
+    default: mod.LocalePageRedirect,
+  })),
 )
 
-const Page = () => <LocalePageRedirect redirectPage="/demos" />
+const Page = () => (
+  <Suspense>
+    <LocalePageRedirect redirectPage="/demos" />
+  </Suspense>
+)
```
gndelia: This component was lazy-loaded because it forced a redirect on the client side (and we could not have that on SSR). But not that everything is on client side, I think we can just import it - no need to lazy load

gndelia: (This applies to all usages of `LocalePageRedirect`)

portal/components/tokenInput/index.tsx
```diff
@@ -1,21 +1,22 @@
...
 
-const Balance = dynamic(
-  () => import('components/cryptoBalance').then(mod => mod.Balance),
-  {
-    loading: () => (
-      <Skeleton className="h-full" containerClassName="basis-1/3" />
-    ),
-    ssr: false,
-  },
+const LazyBalance = lazy(() =>
```
gndelia: Why not.. 

```suggestion
const CryptoBalance = lazy(() =>
```

?

portal/app/[locale]/get-started/_components/addChainAutomatically.tsx
```diff
@@ -14,20 +13,22 @@ const AddChain = lazy(() =>
...
 )
 
 export const AddChainAutomatically = function ({ chain, layer }: Props) {
   const content = (
     <ChainIdentityRow
       chain={chain}
       layer={layer}
-      trailing={<AddChainButton chain={chain} />}
+      trailing={
+        <Suspense fallback={<span aria-hidden="true" className="block h-7" />}>
```
Comment (gndelia) [inline portal/app/[locale]/get-started/_components/addChainAutomatically.tsx]: `content` is rendered in two places: in the fallback of the outer boundary (`<Container>{content}</Container>`) and in the primary tree (`<AddChain>{content}</AddChain>`). These are different tree positions, so React does not reconcile one into the other.

While the `./addChain` chunk is in flight, the fallback tree mounts `AddChainButton`, which shows this `h-7` placeholder and then the real button. When the chunk resolves, React unmounts the fallback tree and mounts the primary tree, so `AddChainButton` mounts a second time and the placeholder flashes again.

The double mount is pre-existing (`dynamic` with `loading` also made a boundary here), but the PR rewrites this exact block. Hoisting `content` out of the fallback, or letting `AddChain` own the fallback shape, removes it.

portal/app/[locale]/tunnel/_components/withdraw.tsx
```diff
@@ -45,14 +44,14 @@ import { SubmitEvmWithdrawal } from './submitEvmWithdrawal'
 import { SubmitWithTwoWallets } from './submitWithTwoWallets'
 import { TunnelProviderToggle } from './tunnelProviderToggle'
 
-const SetMaxEvmBalance = dynamic(
-  () => import('components/setMaxBalance').then(mod => mod.SetMaxEvmBalance),
-  { ssr: false },
+const SetMaxEvmBalance = lazy(() =>
+  import('components/setMaxBalance').then(mod => ({
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/_components/withdraw.tsx]: `components/setMaxBalance` is also imported statically in three files:

- `app/[locale]/stake/_components/manageStake/maxBalance.tsx`
- `app/[locale]/staking-dashboard/_components/stakeReview/increaseAmount/preview.tsx`
- `app/[locale]/hemi-earn/pool/[shareAddress]/_components/withdrawMaxBalance.tsx`

Rollup gives its "dynamically imported by ... but also statically imported by ..." warning and keeps the module in the main chunk, so this split saves no bytes. But `lazy` still resolves through a microtask, so the boundary commits its fallback for one frame on each fresh mount.

The same applies to the other four `SetMaxEvmBalance` / `SetMaxBtcBalance` declarations, in `btcDeposit.tsx`, `evmDeposit.tsx`, `staking-dashboard/_components/stake.tsx` and `hemi-earn/pool/[shareAddress]/_components/deposit.tsx`. Because the PR rewrites these declarations, this is a good moment to make them plain static imports.

portal/components/tokenInput/index.tsx
gndelia: Same as in `withdraw.tsx`: `components/cryptoBalance` is also imported statically in four files:

- `app/[locale]/stake/_components/stakeStrategyTable/index.tsx` (`Balance`)
- `app/[locale]/stake/_components/stakedBalance.tsx`
- `app/[locale]/hemi-earn/pool/[shareAddress]/_components/withdrawAvailableBalance.tsx`
- `app/[locale]/hemi-earn/pool/[shareAddress]/_components/withdrawShareBalance.tsx`

The module stays in the main chunk, so the only effect of the boundary is that the `basis-1/3` skeleton flashes for one frame on each fresh mount. A static import removes the flash and also removes the need for the `Balance` wrapper below.

gndelia:
A relevant AI suggestion:  👇🏽  (Up to you if you feel this works better)

The conversion repeats the same shape about 40 times:

```tsx
const X = lazy(() => import('...').then(mod => ({ default: mod.X })))
// ...
<Suspense fallback={/* ... */}>
  <X />
</Suspense>
```

It also writes the wrapper form by hand twice, in `components/tokenInput/index.tsx` (`Balance`) and in `app/[locale]/stake/_components/manageStake/unstakeOperation.tsx` (`StakedBalance`), because those two components are passed by reference and not rendered where they are declared.

The PR description says the risk is boundary placement, and that no lint rule, type or test catches a boundary that is missing or in the wrong place. A small helper removes that risk, because the boundary then comes with the component:

```tsx
export const lazyWithFallback = function <P extends object>(
  loader: () => Promise<{ default: ComponentType<P> }>,
  fallback?: ReactNode,
) {
  const Lazy = lazy(loader)
  return function (props: P) {
    return (
      <Suspense fallback={fallback}>
        <Lazy {...props} />
      </Suspense>
    )
  }
}
```

This is what `next/dynamic` did, so each call site keeps the same ergonomics as before:

```tsx
const AddChainButton = lazyWithFallback(
  () => import('./addChain/addChainButton').then(mod => ({ default: mod.AddChainButton })),
  <span aria-hidden="true" className="block h-7" />,
)
```

Benefits:

- The two hand-written wrappers stop being special cases.
- The four copies of the `ConnectEvmWallet` declaration plus `<Suspense fallback={<ButtonLoader />}>` (`components/submitWhenConnected.tsx`, `components/submitWhenConnectedToChain.tsx`, `app/[locale]/tunnel/_components/submitWithTwoWallets.tsx`, `app/[locale]/genesis-drop/_components/disconnectedState.tsx`) collapse into one shared module.
- A future `lazy` in the portal cannot forget its boundary.
- It removes a large part of the 424 added lines.

portal/app/[locale]/staking-dashboard/_components/stakeForm.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/staking-dashboard/_components/stakeForm.tsx]: Good call. I went with the helper you suggested below instead of one parent boundary, so it is declared once now and each toast carries its own. Same effect here since only one shows at a time. [`096cf1c2`](https://github.com/hemilabs/ui-monorepo/pull/2210/commits/096cf1c2)

portal/app/demos/page.tsx
ArturDolzan: Yep, all of them are plain imports now in [`096cf1c2`](https://github.com/hemilabs/ui-monorepo/pull/2210/commits/096cf1c2).

portal/components/tokenInput/index.tsx
ArturDolzan: Ended up going further based on your other comment, `cryptoBalance` is a static import now so there is no lazy left here at all, and the `Balance` wrapper is gone. [`096cf1c2`](https://github.com/hemilabs/ui-monorepo/pull/2210/commits/096cf1c2)

portal/app/[locale]/get-started/_components/addChainAutomatically.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/get-started/_components/addChainAutomatically.tsx]: Good catch, and it was actually worse than a flash. The real button was rendering inside the fallback `Container`, which carries no onClick, so it looked enabled but did nothing until the chunk landed. Fixed in [`096cf1c2`](https://github.com/hemilabs/ui-monorepo/pull/2210/commits/096cf1c2), `AddChainButton` is static now and only the placeholder lives in the fallback.

portal/app/[locale]/tunnel/_components/withdraw.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/tunnel/_components/withdraw.tsx]: Right, all five are plain imports now. [`096cf1c2`](https://github.com/hemilabs/ui-monorepo/pull/2210/commits/096cf1c2)

portal/components/tokenInput/index.tsx
ArturDolzan: Same here, static import and the `Balance` wrapper went away with it. [`096cf1c2`](https://github.com/hemilabs/ui-monorepo/pull/2210/commits/096cf1c2)

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2210#issuecomment-5361848226):
> A relevant AI suggestion: (Up to you if you feel this works better)
> [...] A small helper removes that risk, because the boundary then comes with the component

Took it, thanks. It landed in [`096cf1c2`](https://github.com/hemilabs/ui-monorepo/pull/2210/commits/096cf1c2) as `components/lazyWithFallback.tsx`, backing 23 declarations across 20 files, and the four `ConnectEvmWallet` copies now come from one shared module.

One deviation from your snippet: the constraint is `ComponentType<any>` rather than `P extends object`. I tried the narrower one first and it does compile, but it silently drops the weak type check, so things like `<EarnCard>text</EarnCard>` and stray spreads stop erroring. `ComponentType<any>` is what React uses for `lazy` itself. I left a comment on it so it does not get tidied away later.

Going through the call sites also turned up 14 splits that saved nothing (the module was already imported statically somewhere else, so Rollup kept it in the main chunk), and those are plain imports now. That is where most of the line reduction came from.

gabmontes (approved):

portal/app/[locale]/tunnel/transaction-history/page.tsx
```diff
@@ -1,30 +1,24 @@
...
 import { PageLayout } from 'components/pageLayout'
 import { PageTitle } from 'components/pageTitle'
-import dynamic from 'next/dynamic'
 import { useTranslations } from 'next-intl'
 import { useState } from 'react'
 import Skeleton from 'react-loading-skeleton'
 
 import { type FilterOptions, TopBar } from './_components/topBar'
 
-// using CSR because useWindowSize doesn't work on SSR
```
Comment (gndelia) [inline portal/app/[locale]/tunnel/transaction-history/page.tsx:12]: This component was lazyly loaded because of `useWindowSize` not working on SSR, but now with Vite that may no longer be the case. Perhaps it can be directly imported

Comment (ArturDolzan) [reply portal/app/[locale]/tunnel/transaction-history/page.tsx:12]: Yeah, that reason is gone with Vite, the original comment even said it was about `useWindowSize` on SSR. Plain import now in [`2ce8bcf3`](https://github.com/hemilabs/ui-monorepo/pull/2210/commits/2ce8bcf3).

Worth noting it does still save bytes today (that module has no other importer), unlike the ones we collapsed earlier. But once the route itself is lazy in the routing PR we get the split back for free, so the inner boundary would just be a chunk inside a chunk.

gndelia (approved):

PR #2211, opened 2026-08-20
Declare the Inter faces in plain CSS

ArturDolzan:
### Description

This PR replaces `next/font/local` with plain `@font-face` rules, and moves the four Inter files into `public/`.

It is a prerequisite for the routing step rather than a cleanup. `next/font/local` is a Next compiler transform that does not exist under Vite, and the variables it produced were applied as a className on `<body>`, which becomes `index.html` once routing lands. Today it only survives because `fonts/index.ts` is not in the placeholder entry's graph.

**The part that was almost lost is the metric-adjusted fallback.** Next did not just declare the faces, it measured Inter and emitted a second `@font-face` whose `ascent-override`, `descent-override` and `size-adjust` make the local font occupy exactly the same space, then put both families in the CSS variable. That is what stops text from shifting when the woff2 arrives. Copying only the obvious rules would have reintroduced that shift everywhere, with nothing failing. The percentages here are lifted from the previous build output, so they are the values Next
computed rather than a guess.

The files live in `public/` because Vite content-hashes imported assets, and a hashed name cannot be written into a static `index.html` preload. Next injected four `<link rel="preload" as="font">` on every page and dropping them silently was a real regression. The trade is that we lose content hashing, so replacing a woff2 later needs a rename to bust caches, which matters precisely because the override percentages are tied to the file.

Two things ended up better than they were. The stacks now end in `sans-serif`: the only fallback resolved through `local('Arial')`, which does not exist on most Linux desktops, so the page fell through to the browser default, a serif. And `app/global-error.tsx` loses an `inter-font` class that matches nothing, two preconnects to Google Fonts that this app never requests, and a `<link rel="stylesheet" />` with no href.

## Notes

**Stacked on the `next/dynamic` PR**, so its commit shows in this diff until that one merges.

Unlike the two before it, this one is actually verifiable: `dist/` carries the four preloads and the four woff2, the emitted CSS has all six `@font-face` including both `ascent-override` blocks, and the Typography story renders it for real.

Preloading all four (689 KB) is the default inherited from `next/font`, not a considered choice. All three display cuts are reachable (h5 and h6 fall to 400, and nine headings use `font-medium`), but whether they deserve high-priority preload ahead of the render-blocking JS and CSS is a question for when the app actually renders and there is a waterfall to look at.

Storybook's `staticDirs` is scoped to `public/fonts` rather than `public/`, otherwise it publishes `navbarVideo.mp4` into every build and hijacks Storybook's own favicon.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2194 

### Checklist

- [x] Manual testing passed.


gabmontes (approved):

portal/app/global-error.tsx
```diff
@@ -15,11 +15,8 @@ const GlobalError = (props: Pick<ComponentProps<typeof Error500>, 'error'>) => (
     <head>
       <title>Hemi Portal</title>
       <link href="/favicon.ico" rel="icon" />
-      <link href="https://fonts.googleapis.com" rel="preconnect" />
-      <link href="https://fonts.gstatic.com" rel="preconnect" />
-      <link rel="stylesheet" />
```
gndelia, line 20: These were added because Next uses the global-error page in a specific way and loads it if there's an unhandled exception. Because of that, fonts needed to be loaded separately, as this component was rendered in isolation, outside of Next's regular flow (so nothing from Next fonts would be loaded).

Just making an FYI, as I don't know how Vite handles these errors

portal/README.md
```diff
@@ -12,6 +12,7 @@ Some relevant folders are:
 - [/components](./components/) folder, which contains reusable components to the entire app that are not tied to a specific page.
 - [/hooks](./hooks/) folder, which contains reusable hooks to the entire app that are not tied to a specific page.
 - [/messages/](./messages/) folder, which contain a file per locale with all the translated resources.
+- [/public](./public/) folder, which is served verbatim at the site root, so `public/fonts/inter/...` is reachable at `/fonts/inter/...`. Both `index.html` and `styles/globals.css` reference it by absolute URL.
```
gndelia:
```suggestion
- [/public](./public/) folder, which is served verbatim at the site root.
```

I'd simplify it

portal/wrangler.jsonc
```diff
@@ -10,6 +10,10 @@
   // web/src/index.ts in vetro-monorepo sets `'none'` (or `blob:` with Sentry)
   // because it has no web workers; copying that verbatim silently kills the
   // five under portal/workers, and tunnel history simply stops syncing.
+  //
+  // It also needs `font-src 'self'`: the Inter faces are fetched from our own
+  // origin by styles/globals.css, and a CSP that enumerates directives instead
+  // of leaning on `default-src` blocks them into the fallback font, silently.
```
gndelia: not really needed? This comment doesn't seem relevant to what's written below

portal/app/global-error.tsx
ArturDolzan: Thanks for the context, that explains why they were there. They were not doing anything though: nothing in the portal fetches from googleapis or gstatic, and the `inter-font` class does not exist anywhere in the repo, so this page was rendering in a system font.

It is actually better now. The variables moved to `:root` in `styles/globals.css`, which this file already imports on line 3, so the error page picks up Inter for the first time.

On the Vite side, it never mounts this file at all (it is a Next convention), so the routing PR will replace it with a real error boundary.

portal/README.md
ArturDolzan: Fair enough, simplified in [`0062671a`](https://github.com/hemilabs/ui-monorepo/pull/2211/commits/0062671a).

portal/wrangler.jsonc
ArturDolzan: You are right, there is no CSP in that file at all (assets only, no Worker), so the note was describing something that does not exist yet. Dropped in [`0062671a`](https://github.com/hemilabs/ui-monorepo/pull/2211/commits/0062671a), which puts the file back to matching the base. I moved it to #2194 next to the worker-src note so it resurfaces when we write the headers Worker.

portal/styles/globals.css
```diff
@@ -4,6 +4,68 @@
 @import '../components/drawer/drawer.css';
 @import 'tailwindcss/utilities';
 
+@font-face {
+  font-display: swap;
+  font-family: interDisplay;
+  font-style: normal;
+  font-weight: 400;
+  src: url('/fonts/inter/display/InterDisplay-Regular.woff2') format('woff2');
```
gndelia, line 12: `/fonts/...` is a stable path, so the four woff2 lose the content hashing that `next/font` gave them (`/_next/static/media/<hash>.p.woff2`, served `max-age=31536000, immutable`).

Cloudflare Workers Static Assets serves non-hashed assets with `public, max-age=0, must-revalidate` plus an ETag. Each page load then makes four conditional GETs for 689 KB of preloaded fonts before anything is reused from cache. The JS and CSS bundles stay content-hashed, so the fonts become the only unversioned asset.

The extra requests are the small part. The larger risk is that a client which holds a stale cached woff2 pairs it with new override percentages. That is the one mismatch the metric-adjusted fallbacks cannot absorb.

portal/styles/globals.css
```diff
@@ -4,6 +4,68 @@
...
+  descent-override: 22.36%;
+  font-family: 'interVariable Fallback';
+  line-gap-override: 0%;
+  size-adjust: 107.89%;
+  src: local('Arial'), local('Helvetica'), local('Liberation Sans');
+}
+
+:root {
+  --font-inter-display: 'interDisplay', 'interDisplay Fallback', sans-serif;
+  --font-inter-variable: 'interVariable', 'interVariable Fallback', sans-serif;
```
gndelia, line 66: Related to how these two variables are applied further down the file: `* { @apply font-inter-variable }` (line 89) matches inline children of headings directly, while `h1..h6 { @apply font-inter-display }` (line 93) reaches those children only by inheritance. A declared value always beats an inherited one, so any element nested in a heading falls back to interVariable.

`components/connectWallets/walletQRCodeView.tsx:109` renders two `<span>` elements inside an `<h4>`, so that heading text shows in the wrong face. The typography story uses `<h1>{SAMPLE}</h1>` with no nesting, so it does not surface there.

Both rules are pre-existing, but this PR is the point where the two families are wired by hand. Scoping the reset to `body` (or `:where(*)`) makes the heading rule win again.

portal/index.html
```diff
@@ -4,6 +4,34 @@
...
+      type="font/woff2"
+      crossorigin
+    />
+    <link
+      rel="preload"
+      href="/fonts/inter/display/InterDisplay-SemiBold.woff2"
+      as="font"
+      type="font/woff2"
+      crossorigin
+    />
```
gndelia, line 34: These four preloads total 689 KB: InterVariable is 344 KB and the three display cuts are 106 + 111 + 111 KB. `as="font"` is fetched at Highest priority, ahead of the render-blocking CSS and JS that Vite injects later in the same `<head>`, so first paint waits behind all of it on a cold load.

The `size-adjust` / `ascent-override` fallbacks added in this same PR exist so a late woff2 does not shift layout, which is exactly the cost preloading is meant to avoid. InterVariable also carries an `opsz` axis (14-32) that interpolates toward the Display design, so the 328 KB of static display cuts are partly redundant with it.

Worth cutting to InterVariable alone, or dropping the preloads.

gndelia:
LGTM - I left some IA suggestions that I thought could be worth reviewing, but take them with a grain of salt

gndelia (approved):

portal/styles/globals.css
ArturDolzan: Fair point, and it is a trade we took knowingly (it is called out in the description). One thing though: with `max-age=0, must-revalidate` the browser revalidates on every load, so a stale woff2 cannot actually pair with new percentages. The header quoted as the cost is also what removes that risk.

Where it does bite is the day we add long-lived cache headers for `/fonts`, which is the obvious next step.

portal/styles/globals.css
ArturDolzan: Good catch, this one was real. Scoped the reset to `body` in [`e8e04c23`](https://github.com/hemilabs/ui-monorepo/pull/2211/commits/e8e04c23). `walletQRCodeView.tsx:109` is the only heading in the app with a nested element, so that was the single site affected.

One note on the alternative: `:where(*)` would not have fixed it. Specificity only decides between declarations on the same element, and inheritance only applies when no declaration matches at all, so `:where(*)` still declares font-family on the span and still beats the h4. Only scoping it works.

portal/index.html
ArturDolzan: Numbers check out (672 KiB, same figure). The redundancy argument is only half right though: preload prevents the flash of fallback text, the metric overrides prevent the layout shift. Different problems, so dropping the preloads trades one for the other rather than removing something duplicated.

Cutting to InterVariable alone is the version I would take, it saves 328 KiB and headings would swap without shifting. But the app does not render yet (the entry is still a placeholder), so there is no waterfall to check it against. I would rather do it in the routing PR with something to measure. Happy to open an issue so it does not get lost.

PR #2213, opened 2026-08-23
Hide BTC tunnel fees until amount is entered

### LINKED ISSUE #2128: (Tunnel) — Show fee items only when an amount is entered
From [ Hemi ] to [ Bitcoin ]

Fee breakdown items are currently visible even when the input is empty, showing ( `-` ) as a placeholder.

Hide the fee items until the user enters an amount. Once an amount is inputted, reveal the breakdown with the calculated values. (same behaviour as (from or to) ethereum networks)

Image

Godbrand0:
## Summary

Closes #2128.

Fee breakdown items were visible even with an empty amount, showing `-` placeholders — the Bitcoin tunnel flows didn't gate the fees section on whether the form could actually submit, unlike the EVM withdrawal flow (`EvmWithdraw`'s `RenderBelowForm`), which already hides its fees section via `!canWithdraw`.

Applied the same pattern to both Bitcoin tunnel directions:
- **Hemi → Bitcoin** (`withdraw.tsx`, `BtcWithdraw`): the fees block (`HemiBtcFeesSummary` + `EvmFeesSummary`, i.e. "Tunnel fees" / "Hemi gas fee" / "Total") is now wrapped in a `RenderFees` helper gated by the existing `canWithdraw` flag. The receiving-address section is unaffected.
- **Bitcoin → Hemi** (`btcDeposit.tsx`, `BtcDeposit`): same issue existed in this direction too (confirmed with a screenshot showing "Network fees" / "Tunnel fees" visible with an empty/zero amount). `BtcFees` is now gated by the existing `canDeposit` flag the same way.

## Test plan

- [x] `pnpm test` in `portal` (844 tests passing)
- [x] `tsc --noEmit` shows no new errors on the touched files
- [x] Verified locally: fee breakdown stays hidden on both the withdraw (Hemi → Bitcoin) and deposit (Bitcoin → Hemi) forms until a valid amount is entered

image

### DIFF
--- a/portal/app/[locale]/tunnel/_components/btcDeposit.tsx
+++ b/portal/app/[locale]/tunnel/_components/btcDeposit.tsx
@@ -176,7 +176,7 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
   }
 
   function renderFees() {
-    if (!canDeposit) {
+    if (!canDeposit || amountBigInt === BigInt(0)) {
       return null
     }
 
--- a/portal/app/[locale]/tunnel/_components/btcDeposit.tsx
+++ b/portal/app/[locale]/tunnel/_components/btcDeposit.tsx
@@ -175,7 +175,7 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
     )
   }
 
-  function RenderFees() {
+  function renderFees() {
     if (!canDeposit) {
       return null
     }
@@ -197,7 +197,7 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
               },
             )}
           />
-          <RenderFees />
+          {renderFees()}
         </div>
       }
       bottomSection={<WalletsConnected />}
--- a/portal/app/[locale]/tunnel/_components/btcDeposit.tsx
+++ b/portal/app/[locale]/tunnel/_components/btcDeposit.tsx
@@ -175,6 +175,14 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
     )
   }
 
+  function RenderFees() {
+    if (!canDeposit) {
+      return null
+    }
+
+    return <BtcFees amount={amountBigInt} />
+  }
+
   return (
     <TunnelForm
       belowForm={
@@ -189,7 +197,7 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
               },
             )}
           />
-          <BtcFees amount={amountBigInt} />
+          <RenderFees />
         </div>
       }
       bottomSection={<WalletsConnected />}
--- a/portal/app/[locale]/tunnel/_components/withdraw.tsx
+++ b/portal/app/[locale]/tunnel/_components/withdraw.tsx
@@ -202,6 +202,26 @@ const BtcWithdraw = function ({ state }: BtcWithdrawProps) {
     return t('tunnel-page.submit-button.initiate-withdrawal')
   }
 
+  function RenderFees() {
+    if (!canWithdraw) {
+      return null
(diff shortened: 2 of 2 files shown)


gndelia (dismissed):

portal/app/[locale]/tunnel/_components/btcDeposit.tsx
```diff
@@ -189,7 +197,7 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
               },
             )}
           />
-          <BtcFees amount={amountBigInt} />
+          <RenderFees />
```
Comment (ArturDolzan) [inline portal/app/[locale]/tunnel/_components/btcDeposit.tsx]: Perhaps it's worth checking `<RenderFees />` here. `RenderFees` is a new function on every render, so React remounts the subtree and the fee queries refetch each time. Maybe calling `{renderFees()}` like `evmDeposit.tsx` does would be safer?

Comment (Godbrand0) [reply portal/app/[locale]/tunnel/_components/btcDeposit.tsx]: Good catch, thanks — switched both to the plain-function pattern (`renderFees()` called inline as `{renderFees()}`), matching evmDeposit.tsx's `renderBelowForm()`, in both btcDeposit.tsx and withdraw.tsx (I'd used the same JSX-component style in both). Fixed in 2659b682.

portal/app/[locale]/tunnel/_components/withdraw.tsx
```diff
@@ -202,6 +202,26 @@ const BtcWithdraw = function ({ state }: BtcWithdrawProps) {
     return t('tunnel-page.submit-button.initiate-withdrawal')
   }
 
+  function renderFees() {
+    if (!canWithdraw) {
```
Comment (ArturDolzan) [inline portal/app/[locale]/tunnel/_components/withdraw.tsx]: Perhaps it's worth checking the case where only the EVM wallet is connected. `canWithdraw` can be true without `btcAddress`, and then `gas.isError` forces `-` on both the gas and total rows. Wonder if gating on `feeEstimationEnabled` would cover it.

portal/app/[locale]/tunnel/_components/btcDeposit.tsx
```diff
@@ -175,6 +175,14 @@ export const BtcDeposit = function ({ state }: BtcDepositProps) {
     )
   }
 
+  function renderFees() {
+    if (!canDeposit) {
```
Comment (ArturDolzan) [inline portal/app/[locale]/tunnel/_components/btcDeposit.tsx]: Not sure, but maybe the debounced `amountBigInt` is worth a quick look here. When the panel first appears it can still be `0n` for ~300ms, so the tunnel fee row shows `-` before the skeleton. Could adding `amountBigInt > 0` to the gate smooth that out?

portal/app/[locale]/tunnel/_components/withdraw.tsx
Comment (Godbrand0) [reply portal/app/[locale]/tunnel/_components/withdraw.tsx]: Good catch — gated renderFees on feeEstimationEnabled instead of canWithdraw, so the fees block (including the gas row) only renders once we can actually estimate the gas fee (amount valid + btcAddress connected). Fixed in 88ad3926.

portal/app/[locale]/tunnel/_components/btcDeposit.tsx
Comment (Godbrand0) [reply portal/app/[locale]/tunnel/_components/btcDeposit.tsx]: Good catch — added amountBigInt === BigInt(0) to the renderFees gate in btcDeposit.tsx, so the fee row stays hidden during the ~300ms debounce window instead of flashing '-'. Fixed in 88ad3926.

ArturDolzan (approved):

arielmergen (approved):

gndelia (approved):
LGTM - thank you @Godbrand0  for your contribution!

PR #1864, opened 2026-04-09
Add animation to all drawers

arielmergen:
### Description

Add smooth drawer open/close: panel slide (bottom on mobile, left/right on desktop) and backdrop fade via CSS; overlay fade on the backdrop wrapper.

Fix spurious close: only run onClose after a user-initiated close when the panel’s transform transition ends; ignore the opening click as “outside” so it doesn’t close immediately.

Portal drawers to document.body to avoid clipping by the main layout container.

Connect wallets: one Drawer shell in the parent, static ConnectWalletsDrawer import (no nested dynamic + second drawer); tune mobile height (max-h-[95dvh]).

Remove DrawerLoader: avoid double Drawer when lazy chunks mount their own Drawer; use static imports for those shells (tunnel, stake, staking dashboard, custom token, hemi-earn vaultForm, etc.) and Spinner + drawer-content where content loads lazily inside one drawer (e.g. BTC yield). Full-bleed DrawerOverlay when portaled (no md:rounded-lg on the dim layer).

Backdrop interaction: pointer-events + isClosing so the backdrop doesn’t block clicks when hidden, but still blocks during the close animation.

Mobile/tablet nav: one Drawer in AppLayout, NavbarResponsive supplies menu content only; no dynamic + loader for the navbar. Close the nav drawer when viewport crosses xl so an empty shell isn’t left after resize.

### Screenshots

https://github.com/user-attachments/assets/19b48c79-fa99-43e5-b1e6-dc7c5cd3276a

### Related issue(s)

Closes #1583 
Fixes #
Related to #

### Checklist

- [ ] Manual testing passed.
- [ ] Automated tests added, or N/A.
- [ ] Documentation updated, or N/A.
- [ ] Environment variables set in CI, or N/A.

PR #2214, opened 2026-08-24
Fix Wallet Connect connection

gabmontes:
### Description

Fixes two bugs in the EVM connect flow:

- WalletConnect QR never rendered. `EvmWallet` returned a loading skeleton whenever `status === 'connecting'`, unmounting the QR view the instant it started the pairing — but WalletConnect is always "connecting" while waiting for a scan. The pairing was then torn down by the unmounted view's cleanup, eventually surfacing as `Proposal expired`. This regressed in #1864 (which reordered the render branches so `connecting` shows the skeleton instead of the accordion); #1910's 10s abort timer later compounded it by also killing the pairing mid-scan. Now the skeleton and the abort timer are skipped while the QR view is open, so `connecting` keeps the QR mounted. Direct connectors keep the skeleton + abort behavior unchanged.

- Connect then instant disconnect. After a successful scan the QR view unmounts (switching to the connected state) and its cleanup called `disconnect()` unconditionally, dropping the fresh session. Cleanup now only disconnects when the account isn't already connected.

This PR also fixes a UI issue: WalletConnect/Binance showed an "Install" hover label on desktop even though there's nothing to install. They now show a "Scan" label with a QR icon.

### Screenshots

https://github.com/user-attachments/assets/9a824e8b-77bd-465a-a851-532db0049ef4

### Checklist

- [x] Manual testing passed.


portal/components/connectWallets/walletQRCodeView.tsx
```diff
@@ -90,11 +92,16 @@ export function WalletQRCodeView({ onBack, wallet }: Props) {
...
+        // connecting. After a successful scan the account is already connected,
+        // so disconnecting here would drop the freshly established session.
+        if (getAccount(config).status !== 'connected') {
+          walletConnectConnector.disconnect?.()
+        }
         reset()
       }
     },
-    [connect, connectors, reset],
+    [config, connect, connectors, reset],
```
gndelia, line 109: I am unsure if `config` is a stable object (if not, it could cause issues with effects), but maybe you can call 

```tsx
const configStatus = getAccount(config).status
```

outside of the effect and pass that status (a string) to the effect

portal/components/connectWallets/walletQRCodeView.tsx
```diff
@@ -90,11 +92,16 @@ export function WalletQRCodeView({ onBack, wallet }: Props) {
       connect({ connector: walletConnectConnector })
 
       return function cleanup() {
-        walletConnectConnector.disconnect?.()
+        // Only tear down the pairing when the user leaves the QR view without
+        // connecting. After a successful scan the account is already connected,
+        // so disconnecting here would drop the freshly established session.
+        if (getAccount(config).status !== 'connected') {
+          walletConnectConnector.disconnect?.()
+        }
```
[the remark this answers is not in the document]
gabmontes: Good catch — fixed in 329205af. Switched from the global `getAccount(config).status` to checking this connector's own active connection: `getConnections(config).some(c => c.connector.uid === walletConnectConnector.uid)`. Now the pending pairing is only kept when *this* WalletConnect connector connected, so a different connector reconnecting no longer leaks it. — Gabi 🤖

portal/components/connectWallets/walletQRCodeView.tsx
gabmontes: Thanks. I kept the read inside the cleanup on purpose: we need the value at unmount time, not at render time, so capturing `getAccount(config).status` during render would read the stale "disconnected" state. `config` from `useConfig()` is a stable singleton (created once via `createConfig`), so it's safe in the deps and won't retrigger the effect. Also switched the check to `getConnections` per Copilot's note (329205af). — Auto Gabi

gndelia (approved):

ArturDolzan (approved):

PR #2215, opened 2026-08-25
Serve the portal from react-router

ArturDolzan:
### Description

This PR replaces the Next app router with react-router across three commits, and it is the first half of the routing step. Only the tunnel routes are wired here; the other nine pages follow in a second PR once this pattern has been reviewed.

**The point of this one is that nothing had actually run yet.** The Vite entry rendered a placeholder, so no `page.tsx` was ever in the build graph, and the toolchain, env var, `next/dynamic` and `next/font` PRs before it had never been exercised by a real bundle. Mounting the router is what finally pulls in the five web workers, the node polyfills and the wallet contexts. That is also how the build broke for the first time: `packages/btc-wallet` reaches for the `Buffer` global, and the polyfill shim only resolves from the portal, which is what the alias in `vite.config.ts` fixes.

The root layout was a server component (async, `generateMetadata`, `generateStaticParams`, `setRequestLocale`). Its provider stack becomes a route layout, messages load through a cached dynamic import consumed with `use()`, and `lang` moves to an effect now that `<html>` lives in `index.html`.

**Three behaviours had no equivalent once Next was gone**, and dropping them would have been silent. Per-route titles are registered through a context rather than written directly, because effects run child first and a layout would always overwrite the route. Two nested error boundaries contain a failed chunk load, which the `error.tsx` files used to catch and the Vite entry never mounts. And the scroll container resets on navigation, which the Next router did for free.

The trailing slash is worth flagging for reviewers: Next's static export produced `/tunnel/`, so call sites compared against paths with a trailing slash. react-router pathnames have none, which made the tunnel tabs disappear on `/tunnel` and highlight nothing on the history page. `isSamePathOrUnder` replaces the raw `startsWith` and, as a side effect, stops the navbar lighting up both stake entries on `/staking-dashboard`.

## Notes

`SentryRoutes` is not here: `@sentry/nextjs` does not export the react-router instrumentation, so route spans arrive with the `@sentry/react` migration. `next` itself also stays for now, since Sentry declares it as a peer dependency.

Nothing about the workers could be validated at runtime without an environment, so instead the build output was checked against the failure mode #2194 documents: all five pass `{ type: 'module'}`, no bare imports survive in the worker bundles, and the only remaining `global` reference is a property probe rather than the module-scope read that used to throw.

`sitemap.xml` is gone with `app/sitemap.ts`. It was a Next-only mechanism that already produced nothing under Vite, but it does mean crawlers lose the per-locale alternates, and replacing it is its own piece of work.

Bookmarks to non-localized paths (`/stake`, `/get-started`) now land on the tunnel instead of their page. Those destinations do not exist yet either way, so the redirect table comes back with the routes.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2194 

### Checklist

- [x] Manual testing passed.


portal/i18n/navigation.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const locale = useLocale()
+  const { pathname } = useLocation()
+  const prefix = `/${locale}`
+
+  if (pathname === prefix) {
+    return '/'
+  }
+  return pathname.startsWith(`${prefix}/`)
+    ? pathname.slice(prefix.length)
+    : pathname
```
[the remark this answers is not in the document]
ArturDolzan: Good catch, and it is the same class as the tab bug earlier in this PR. The old static export redirected to `/en/tunnel/`, so that trailing slash is what sits in existing bookmarks. Normalized at the source in [`bce7e21f`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/bce7e21f), and pulled into a tested `unlocalizedPathname` since this path logic has bitten us twice now.

portal/app/[locale]/_components/appLayout.tsx
```diff
@@ -51,6 +53,15 @@ export const AppLayout = function ({ children }: Props) {
...
   const { closeDrawer, isDrawerOpen } = useDrawerContext()
+  const { key: locationKey } = useLocation()
+  const scrollContainer = useRef<HTMLDivElement>(null)
+
+  useEffect(
+    function scrollToTopOnNavigation() {
+      scrollContainer.current?.scrollTo({ top: 0 })
+    },
+    [locationKey],
+  )
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/_components/appLayout.tsx]: You are right. I had switched this to the location key on an earlier suggestion and did not verify it. `operation` and `txHash` are nuqs `useQueryState`, which was shallow under Next and never scrolled, so keying on the whole location was yanking the view to the top mid-form. Back to pathname in [`bce7e21f`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/bce7e21f).

portal/components/error500/index.tsx
```diff
@@ -79,3 +60,21 @@ export const LocalizedError500 = function (
...
+export const UntranslatedError500 = (props: Pick<Props, 'reset'>) => (
+  <Error500
+    {...props}
+    description={
+      <>
+        An unexpected error has occurred. Please try again or
+        <ExternalLink className="hoverable-text" href={discordUrl}>
+          {' '}
+          contact us{' '}
+        </ExternalLink>
```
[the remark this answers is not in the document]
ArturDolzan: I went back and forth on this one and ended up keeping it English-only. Bundling all three locales so the provider could sit above the boundary cost about 17 KB gzipped, and it still picked the locale from the browser rather than the URL, so a shared `/es/...` link could have rendered English anyway. It also pushed the boundary out of the outermost position, which meant a throw while reading the browser language had nothing to catch it.

So the fallback stays inlined the way vetro's ErrorPage falls back, and the boundary is outermost again. The copy duplication is real, and it gets revisited when this collapses into `instrument.ts` with the `@sentry/react` migration in #2194.

gndelia:
image


😄 🤖

portal/app.tsx
```diff
@@ -0,0 +1,66 @@
...
+      to={{ hash, pathname: `/${preferredLocale()}`, search }}
+    />
+  )
+}
+
+const ToTunnel = function () {
+  const { hash, pathname, search } = useLocation()
+
+  return (
+    <Navigate replace to={{ hash, pathname: `${pathname}/tunnel`, search }} />
```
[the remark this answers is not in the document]
ArturDolzan: Right, and it is the same trailing slash again: I normalized it in `usePathname` but this redirect reads the raw location. Trimmed before appending in [`5c1a8087`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/5c1a8087), along with the unknown-locale redirect, which was passing its slash through to the next comparison layer.

portal/utils/url.ts
```diff
@@ -15,3 +18,75 @@ export const queryStringObjectToString = function (
...
+      pathname,
+      search: prefixed(searchParts.join('?'), '?'),
+    }
+  }
+
+  return {
+    hash: prefixed(href.hash ?? '', '#'),
+    pathname: href.pathname ?? '',
+    search: prefixed(toSearchParams(href.query).toString(), '?'),
+  }
```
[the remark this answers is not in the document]
ArturDolzan: Good catch, the type promised it and the conversion ignored it. I went with narrowing rather than honouring it: `components/link` merges `networkType` into `query`, so supporting both would mean one of them silently losing. Passing `search` is a compile error now, in [`b16e6f45`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/b16e6f45).

ArturDolzan:
> image
> 😄 🤖

Where? 😆

arielmergen (approved):

portal/utils/url.ts
```diff
@@ -15,3 +18,77 @@ export const queryStringObjectToString = function (
...
+}
```
gndelia, line 69: Does this [react-router utility help](https://reactrouter.com/api/utils/parsePath) ?

portal/utils/url.ts
```diff
@@ -15,3 +18,77 @@ export const queryStringObjectToString = function (
...
+
+// Next's static export produced trailing slashes, so call sites used to compare
+// against `/tunnel/`. react-router pathnames have none, which silently made
+// those checks miss the section root.
```
gndelia: I don't think we should reference Next's here. Let's simplify or drop

portal/utils/url.ts
```diff
@@ -15,3 +18,77 @@ export const queryStringObjectToString = function (
...
+export const isSamePathOrUnder = (pathname: string, base: string) =>
+  pathname === base || pathname.startsWith(`${base}/`)
+
+// Strips the locale segment and any trailing slash. Next's static export
+// produced both, so bookmarks from before the migration still carry a trailing
+// slash that every comparison against a plain path would miss.
```
gndelia: Same here as https://github.com/hemilabs/ui-monorepo/pull/2215/changes#r3855905443

portal/app/[locale]/_components/analytics.tsx
```diff
@@ -14,14 +15,21 @@ export const Analytics = function ({
...
   const locale = useLocale()
 
-  const removeLocaleAndTrailingSlash = (url: string) =>
-    (url.endsWith('/') ? url.slice(0, -1) : url).replace(`/${locale}`, '')
+  const processUrl = useCallback(
+    function stripLocale(url: string) {
+      const [pathname = '', ...rest] = url.split('?')
+      const search = rest.join('?')
+      const path = unlocalizedPathname(pathname, locale)
+      return search ? `${path}?${search}` : path
```
Comment (gndelia) [inline portal/app/[locale]/_components/analytics.tsx]: `locale` is always at the beginning. Wouldn't it be simpler to split and drop the first path before the first `/` ?

portal/app/[locale]/ecosystem/layout.tsx
```diff
@@ -1,10 +1,8 @@
-import type { Metadata } from 'next'
+import { useDocumentTitle } from 'hooks/useDocumentTitle'
 import { PropsWithChildren } from 'react'
 
-export const metadata: Metadata = {
-  title: 'Ecosystem | Hemi Portal',
-}
-
 export default function EcosystemLayout({ children }: PropsWithChildren) {
+  useDocumentTitle('Ecosystem | Hemi Portal')
```
Comment (gndelia) [inline portal/app/[locale]/ecosystem/layout.tsx:5]: Given this is on the client side - perhaps now we can set the appropriate title considering the language (something we couldn't do before with Next). I don't know what the best practices for this are

portal/hooks/useDocumentTitle.tsx
```diff
@@ -0,0 +1,52 @@
...
+  defaultTitle,
+}: Props) {
+  const [title, setTitle] = useState<string>()
+
+  useEffect(
+    function applyDocumentTitle() {
+      document.title = title ?? defaultTitle
+    },
+    [defaultTitle, title],
+  )
```
gndelia, line 33: I think this effect can be dropped in favor of a regular function

```ts
const updateTitle = function (newTitle: string) {
  setTitle(newTitle)
  document.title = title ?? defaultTitle
}
```

and then export this on the Context

portal/hooks/usePathnameWithoutLocale.ts
```diff
@@ -1,14 +1,10 @@
...
+ * current pathname, without changing the locale (for example, adding a
+ * particular query string).
+ * @returns The pathname, locale segment and trailing slash removed
  */
-export const usePathnameWithoutLocale = function () {
-  const locale = useLocale()
-  // replace only replaces the first match, which should always be at the beginning
-  return usePathname().replace(`/${locale}`, '')
-}
+export const usePathnameWithoutLocale = usePathname
```
gndelia, line 10: Seems like we should just drop `usePathnameWithoutLocale` in favor of `usePathname`. But I'm ok if you are going to do it in a different PR - just don't forget

portal/i18n/messages.ts
```diff
@@ -0,0 +1,25 @@
...
+    module => module.default as Messages,
+  )
+  // Drop a rejected load, otherwise the cache replays the failure forever and
+  // the error page's retry can never recover.
+  loading.catch(function forgetFailedLoad() {
+    pending.delete(locale)
+  })
+  pending.set(locale, loading)
+  return loading
+}
```
gndelia, line 25: are we not using the same translation libs as in vetro for any reason? or this is temporal?

portal/i18n/routing.ts
```diff
@@ -1,3 +1,4 @@
+import { hasLocale } from 'next-intl'
```
gndelia, line 1: Sounds like this is temporal because we won't be including next-intl in the future 🤔

portal/main.tsx
```diff
@@ -1,13 +1,51 @@
+// Next auto-loaded this by convention and Vite does not, so without the import
+// no Sentry client is ever configured and every captureException is a no-op.
```
gndelia: Same as before re: comments and next

portal/utils/url.ts
ArturDolzan: Nice, that is exactly it. Swapped in [`78e27bf8`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/78e27bf8), and the existing tests pass against it unchanged, including the odd ones with a second `?` inside the search or a second `#` inside the hash.

portal/utils/url.ts
ArturDolzan: Agreed, that is archaeology once this lands. Dropped in [`78e27bf8`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/78e27bf8).

portal/utils/url.ts
ArturDolzan: Same, rewritten in [`78e27bf8`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/78e27bf8) to just say that older bookmarks still carry a trailing slash, without naming where they came from.

portal/app/[locale]/_components/analytics.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/_components/analytics.tsx]: Good call, it was splitting the query by hand for no reason. Uses `parsePath` now in [`78e27bf8`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/78e27bf8).

I left `unlocalizedPathname` itself alone though: dropping the first segment unconditionally would turn `/entities` into `tities` when the locale is `en`, so it checks that the segment actually is the locale. There is a test for that one.

portal/main.tsx
ArturDolzan: Agreed. Rewritten in [`5074bceb`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/5074bceb) to say what the import does rather than which framework used to do it for us. That is the part worth keeping, since deleting that line silently turns every captureException into a no-op.

portal/hooks/useDocumentTitle.tsx
ArturDolzan: I tried it and it does not quite hold. With the provider reduced to a callback, `defaultTitle` only ever gets written by the cleanup, so a route that claims no title keeps whatever `index.html` shipped instead of the localized default. The 404 route is one of those today.

The state is what lets the provider know whether anything claimed the title, so I kept the shape and just tightened the comment to say that, in [`5074bceb`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/5074bceb).

portal/app/[locale]/ecosystem/layout.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/ecosystem/layout.tsx:5]: We went back and forth on this one and landed on keeping them as production serves them. Translating would be the only thing in this PR changing behaviour rather than porting it, and the migration does not need it.

Happy to open an issue so we can pick it up properly once the routes are all in, since it also means adding four new keys across the three locales.

portal/hooks/usePathnameWithoutLocale.ts
ArturDolzan: Fair, and it is overdue. I will do it in the stacked PR (#2217) rather than here, since that one is already the cleanup pass and this branch is deep in review. Will ping you when it lands.

portal/i18n/messages.ts
ArturDolzan: Temporal, yes, but the cheap path out is probably not the one the issue assumes.

`next-intl` is a wrapper around `use-intl`, and everything the portal touches (`useTranslations`, `IntlProvider`, `useLocale`, `hasLocale`) already comes from there, so the client side is framework-neutral today. That gives two very different ways to converge with web/. Going to i18next rewrites the call sites in 229 files and drops the typed messages, since `intl.d.ts` makes a wrong key a compile error. Dropping to `use-intl` leaves those 229 files untouched and only removes the package plus the `defineRouting` in `i18n/routing.ts`.

Same shape as the umami discussion in #2194: the portal has typing web/ does not, and the conclusion there was to keep it rather than adopt the untyped version. Worth deciding before phase 3, since the issue currently says i18next. Do you know if there was a reason behind that pick on the web side, or was it just what came first?

portal/i18n/routing.ts
ArturDolzan: Same answer as [the one on messages.ts](https://github.com/hemilabs/ui-monorepo/pull/2215#discussion_r3856044293): `defineRouting` is the only Next-shaped thing left in this file, and it goes away with the package. Which package replaces it is the open question there.

portal/i18n/navigation.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const usePathname = function () {
+  const locale = useLocale()
+  const { pathname } = useLocation()
+
+  return unlocalizedPathname(pathname, locale)
+}
+
+export const useRouter = function () {
+  const locale = useLocale()
+  const navigate = useNavigate()
```
gabmontes, line 25: `BrowserRouter` already wraps every navigation in `startTransition` by default via `NavigationContext` (`useTransitions` defaults to `true`), so the explicit `startTransition` calls in `push` and `replace` are redundant. The same applies to the comment above — `RouterLink` goes through the same path, so there is no asymmetry between `Link` and `useRouter`.

*Auto Gabi*

portal/i18n/messages.ts
gndelia: ok. Not really fan of temporal code. I'd prefer to leave CI broken rather than adding code that will be soonish removed. But I'm ok if you go this way. Please do not forget of this temporal changes

portal/i18n/navigation.tsx
```diff
@@ -0,0 +1,57 @@
...
+      replace: (href: Href, options?: NavigateOptions) =>
+        startTransition(() =>
+          navigate(withLocale(href, options?.locale ?? locale), {
+            replace: true,
+          }),
+        ),
+    }),
+    [locale, navigate],
+  )
+}
```
gndelia, line 41: I must assume all of these hooks are temporal until the appropiate lib for intl is added right?

portal/app.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <UntranslatedError500 reset={() => window.location.replace('/')} />
+      </div>
+    }
+  >
+    <BrowserRouter>
+      <NuqsAdapter>
+        <Routes>
+          <Route element={<ToPreferredLocale />} path="/" />
+          <Route element={<LocaleLayout />} path="/:locale">
+            <Route element={<ToTunnel />} index />
```
gndelia, line 59: I think here we're missing the `enableHemiEarnPage` feature flag check - if enabled, we should redirect to that page instead of to the tunnel by default (just like staging does now)

portal/app/[locale]/_components/analytics.tsx
```diff
@@ -14,14 +16,19 @@ export const Analytics = function ({
 }: Pick<ComponentProps<typeof UmamiAnalyticsProvider>, 'children'>) {
   const locale = useLocale()
 
-  const removeLocaleAndTrailingSlash = (url: string) =>
-    (url.endsWith('/') ? url.slice(0, -1) : url).replace(`/${locale}`, '')
+  const processUrl = useCallback(
+    function stripLocale(url: string) {
```
Comment (gndelia) [inline portal/app/[locale]/_components/analytics.tsx:20]: I think the `url` here comes from `next/navigation` inside the lib. Nothing to do here, but let's take note of this because we need to fix it (maybe in the lib itself?)

portal/i18n/navigation.tsx
ArturDolzan: You are right, I went and checked the source. `BrowserRouter` sets its state with `useTransitions === false ? setStateImpl : startTransition(...)`, we never pass the prop, and the call hangs off `history.listen`, so `Link` goes through the very same path. The asymmetry my comment claimed does not exist.

Both calls and the comment are gone in [`f42b64d5`](https://github.com/hemilabs/ui-monorepo/pull/2215/commits/f42b64d5). Thanks, this also undoes a workaround I had added for a blank screen on locale change that the router was already preventing.

portal/i18n/navigation.tsx
ArturDolzan: Less than it looks. The hooks themselves are react-router now, and the only next-intl touchpoint left in that file is `useLocale`, which `use-intl` exports too. So whichever way the intl decision goes, this survives with one import line changed.

portal/app/[locale]/_components/analytics.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/_components/analytics.tsx:20]: Yes, exactly that. `usePathname` from `next/navigation` returns `null` outside Next's router, so every view ends up reported as the landing URL. I added a note under the Umami checkbox in #2194 so it reads as something already broken rather than a pending refactor.

portal/app.tsx
ArturDolzan: Already done, just in the stacked PR (#2217). I left it out here on purpose because `/hemi-earn` is not routed in this branch yet, so honouring the flag would send people to the 404.

One thing worth flagging while we are here: I reproduced what main does today, where the site root honours the flag and the locale index always goes to the tunnel. Those two disagree in production. Happy to make both honour it if you prefer, it is just a behaviour change rather than a port.

portal/app/[locale]/layout.tsx
```diff
@@ -1,104 +1,86 @@
...
+export const LocaleLayout = function () {
+  const { locale } = useParams()
+  const { hash, pathname, search } = useLocation()
 
   if (!hasLocale(routing.locales, locale)) {
-    notFound()
-    return undefined
+    // Keep everything after the unknown locale, so a link shared with a locale
+    // we do not support still lands on the page it pointed at.
+    const rest = pathname.replace(/^\/[^/]*/, '').replace(/\/+$/, '')
```
Comment (gndelia) [inline portal/app/[locale]/layout.tsx:72]: This wasn't supported, and I wouldn't add support to it. Let's make people share the appropriate links, heh

gndelia (approved):
left a minor suggestion, but ok to move forward (?)

portal/app/[locale]/layout.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/layout.tsx:72]: Agreed, and it is already gone. #2217 replaced this with a fallback that treats an unknown first segment as a path instead of a locale, so `/fr/tunnel` reaches the 404 again just like the static export did, in [`1fc3c95c`](https://github.com/hemilabs/ui-monorepo/pull/2217/commits/1fc3c95c). It only survives on this branch until that one lands.

gabmontes (approved):

PR #2217, opened 2026-08-25
Route the remaining portal pages

ArturDolzan:
### Description

This PR wires the eight pages the routing PR left out, so every navbar destination reaches a page instead of the catch-all 404. It stacks on #2215, which is where the risky part lives, and this one is mostly the same pattern repeated.

**The one real conversion is the pool page.** It was the last file in the portal using a server API: an async component awaiting `params` plus `generateStaticParams` to emit one page per known share at build time. A dynamic route matches any string, and without a guard the value reaches viem's `isAddressEqual`, which throws on a non-address. So `/hemi-earn/pool/foo` would have rendered the error page and reported itself to Sentry, where the static export just answered with the 404. It now checks `isAddress` first.

The other seven pages are already client components and moved as they were. The six layouts swap `children` for `<Outlet />`, with two needing care: genesis-drop passes content through a component that gates on testnet, and hemi-earn through the provider that follows its status updaters.

`/demos` turned out not to be a page at all, just a client-side redirect to `/ecosystem`, so it became a route that navigates.

**The non-localized entry paths come back**, since their destinations exist again. Worth knowing why the seemingly redundant entries matter: without one, `/get-started` matches `/:locale` with the segment read as a language, fails the check and lands on the tunnel. Verified with `matchRoutes`. Two of the seven also rewrite the path, one of them fixing a folder name that shipped misspelled.

## Notes

The feature flag needed no wiring, since the hemi-earn layout already answers with the 404 when it is off. The site root honours it and the locale index does not, which is how production behaves today, so it is reproduced rather than reconciled.

Titles stay in English exactly as production serves them. `/staking-dashboard` and `/ecosystem` without a locale land on the tunnel, because Next had no redirect for either; they used to 404, so that is a difference worth knowing even though adding the missing aliases would be a change the migration does not need.

Two dead Next entry points went with this: `instrumentation.ts` (a server request hook this app never had) and `onRouterTransitionStart` (an App Router hook). Neither has run since the toolchain swap.

Also here: a malformed URL fix. The stake dashboard builds links as one string with path and query together, `components/link` wraps that into an object to inject `networkType`, and the conversion appended a second `?`. The click handler preventDefaults, which is why nothing broke visibly.

### Screenshots

N/A - No visual changes intended, the pages should look as they do today.

### Related issue(s)

Related to #2194

### Checklist

- [x] Manual testing passed.


portal/utils/url.ts
```diff
@@ -54,21 +54,33 @@ const prefixed = function (value: string, character: string) {
...
+    return splitPath(href)
   }
 
+  const fromPathname = splitPath(href.pathname ?? '')
+  const query = toSearchParams(href.query)
+  new URLSearchParams(fromPathname.search).forEach(function (value, key) {
+    if (!query.has(key)) {
+      query.append(key, value)
+    }
+  })
```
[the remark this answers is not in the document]
ArturDolzan: Good catch, and it was self-inflicted: `query.has(key)` turns true as soon as the first embedded value is appended, so every later value for that key was dropped. Snapshotting the keys before the merge in [`330f1a40`](https://github.com/hemilabs/ui-monorepo/pull/2217/commits/330f1a40), with tests for both the repeated case and the override.

portal/app/[locale]/hemi-earn/pool/[shareAddress]/page.tsx
```diff
@@ -1,24 +1,16 @@
...
+import { isAddress } from 'viem'
+
+import NotFound from '../../../not-found'
 
 import { PoolPageContent } from './_components/poolPageContent'
 
-type Props = {
-  params: Promise<{ shareAddress: string }>
-}
+export default function Page() {
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/pool/[shareAddress]/page.tsx]: When `shareAddress` is invalid the pool page returns `<NotFound />` as the `<Outlet>` child, but `LocalEarnOperationsProvider` and `EarnStatusUpdaters` are still mounted by the parent layout — including the earn delivery watcher that polls and holds subscriptions. A 404 screen shouldn't be running earn-specific side effects. Consider validating the address earlier (e.g. in the layout, before mounting the provider) so those hooks never fire for a bad URL.

— Auto Gabi

portal/app.tsx
```diff
@@ -15,13 +30,26 @@ import {
...
-const ToPreferredLocale = function () {
+// The paths the Next redirect pages served, kept so existing bookmarks and
+// shared links still land where they did. `/stake` and the `staking-dashbord`
+// typo were aliases, not plain locale prefixes.
+const legacyPaths = {
+  '/demos': '/ecosystem',
+  '/genesis-drop': '/genesis-drop',
+  '/get-started': '/get-started',
+  '/hemi-earn': '/hemi-earn',
+  '/stake': '/stake/dashboard',
```
gabmontes, line 37: `legacyPaths` covers `/staking-dashbord` (the legacy typo) but not the correct `/staking-dashboard`, and `/ecosystem` is also missing. Both unguarded paths match `/:locale` with the segment misread as a locale code, fail the locale check, and silently redirect to the tunnel — which is harder to debug than a 404. The PR description notes this as a known difference, but since the correct paths are already routed under `/:locale`, adding two more entries to `legacyPaths` would be a one-liner fix each.

— Gabi 🤖

portal/app/[locale]/hemi-earn/_fetchers/fetchHemiEarnAssetConfigs.ts
```diff
@@ -23,7 +23,6 @@ export type HemiEarnAssetConfig = AssetData & {
 // One representative config per share OFT. A share can accept multiple deposit
 // assets (e.g. USDC + USDT → sVUSD), so `share`/`remoteShare` repeat across
 // configs while `asset` does not — consumers that key off the share (pools,
-// TVL, static params) must dedupe first or they double-count.
```
Comment (gndelia) [inline portal/app/[locale]/hemi-earn/_fetchers/fetchHemiEarnAssetConfigs.ts:26]: this just leaves the comment hanging in `(pools,`

gndelia (changes requested):

portal/app.tsx
```diff
@@ -1,8 +1,23 @@
...
+import HemiEarnLayout from 'app/[locale]/hemi-earn/layout'
+import HemiEarnPage from 'app/[locale]/hemi-earn/page'
+import PoolPage from 'app/[locale]/hemi-earn/pool/[shareAddress]/page'
 import { LocaleLayout } from 'app/[locale]/layout'
 import NotFound from 'app/[locale]/not-found'
+import StakeDashboardPage from 'app/[locale]/stake/dashboard/page'
+import StakeLayout from 'app/[locale]/stake/layout'
+import StakePage from 'app/[locale]/stake/page'
+import StakingDashboardLayout from 'app/[locale]/staking-dashboard/layout'
+import StakingDashboardPage from 'app/[locale]/staking-dashboard/page'
```
gndelia: We were using default exports due to Next, but we can now use named exports 😄

portal/app.tsx
```diff
@@ -15,13 +30,26 @@ import {
...
+// typo were aliases, not plain locale prefixes.
+const legacyPaths = {
+  '/demos': '/ecosystem',
+  '/genesis-drop': '/genesis-drop',
+  '/get-started': '/get-started',
+  '/hemi-earn': '/hemi-earn',
+  '/stake': '/stake/dashboard',
+  '/staking-dashbord': '/staking-dashboard',
+  '/tunnel': '/tunnel',
+}
```
gndelia, line 39: No path has changed here. I don't think this is needed.

- `'/stake'` redirecting to `'/stake/dashboard'` is a redirect that can be solved client-side, just like redirecting `/` to `/tunnel`. 
- `/demos` to `/ecosystem` is a redirect that already existed on Next, so I don't think we need to track this list of "legacy" paths - technically, all of these existed on next and exist now on vite

portal/app/[locale]/hemi-earn/pool/[shareAddress]/page.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/pool/[shareAddress]/page.tsx]: Good catch, moved the check up into the layout in [`676c5371`](https://github.com/hemilabs/ui-monorepo/pull/2217/commits/676c5371). A bad share now never mounts the operations provider or the status updaters, and the page can just trust the address it gets.

portal/app.tsx
ArturDolzan: Rather than adding the two entries, I made the locale layout treat an unknown first segment as a path and redirect to `/{preferredLocale}{pathname}`, in [`1fc3c95c`](https://github.com/hemilabs/ui-monorepo/pull/2217/commits/1fc3c95c). That covers `/ecosystem`, `/staking-dashboard` and anything we add later without anyone having to remember the table. It also brings back the 404 for a locale we do not support, so `/fr/tunnel` no longer lands on the tunnel.

portal/app/[locale]/hemi-earn/_fetchers/fetchHemiEarnAssetConfigs.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_fetchers/fetchHemiEarnAssetConfigs.ts:26]: Yep, a regex ate the last line of the block. Restored in [`5bb65638`](https://github.com/hemilabs/ui-monorepo/pull/2217/commits/5bb65638).

portal/app.tsx
ArturDolzan: Nice, done in [`c6b66ba2`](https://github.com/hemilabs/ui-monorepo/pull/2217/commits/c6b66ba2). Every page and layout is a named export now.

portal/app.tsx
ArturDolzan: Fair point, the "legacy" framing was wrong, all of these existed on Next too. Dropped in [`1fc3c95c`](https://github.com/hemilabs/ui-monorepo/pull/2217/commits/1fc3c95c): the five entries that mapped a path to itself are gone (the locale layout now falls back for any unknown first segment), and what is left is the two that actually rewrite, `/stake` to `/stake/dashboard` and the misspelled `/staking-dashbord` folder. `/demos` to `/ecosystem` was never in the table, it is a plain route under `/:locale`.

gabmontes (approved):

PR #2225, opened 2026-08-27
Move Sentry off the Next SDK and restore error reporting

ArturDolzan:
### Description

This PR moves the portal off `@sentry/nextjs` onto `@sentry/react` plus `@sentry/vite-plugin`, and in doing so fixes error reporting, which has been completely dead since the Vite switch.

The dead part is worth explaining because it is invisible from both ends. `instrumentation-client.ts` registers `thirdPartyErrorFilterIntegration`, which matches each stack frame against bundle keys that the Sentry bundler plugin injects as module metadata. That plugin lived in `withSentryConfig` inside `next.config.ts`, which was deleted when the build moved to Vite. With no metadata every frame comes back as an empty key list, `drop-error-if-exclusively-contains-third-party-frames` matches every frame, and `processEvent` returns null. Sentry accepts the DSN, the build is green, and not a single event is ever sent. #2215 fixed the other half of this (the client was not being initialised at all) but only that half.

Two things needed care. The application key the plugin stamps only lives in the committed `.env`, no CI job passes it, and unlike Next, Vite does not load `.env` files into `process.env` inside the config. So the config uses `loadEnv`, and reads the DSN from there too so the plugin runs exactly when the client initialises. Separately, the filter integration now stays out of the array when the key is missing, since filtering against nothing is not filtering, it is discarding.

The SDK swap itself is the import line in four files, with one exception: `@sentry/nextjs` pushed `browserTracingIntegration` into its defaults and `@sentry/react` does not, so it is added explicitly. Without that, `tracesSampleRate` would have quietly become dead config.

Source maps behave like they did under Next: generated only when Sentry is on, `hidden` so browsers cannot discover them, and deleted once the plugin is done, so nothing ever ships the sources. `reactComponentAnnotation` is carried over from the old config too, since without it click breadcrumbs degrade from component names to minified DOM selectors.

Broken down into:
- `8493647` - Swap the Sentry SDK for its React build
- `8650387` - Put the Sentry bundler plugin back
- `6766c77` - Rename the instrumentation entry point

Verified against the built bundle, since every failure mode here is silent: with a DSN set, the application key is stamped into 194 chunks and `data-sentry-component` appears 349 times; with the key blanked, the filter integration is absent instead of dropping everything; with no DSN, none of it ships. No `.map` survives in any of the three. Also `pnpm lint`, `pnpm format:check`, `pnpm deps:check`, plus `pnpm typecheck`, `pnpm build` and `pnpm test` in the portal (74 files, 906 tests). Bundle size is unchanged at 2,932 kB.

## Notes

Three things I left alone. The five web workers get their own plugin pipeline via `worker.plugins`, so their 12 chunks carry no metadata; that is latent rather than live, since Sentry does not run inside the workers today. `build.sourcemap` keys off the DSN rather than off the upload credentials, so a build with a DSN and no auth token generates maps and deletes them unused. And the `rewriteFrames` iteratee calls `new URL(frame.filename!)` with no guard, so one filename-less frame silently disables rewriting for the whole event, but that code is identical on `main` and does not belong in this PR.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2194 

### Checklist

- [x] Manual testing passed.


portal/instrument.ts
```diff
@@ -81,17 +82,18 @@ function enableSentry() {
     ignoreErrors,
     // Integrations listed here are added alongside the default ones.
     integrations: [
-      // See https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/integrations/captureconsole/
+      // `@sentry/nextjs` added this to its defaults and `@sentry/react` does
```
gndelia: no need to mention `@sentry/nextjs` - this comment will make no sense once everything lands

ArturDolzan: Fair point, that only makes sense while the migration is in flight. Trimmed it down to why the line has to be there at all, without naming any package, in [`1ae8ade`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/1ae8ade3).

pnpm-workspace.yaml
```diff
@@ -19,7 +19,7 @@ minimumReleaseAgeExclude:
   - 'hemi-viem'
 
 onlyBuiltDependencies:
-  - '@sentry/cli' # downloads sentry-cli binary for sourcemap upload during `next build`
+  - '@sentry/cli' # downloads the sentry-cli binary @sentry/vite-plugin uses to upload source maps
   - 'esbuild' # platform binary used by vitest (via vite) for test transforms
   - 'sharp' # libvips bindings required by Next.js production image optimization
```
gndelia, line 24: FYI this is also next related - just so you keep an eye on it, it doesn't seem to be needed for this PR

portal/vite.config.ts
```diff
@@ -1,34 +1,75 @@
...
+        project: env.SENTRY_PROJECT,
+        reactComponentAnnotation: { enabled: true },
+        release:
+          env.SENTRY_ENVIRONMENT && env.VITE_SENTRY_RELEASE
+            ? {
+                deploy: { env: env.SENTRY_ENVIRONMENT },
+                name: env.VITE_SENTRY_RELEASE,
+              }
+            : undefined,
+        // Matches what `@sentry/nextjs` did: the maps exist only long enough to
```
gndelia: same comment about `@sentry/nextjs`

portal/vite.config.ts
```diff
@@ -1,34 +1,75 @@
...
+        // be uploaded, so the bundle never ships the sources. Note the plugin
+        // deletes them even when no upload happened, which is also what the
+        // Next build did.
+        sourcemaps: { filesToDeleteAfterUpload: ['./dist/**/*.map'] },
```
gndelia:
```suggestion
        sourcemaps: { filesToDeleteAfterUpload: ['dist/**/*.map'] },
```

It seems this will allow us to run the build from the root of the monorepo, making it more robust. Suggestion from IA - double-check btw

portal/README.md
```diff
@@ -71,9 +71,9 @@ VITE_PROXY_OVM_L1_CROSS_DOMAIN_MESSENGER=<address>
 VITE_PROXY_OVM_L1_STANDARD_BRIDGE=<address>
 # Use it to enable wallet connect
 VITE_WALLET_CONNECT_PROJECT_ID=<wallet-connect-id>
-# Error reporting with Sentry. Not wired under Vite yet: instrumentation-client.ts is a Next convention file that nothing imports
-VITE_SENTRY_DSN=<dsn> # Sentry DSN
-VITE_SENTRY_FILTER_KEY_ID=<string> # Application key used to tell first-party frames from third-party ones
+# Error reporting with Sentry
+VITE_SENTRY_DSN=<dsn> # Sentry DSN. Also what enables the build-time plugin
+VITE_SENTRY_FILTER_KEY_ID=<string> # Application key used to tell first-party frames from third-party ones. Must reach the build, or every event is dropped
```
gndelia: > Must reach the build, or every event is dropped

`VITE_SENTRY_FILTER_KEY_ID` seems to only disable `thirdPartyErrorFilterIntegration` integration, not all events

portal/instrument.ts
```diff
@@ -81,17 +82,17 @@ function enableSentry() {
     ignoreErrors,
     // Integrations listed here are added alongside the default ones.
     integrations: [
-      // See https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/integrations/captureconsole/
+      // Required for `tracesSampleRate` below to do anything.
+      Sentry.browserTracingIntegration(),
```
gndelia: IA recommended here using [reactRouterBrowserTracingIntegration](https://docs.sentry.io/platforms/javascript/guides/react/features/react-router/v8/#usage-with-createbrowserrouter-or-creatememoryrouter-data-mode)

it's odd because it's not easily findable in Sentry docs. Check if it works better or not

portal/vite.config.ts
```diff
@@ -1,34 +1,75 @@
...
+  const env = loadEnv(mode, process.cwd(), '')
+
+  const plugins: PluginOption[] = [react(), polyfills()]
+
+  if (env.VITE_SENTRY_DSN) {
+    plugins.push(
+      sentryVitePlugin({
+        applicationKey: env.VITE_SENTRY_FILTER_KEY_ID,
+        authToken: env.SENTRY_AUTH_TOKEN,
+        org: env.SENTRY_ORG,
```
gndelia, line 21: ok so no technically your PR, but it seems these envs were never documented. Maybe we can check if that's the case and create an issue to document them properly

pnpm-workspace.yaml
ArturDolzan: Yeah, that one is on my list too. It is stale in the same way (`next/image` is gone), I just did not want to widen this PR. Happy to fold it into whichever PR finally drops `next`.

portal/vite.config.ts
ArturDolzan: Same as the other one, you are right. Rewrote it to say what the option does instead of what it matches, in [`5cedab3`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/5cedab34).

portal/vite.config.ts
ArturDolzan: Double-checked and the two are identical: `globFiles` calls `glob(patterns, { absolute: true, nodir: true })` with no `cwd`, so both resolve against `process.cwd()` and the leading `./` is a no-op. I tested it with fixture files, 2 matches from the right directory and 0 from its parent, same for both patterns.

The concern behind it is real though, so I anchored the glob to the config file instead of the cwd in [`5cedab3`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/5cedab34). That makes it independent of where the build was started from, which is what the suggestion was going for.

portal/README.md
ArturDolzan: You are right, and my wording was left over from an earlier version. It used to drop everything, then I added the guard that keeps the integration out of the array when the key is missing, and I never went back to the README. Fixed in [`5cedab3`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/5cedab34).

portal/instrument.ts
ArturDolzan: Good call, and I was wrong to skip it. I had assumed the router-aware ones stopped at v7 while we are on 8.3.0, but `reactRouterV7BrowserTracingIntegration` is actually marked deprecated in favour of the unversioned `reactRouterBrowserTracingIntegration`, which says "Works with React Router v6+". Checked that react-router 8.3.0 exports all five things it needs.

Switched in [`367053c`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/367053cc). It is also the more faithful port, since `@sentry/nextjs` gave us router-aware tracing, and it means a share address no longer becomes its own transaction.

portal/vite.config.ts
ArturDolzan: Agreed, and it predates this PR: the same four were read by `next.config.ts` and were never documented either. Opened #2228 for it.

portal/instrument.ts
```diff
@@ -81,17 +89,25 @@ function enableSentry() {
     ignoreErrors,
     // Integrations listed here are added alongside the default ones.
     integrations: [
-      // See https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/integrations/captureconsole/
+      // Required for `tracesSampleRate` below to do anything. The router-aware
+      // one names transactions after the route pattern rather than the raw URL,
+      // so every share address does not become its own transaction.
+      Sentry.reactRouterBrowserTracingIntegration({
+        createRoutesFromChildren,
```
gabmontes, line 96: `reactRouterBrowserTracingIntegration` is configured here for route-pattern transaction naming, but `<Routes>` in `portal/app.tsx` doesn't seem to be wrapped with `Sentry.wrapReactRouterRouting`, which Sentry requires for that naming to take effect.

Without the wrap, navigations still get named after the raw pathname instead of the route pattern — e.g. `/hemi-earn/pool/0xABC...` instead of `/hemi-earn/pool/:shareAddress` — so each distinct address becomes its own transaction, which is the exact problem this integration is meant to solve. Doesn't fail the build or tests, so it'd pass silently.

Could you wrap the routes with `Sentry.wrapReactRouterRouting`?

— Gabi 🤖

ArturDolzan: Confirmed, and thanks for catching it. Went through the source: `wrapReactRouterRouting` is what calls `createRoutesFromChildren` and `addRoutesToAllRoutes`, and until that runs the span keeps `source: "url"`. The integration on its own never learns the tree, and it only warns about missing params, not about the missing wrap.

Wrapped `<Routes>` in `app.tsx` in [`6d38bb6`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/6d38bb69). My reply on the other thread claimed the pool addresses were already collapsed, which was not true yet.

portal/vite.config.ts
```diff
@@ -1,34 +1,80 @@
...
-  },
-  worker: {
-    // Worker bundles get their own plugin pipeline, the top-level `plugins`
-    // above do not apply to them.
-    format: 'es',
-    plugins: () => [polyfills()],
-  },
+export default defineConfig(function ({ mode }) {
+  // `process.env` never sees the `.env` files here, unlike under Next, and the
+  // filter key only lives there: no CI job passes it.
```
gndelia:
```suggestion
```

I would just drop it entirely - it also references Next again

portal/vite.config.ts
```diff
@@ -1,34 +1,80 @@
...
+      // Only Sentry consumes these, and the plugin above deletes them once it
+      // is done, so nothing generates them when it is not running. "hidden"
+      // also drops the sourceMappingURL, keeping them out of reach in the
+      // window between writing and deleting.
+      sourcemap: env.VITE_SENTRY_DSN ? 'hidden' : false,
+    },
+    // Not for our code: stream-http and readable-stream, which the http/https
+    // polyfills pull in, read the bare `global` at module scope. The plugin
+    // injects its shim into the main bundle but not into worker ones, so without
+    // this the workers throw ReferenceError as soon as they hit the network.
```
gndelia: Extra verbose, it could be something simpler like


```suggestion
    // Needed for stream-http and readable-stream, pulled in by http/https polyfills
```

portal/vite.config.ts
ArturDolzan: Fair, dropped it in [`82b759f`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/82b759ff). That is three Next references you have caught in this file, sorry about that.

portal/vite.config.ts
ArturDolzan: Trimmed it in [`82b759f`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/82b759ff). I kept the "main bundle but not worker ones" bit, since without it the `define` reads as redundant with the `nodePolyfills` already in `plugins` and someone will delete it.

Worth knowing this one is not from this PR, it came in with `ac8aba9` in #2202 and only shows up in the diff because the config changed shape.

gndelia (approved):

portal/vite.config.ts
```diff
@@ -1,34 +1,77 @@
...
-    // above do not apply to them.
-    format: 'es',
-    plugins: () => [polyfills()],
-  },
+export default defineConfig(function ({ mode }) {
+  const env = loadEnv(mode, process.cwd(), '')
+
+  const plugins: PluginOption[] = [react(), polyfills()]
+
+  if (env.VITE_SENTRY_DSN) {
```
gndelia: So IA suggested that here (In relation to #2227 )

```
const uploadToSentry = !!env.VITE_SENTRY_DSN && !process.env.STORYBOOK
```

Apparently, `STORYBOOK` is set when running storybook (See [here](https://github.com/storybookjs/storybook/blob/09b46444b893ac6147f14dca775e51a220f15e7b/code/core/src/bin/core.ts#L28) which I found on my own)

IF that's the case, maybe we can use it so we don't add Sentry on Storybook mode, and then on PR #2227 you can drop the `withoutVitePlugins` that removes `sentry-vite-plugin`  (As it won't be added here)

See if it makes sense

arielmergen (approved):
I’ve upgraded OpenSSL from 3.5.7-r0 to 3.5.8-r0 to resolve the Docker check. If you need to fix it in order to continue, you’ll need to upgrade the version as well. Otherwise, if it doesn’t block your work, 

LGTM 🚀

gabmontes (approved):
LGTM but let's resolve the comment about `STORYBOOK` before merging.

portal/vite.config.ts
ArturDolzan: Nice, that works. Verified it in three places: `process.env.STORYBOOK` comes back `"true"` in both `storybook:build` and `storybook dev`, and `undefined` in the app build. The CLI sets it at the top of `dist/bin/core.js`, well before the config is loaded.

Gated in [`cea7437`](https://github.com/hemilabs/ui-monorepo/pull/2225/commits/cea74379). I put it on `build.sourcemap` too, otherwise a Storybook build with a DSN would still generate maps with nothing left to delete them.

Doing it here rather than in the Storybook config is also better than what I had: it means the next thing that reads this config does not need to know Sentry's plugin names to opt out. Will drop the `withoutVitePlugins` in #2227 once this lands.

gndelia (approved):

portal/vite.config.ts
ArturDolzan: Done in [`27ee012`](https://github.com/hemilabs/ui-monorepo/pull/2227/commits/27ee0127) on #2227. Merged this branch in first so I could check it for real: `storybook:build` with a DSN set now has no Sentry plugin warnings, no application key in the bundle and no `.map` files, with nothing filtering it.

Also dropped `@storybook/builder-vite`, which was only a direct dependency for the helper that did the filtering.

PR #2227, opened 2026-08-27
Build Storybook with Vite instead of webpack

ArturDolzan:
### Description

This PR swaps `@storybook/nextjs` for `@storybook/react-vite`, which was the last thing in the repo still running webpack. Until now the same components were built twice by two different toolchains: Babel and webpack for Storybook, Vite for the app.

The mechanical part is smaller than the file count suggests. All 16 stories import `Meta` and `StoryObj` as **types only**, none of them touch `next/*`, i18n or react-router, and the components they render are plain UI. So it is the module name and nothing else. The interesting number is the lockfile: **314 packages out, 11 in**, all of the removed ones Babel or webpack. Build time for `storybook:build` went from a full webpack pass to about 1.3s.

Two things needed handling because the Vite builder reads the app's own `vite.config.ts`.

First, that config now carries the Sentry plugin (from the stacked Sentry PR). A Storybook build is not a release, and leaving the plugin in would make it upload this bundle as the app's own and then delete the app's real sourcemaps, since its cleanup glob points at `./dist`. Vetro solves the same class of problem by filtering plugins in `viteFinal`, so this does that, using `withoutVitePlugins` from `@storybook/builder-vite` rather than hand rolling it (it recurses into nested arrays and resolves async plugins, which a plain filter does not).

Second, Vite's `publicDir` copies all of `public` into the output. The comment on `staticDirs` explicitly promises the opposite, that only the fonts ship and not the video or the favicon, and that promise was true under webpack. It is restored with `publicDir: false`. Worth knowing: this changes the emitted font URLs from relative to absolute, which is fine because Storybook is always served from a root (we do not deploy it at all today, and vetro serves theirs at the root of its own worker).

Verified with `pnpm lint`, `pnpm format:check`, `pnpm deps:check`, plus `pnpm typecheck`, `pnpm build` and `pnpm test` in the portal (74 files, 906 tests). For the Storybook build specifically: zero webpack lines in the output, no video in `storybook-static`, fonts present, and with a DSN set there are no Sentry plugin warnings and no `.map` files. I checked the plugin filter is actually doing something by pointing it at a bogus name, which brings the warnings back.

## Notes

`next` stays in `package.json`. It is no longer imported anywhere in the portal, but `umami-analytics-next` (and `nuqs`) declare it as a peer and the former imports `next/navigation` and `next/script` at module scope, so removing it would just make the dependency implicit and rely on peer auto install. It leaves for real with the Umami item.

Also unrelated but noticed while in here: the `WithGoTo` story in `toast.stories.tsx` renders an internal `Link`, which needs router, intl and nuqs providers that `preview.ts` does not supply. That predates this PR, so I left it alone.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2194 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/.storybook/main.ts
+++ b/portal/.storybook/main.ts
@@ -1,9 +1,10 @@
-import type { StorybookConfig } from '@storybook/nextjs'
+import { withoutVitePlugins } from '@storybook/builder-vite'
+import type { StorybookConfig } from '@storybook/react-vite'
 
 const config: StorybookConfig = {
   addons: ['storybook-addon-pseudo-states'],
   framework: {
-    name: '@storybook/nextjs',
+    name: '@storybook/react-vite',
     options: {},
   },
   // globals.css is the only font source and points at /fonts, which Storybook
@@ -11,6 +12,18 @@ const config: StorybookConfig = {
   // (the video, the favicon) gets published in the Storybook build.
   staticDirs: [{ from: '../public/fonts', to: '/fonts' }],
   stories: ['../stories/**/*.stories.@(ts|tsx)'],
+  viteFinal: async viteConfig => ({
+    ...viteConfig,
+    // Storybook is not a release: with a DSN set, the plugin would upload
+    // this bundle as the app's own, and delete the app's sourcemaps on its
+    // way out.
+    plugins: await withoutVitePlugins(viteConfig.plugins, [
+      'sentry-vite-plugin',
+    ]),
+    // Vite would otherwise copy the whole of `public`, which is what makes
+    // the promise above about `staticDirs` true rather than aspirational.
+    publicDir: false,
+  }),
 }
 
 export default config
--- a/portal/.storybook/main.ts
+++ b/portal/.storybook/main.ts
@@ -1,4 +1,3 @@
-import { withoutVitePlugins } from '@storybook/builder-vite'
 import type { StorybookConfig } from '@storybook/react-vite'
 
 const config: StorybookConfig = {
@@ -12,14 +11,8 @@ const config: StorybookConfig = {
   // (the video, the favicon) gets published in the Storybook build.
   staticDirs: [{ from: '../public/fonts', to: '/fonts' }],
   stories: ['../stories/**/*.stories.@(ts|tsx)'],
-  viteFinal: async viteConfig => ({
+  viteFinal: viteConfig => ({
     ...viteConfig,
-    // Storybook is not a release: with a DSN set, the plugin would upload
-    // this bundle as the app's own, and delete the app's sourcemaps on its
-    // way out.
-    plugins: await withoutVitePlugins(viteConfig.plugins, [
-      'sentry-vite-plugin',
-    ]),
     // Vite would otherwise copy the whole of `public`, which is what makes
     // the promise above about `staticDirs` true rather than aspirational.
     publicDir: false,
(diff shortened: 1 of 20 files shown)


gabmontes (commented):
The Docker checks that are failing were fixed in #2221. You may want to rebase this PR to include those.

ArturDolzan:
> [@gabmontes commented](https://github.com/hemilabs/ui-monorepo/pull/2227#pullrequestreview-5046036247):
> The Docker checks that are failing were fixed in #2221. You may want to rebase this PR to include those.

Thanks for flagging. These are stacked PRs (#2222 into #2225 into this one) and the failures are all in the portal-backend Dockerfiles, so it is cosmetic as far as these changes go. Instead of rebasing every branch in the chain, we will let the stack land on `vite-migration` first and then merge `main` into it, which picks up `7fe52c0` in one shot. Simpler and safer than touching all of them.

portal/.storybook/main.ts
```diff
@@ -1,16 +1,29 @@
...
   staticDirs: [{ from: '../public/fonts', to: '/fonts' }],
   stories: ['../stories/**/*.stories.@(ts|tsx)'],
+  viteFinal: async viteConfig => ({
+    ...viteConfig,
+    // Storybook is not a release: with a DSN set, the plugin would upload
+    // this bundle as the app's own, and delete the app's sourcemaps on its
+    // way out.
+    plugins: await withoutVitePlugins(viteConfig.plugins, [
+      'sentry-vite-plugin',
+    ]),
```
gndelia: FYI let a suggestion to drop this on https://github.com/hemilabs/ui-monorepo/pull/2225/changes#r3881213742

gndelia (commented):
Not a problem of this PR, but it seems stories are excluded from Tailwind 

https://github.com/hemilabs/ui-monorepo/blob/02e06834846195a44d24b6d0438cacf1a2b77251/portal/tailwind.config.ts#L4

We may want to include them in a separated PR (cc @arielmergen  since you implemented this)

gabmontes (approved):

portal/.storybook/main.ts
ArturDolzan: Dropped in [`27ee012`](https://github.com/hemilabs/ui-monorepo/pull/2227/commits/27ee0127), thanks. The gate lives in `vite.config.ts` now, so this config does not need to know about Sentry at all.

portal/.storybook/main.ts
```diff
@@ -1,16 +1,22 @@
...
   // globals.css is the only font source and points at /fonts, which Storybook
   // does not serve on its own. Scoped to `fonts` so nothing else under `public`
   // (the video, the favicon) gets published in the Storybook build.
   staticDirs: [{ from: '../public/fonts', to: '/fonts' }],
   stories: ['../stories/**/*.stories.@(ts|tsx)'],
+  viteFinal: viteConfig => ({
+    ...viteConfig,
+    // Vite would otherwise copy the whole of `public`, which is what makes
+    // the promise above about `staticDirs` true rather than aspirational.
+    publicDir: false,
```
gndelia, line 18: doesn't this leave the fonts off?

ArturDolzan: No, they come from `staticDirs`, which is Storybook's own mechanism and does not go through Vite's `publicDir`. Checked the build output: the four woff2 files are in `storybook-static/fonts/` and every `url()` the CSS emits resolves on disk.

I ran it without `publicDir: false` to be sure of what it actually changes, and the only difference is `navbarVideo.mp4` and the app favicon showing up at the root. Fonts are 4 either way. That is what the comment on `staticDirs` promises and what webpack used to do, so this just restores it.

gndelia (approved):

PR #2229, opened 2026-08-27
Update Toast to match new design

### LINKED ISSUE #2062: Toast: "none" variant looks identical to "success", and "error" icon changes toast size
### Description

Two visual issues surfaced while reviewing the Toast Storybook story (PR #2058), on the `Toast` component itself (`portal/components/toast/index.tsx`):

1. **`variant="none"` is indistinguishable from `variant="success"`.** When `variant` is `undefined`, the component renders `<CheckMark className="[&>path]:fill-emerald-500" />` — a green 16×16 check. `variant="success"` renders `GreenCheckIcon`, also a green 16×16 check. The two are visually near-identical, so the "no variant" state reads as a success toast.

   **Variant: None**
   Variant none

   **Variant: Success**
   Variant success

2. **The `error` icon is larger and changes the toast's size.** The `error` icon is wrapped in a `div` with `size-5` (20px) plus a `rose-950` background, while `success`/`none` use the bare 16px SVG. Given the same title/description, the error toast ends up a different height than the others.

   https://github.com/user-attachments/assets/a1ef802a-3e42-47d1-b765-e38043035535

*(Originally reported by @gndelia in PR #2058.)*

### Steps to reproduce

1. Open the `Components/Toast` story in Storybook.
2. Toggle the `variant` control between `none` and `success` → the icons look the same.
3. Set `variant` to `error` with the same title/description as `success` → the toast renders at a different size.

### Proposed fix

- Give `variant="none"` a distinct icon (or drop the green fill) so it doesn't read as success.
- Normalize the icon box size across variants so the toast height stays consistent (e.g. give `success`/`none` the same 20px container as `error`, or vice versa).

### LINKED ISSUE #2087: Update Toast component to match new design

Update the base `Toast` component (`portal/components/toast/index.tsx`) to match the new design provided by design in Figma:

- **Figma:** https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=20261-30873

The component should support the variants defined in the design (`success` / `error`, with an optional CTA and a hover state that reveals the close action).

The current toast does not match the design system. The redesign aligns the toast with Figma across background, radius, typography, icons and states, and adds the CTA and hover-to-close behavior the design specifies.

### Proposed implementation

Adjust the existing `Toast` to the Figma spec (keep the current `body` portal and auto-close behavior):

**Container**
- Background `neutral-950` (`#0a0a0a`) — currently `neutral-800`.
- `rounded-lg` (8px) — currently `rounded-xl`.
- Padding `p-3` (12px); width `384px` on desktop.
- Drop shadow `sm` (per Figma effect); drop the current border.

**Title row**
- Icon (16px) + title, `13px` medium, white.
- `success` → orange check icon; `error` → pink/magenta info icon (currently emerald check / rose error — colors change per Figma).
- On **hover**, show a close (`X`) button on the right.

**Description row**
- Aligned under the title text (`pl-6` / 24px), `13px` regular, `neutral-400` (`#a3a3a3`).

**Optional CTA**
- White button ("Call to action"), `rounded-md` (6px), `12px` semibold, with the design's shadow. Aligned under the title (`pl-6`).

**Notes**
- Exact icon assets/colors and spacing to be taken from the Figma node above.
- Keep all user-facing strings translated (`next-intl`), as today.
- After the component is updated, the existing Storybook Toast story should reflect the new design.

arielmergen:
### Description

This commit update the Toast component to match new design
### Screenshots

05-portal-desktop

06-portal-mobile

01-success-default

02-success-hover-close

03-error

04-cta

### Related issue(s)

Closes #2087
Closes #2062
Fixes #
Related to #

### Checklist

- [X] Manual testing passed.
- [X] Documentation updated, or N/A.
- [X] Environment variables set in CI, or N/A.


gabmontes (dismissed):

portal/components/toast/index.tsx
```diff
@@ -70,20 +65,21 @@ export const Toast = function ({
   // ancestor in the app layout would create a subordinate stacking context
   // and trap the toast behind any open drawer regardless of z-index.
   return ReactDOM.createPortal(
-    <div className="fixed inset-x-4 bottom-20 z-40 flex items-center gap-3 rounded-xl border border-black/88 bg-neutral-800 p-3.5 pb-4 text-white shadow-soft md:bottom-auto md:left-auto md:right-8 md:top-20">
-      <div className="shrink-0">
-        {variant !== undefined ? (
+    <div className="group fixed inset-x-4 bottom-20 z-40 flex flex-col gap-y-3 overflow-hidden rounded-lg bg-neutral-950 p-3 text-white shadow-sm md:bottom-auto md:left-auto md:right-8 md:top-20 md:w-96">
+      <div className="flex flex-col gap-y-0.5">
+        <div className="flex h-4.5 items-center gap-x-2">
```
ArturDolzan: Perhaps it's worth checking how a longer title behaves here. With `h-4.5` plus `overflow-hidden` on the root, the pt/es "Não foi possível adicionar o token à sua carteira" string looks like it wraps and gets clipped on mobile. Would `min-h-4.5` be safer?

portal/components/toast/index.tsx
```diff
@@ -70,20 +65,21 @@ export const Toast = function ({
...
-        )}
-      </div>
-      <div className="flex flex-[1_0_0] flex-col items-start">
-        <p className="text-base font-medium leading-4 tracking-[-0.32px] text-white">
-          {title}
-        </p>
+          <p className="text-sm font-medium text-white">{title}</p>
+          <button
+            aria-label="Close notification"
+            className="ml-auto size-4 shrink-0 transition-opacity duration-200 focus-visible:opacity-100 md:opacity-0 md:group-hover:opacity-100"
```
ArturDolzan: Might be worth a quick look at the hover-reveal on tablets. Since `:hover` doesn't really fire on touch, the X seems to stay at `opacity-0` on iPad and users would have to wait for the auto-close. Wonder if scoping it to `@media (hover: hover)` would help.

portal/components/toast/index.tsx
```diff
@@ -70,20 +65,21 @@ export const Toast = function ({
...
-          <CheckMark className="[&>path]:fill-emerald-500" />
-        )}
-      </div>
-      <div className="flex flex-[1_0_0] flex-col items-start">
-        <p className="text-base font-medium leading-4 tracking-[-0.32px] text-white">
-          {title}
-        </p>
+          <p className="text-sm font-medium text-white">{title}</p>
+          <button
+            aria-label="Close notification"
```
gndelia: This needs to be translated

portal/components/toast/toastLoader.tsx
```diff
@@ -2,7 +2,7 @@ import Skeleton from 'react-loading-skeleton'
 
 export const ToastLoader = () => (
   <Skeleton
-    className="h-26 w-full rounded-xl md:w-96"
+    className="h-16 w-full rounded-lg md:w-96"
```
gndelia: I think the height of the toast loader should be defined when using it. This is because the goal is to show a skeleton while the real component loads - so they should match the size. Because of this, the height of the Toast this loader is replacing will depend on the toast having or not a subtitle, CTA, etc.

portal/components/toast/index.tsx
arielmergen: Reproduced at 358px (iPhone width minus `inset-x-4`): the title wraps to two lines inside the fixed 18px row and, with `items-center`, overflows both ways — the second line overlaps the description rather than just being clipped.

Changed `h-4.5` to `min-h-4.5`. The row grows, the toast goes from 62px to 80px and nothing is cut. Single-line toasts are unchanged at 62px / 98px.

portal/components/toast/index.tsx
arielmergen: The breakpoint was the wrong criterion: an iPad matches `md:`, so the X stays at `opacity-0` with no way to dismiss the toast before the auto-close.

Now `[@media(hover:hover)]:opacity-0 [@media(hover:hover)]:group-hover:opacity-100` — an arbitrary variant, since 3.4 has no `pointer-*`. Both rules compile inside `@media (hover: hover)`, so where it does not match the X stays visible. One side effect worth noting: a narrow desktop window now hides it until hover, which was not the case before.

portal/components/toast/index.tsx
arielmergen: Done: `aria-label={t('close-notification')}`.

I added a new `common.close-notification` key in en/es/pt instead of reusing `common.close`, which is used as visible button text — reusing it would announce just "Close" and drop what is being closed. The stories now get `NextIntlClientProvider` from `meta.decorators`, so all three render.

portal/components/toast/toastLoader.tsx
arielmergen: `ToastLoader` now takes a required `className` and each call site passes its own height.

Measured heights are 62px without CTA and 98px with. `h-24.5` is exactly 98px, so `stakeOperation` uses that; the other two keep `h-16`.

portal/components/toast/index.tsx
```diff
@@ -70,20 +67,21 @@ export const Toast = function ({
...
-        <p className="text-base font-medium leading-4 tracking-[-0.32px] text-white">
-          {title}
-        </p>
+          <p className="text-sm font-medium text-white">{title}</p>
+          <button
+            aria-label={t('close-notification')}
+            className="ml-auto size-4 shrink-0 transition-opacity duration-200 focus-visible:opacity-100 [@media(hover:hover)]:opacity-0 [@media(hover:hover)]:group-hover:opacity-100"
+            onClick={() => setClosedToast(true)}
+          >
+            <CloseIcon className="size-full [&>path]:fill-neutral-400 [&>path]:hover:fill-white" />
```
ArturDolzan, line 80: Perhaps it's worth checking the stacked-modifier order here. `[&>path]:hover:` compiles to `.cls > path:hover`, so the white only kicks in over the thin X strokes themselves. Wonder if `hover:[&>path]:fill-white` (`.cls:hover > path`) would feel better.

portal/components/toast/index.tsx
```diff
@@ -70,20 +67,21 @@ export const Toast = function ({
...
           <ToastIcon variant={variant} />
-        ) : (
-          <CheckMark className="[&>path]:fill-emerald-500" />
-        )}
-      </div>
-      <div className="flex flex-[1_0_0] flex-col items-start">
-        <p className="text-base font-medium leading-4 tracking-[-0.32px] text-white">
-          {title}
-        </p>
+          <p className="text-sm font-medium text-white">{title}</p>
```
ArturDolzan, line 74: Not sure it matters much, but the root already sets `text-white`, so the title `<p>` might be able to drop it and just inherit.

portal/components/toast/toastLoader.tsx
```diff
@@ -1,8 +1,8 @@
 import Skeleton from 'react-loading-skeleton'
 
-export const ToastLoader = () => (
+export const ToastLoader = ({ className }: { className: string }) => (
   <Skeleton
-    className="h-26 w-full rounded-xl md:w-96"
+    className={`w-full rounded-lg md:w-96 ${className}`}
```
gndelia: I am not really a fan of `className` - because it basically allows for _any_ CSS class. If the different heights are from a limited set (and I think they are) - maybe you can use variants to provide height. You can see throughout the app how we've applied this pattern

portal/components/toast/index.tsx
arielmergen: I checked the suggestion, and it resolves the other way around: [&>path]:hover: → .svg:hover > path (the whole 16×16 area, since the svg is size-full), hover:[&>path]: → .svg > path:hover (painted pixels only). I tried it on the rendered button: with the suggested order only 40 of the 256 pixels trigger the white, with gaps between the arms of the X, while the click target stays the full button. Leaving as is.

portal/components/toast/index.tsx
arielmergen: I checked it, and it is not redundant: globals.css:63 has p { @apply text-sm font-normal text-neutral-950 }, and an element selector beats inheritance. Removing it renders the title at rgb(10, 10, 10) on the dark background.

portal/components/toast/toastLoader.tsx
arielmergen: Dropped the className prop and ToastLoader altogether: the skeleton now lives inline in each dynamic() loading callback with its own height, like transaction-history/page.tsx or the other dynamic() in unstakeOperation itself. h-24.5 is 98px, the measured height of the toast with a CTA; the other two keep 64px.

portal/app/[locale]/stake/_components/manageStake/stakeOperation.tsx
```diff
@@ -36,7 +36,12 @@ import { useStake } from '../../_hooks/useStake'
 const StakeToast = dynamic(
   () => import('../stakeToast').then(mod => mod.StakeToast),
   {
-    loading: () => <ToastLoader />,
+    loading: () => (
+      <Skeleton
+        className="h-24.5 w-full rounded-lg md:w-96"
+        containerClassName="fixed bottom-20 inset-x-4 z-40 md:bottom-auto md:left-auto md:right-8 md:top-20"
+      />
+    ),
```
Comment (gabmontes) [inline portal/app/[locale]/stake/_components/manageStake/stakeOperation.tsx]: Replacing `ToastLoader` with a styled `Skeleton` in each component may not be optimal either. Now, to restyle the toast, we need to modify the Toast component and all the instances where the skeletons are loaded. What do you think @gndelia?

— Real Gabi

Comment (arielmergen) [reply portal/app/[locale]/stake/_components/manageStake/stakeOperation.tsx]: @gabmontes ToastLoader is back, now using a variant instead of a className: default (62px) and withCallToAction (98px). Positioning has been moved to a shared .toast-position class, since it was duplicated in the Toast root.

I followed @gndelia’s suggestion from a few comments above.

1071f551 — Extract the toast position into a class
001d2230 — Restore ToastLoader with a height variant

Let me know what you think.

portal/components/toast/toastLoader.tsx
```diff
@@ -1,8 +1,17 @@
 import Skeleton from 'react-loading-skeleton'
 
-export const ToastLoader = () => (
+const heights = {
+  default: 'h-16',
```
gabmontes: With this project's `text-sm` override (13px/18px line-height), the no-CTA toast (`p-3` + 18px title row + `gap-y-0.5` + 18px description row) renders at exactly 62px, but the `default` variant's skeleton is `h-16` (64px) — a 2px mismatch. `withCallToAction` doesn't have this problem: it computes to exactly 98px, matching `h-24.5`.

arielmergen: Confirmed, the Figma node for that variant is 384×62, so the skeleton was the one off, not the toast.

Fixed with h-[62px]. 62px isn't on the spacing scale, and it's a single-use value, so I kept it local to the component rather than adding a key to tailwind.config.ts.

Added in [15a6ef1](https://github.com/hemilabs/ui-monorepo/pull/2229/commits/15a6ef1956491d34d985d20afd97130672438180)

gndelia (approved):

ArturDolzan (approved):

gabmontes (approved):

PR #2230, opened 2026-08-31
Replace the deleted Next sitemap with a Vite plugin

### LINKED ISSUE #2228: Document the build-time SENTRY_* environment variables

`portal/README.md` lists every environment variable the portal reads, but only the `VITE_`-prefixed ones. Four variables are read at build time by `portal/vite.config.ts` and are not documented anywhere:

- `SENTRY_AUTH_TOKEN`
- `SENTRY_ENVIRONMENT`
- `SENTRY_ORG`
- `SENTRY_PROJECT`

They carry no `VITE_` prefix on purpose, since they must never reach the browser, which is also why they are easy to miss when reading the list.

Raised by @gndelia in [#2225](https://github.com/hemilabs/ui-monorepo/pull/2225#discussion_r3875238213). The gap predates that PR: the same four were read by `next.config.ts` before the Vite migration and were never listed either, so this is not a regression, just something that became visible while moving them.

It matters now because the Cloudflare migration ([#2194](https://github.com/hemilabs/ui-monorepo/issues/2194)) configures variables per environment by hand. Anyone setting that up reads `portal/README.md` to know what to enter, and these four are missing from it, so source map upload would silently not happen with a green build.

Today the only place they appear is `.github/actions/deploy-portal/action.yml`, as inputs with a one-line description, and that file is scheduled to be retired by the migration.

Worth checking whether other workspaces have the same gap before closing this, rather than only fixing the portal.

### Proposed implementation

Add them to the list in `portal/README.md`, in their own block, making clear they are read at build time and never shipped to the browser. `portal/Environment Variables.md` already points at `portal/README.md` as the place documentation lives, so no new convention is needed.

ArturDolzan:
### Description

This PR closes the last open item of phase 2: the portal has not emitted a `sitemap.xml` since `app/sitemap.ts` was deleted along with the Next app router.

It is not a restoration though. **The sitemap in production right now is broken in two independent ways**, and both predate the migration:

<loc>http://localhost:3000/en/hemi-earn/pool/[shareAddress]</loc>

All 30 URLs point at `http://localhost:3000`, because the generator read `process.env.HEMI_DOMAIN` and that variable was never an input to `deploy-portal` (it only exists in `hostinger-deployment.yml` to build the SSH path), so the `'http://localhost:3000'` fallback always won. And three of them carry a literal `[shareAddress]`, because the generator walked directories and did not skip dynamic segments.

So the goal here was to ship a correct one for the first time rather than reproduce the old behaviour. The new one is gated on `PORTAL_SITE_URL` **with no fallback**, which is what makes the first bug unrepeatable: there is no host to invent. The route list is explicit, so the dynamic pool route and the two redirect-only paths are out by construction, and `/hemi-earn` follows its feature flag since the layout answers 404 when it is off.

What replaces the directory walk is a test rather than the generator: for every `page.tsx` under `app/[locale]/`, either the route is listed or it is in a named exclusion list. That catches drift in both directions, and the exclusions read as `'/demos': 'redirects to /ecosystem'` instead of the old `!baseRoute.endsWith('demos')`, which is exactly the pattern that let `[shareAddress]` through.

Also dropped `lastModified`. It was a fresh `new Date()` on every build, telling crawlers every page changed on every deploy.

Broken down into:
- `9888546` - Add a sitemap generator
- `1697d7d` - Emit the sitemap at build time
- `327b57a` - Document the build-time variables

Verified against the build rather than by reading: with the variable set the output has 27 `<url>` entries, zero `localhost`, zero dynamic segments and valid XML per `xmllint`. Without it no file is emitted at all. With the earn flag off it drops to 24 and mentions no `hemi-earn`. I also broke the anti-drift test on purpose in both directions to confirm it fails when it should. Plus `pnpm lint`, `pnpm format:check`, `pnpm deps:check`, `pnpm typecheck`, `pnpm build` and `pnpm test` (76 files, 917 tests).

#### Notes

The third commit documents `PORTAL_SITE_URL` and, since the block for build-time variables had to be created anyway, the four `SENTRY_*` that were never listed either. That closes #2228.

Worth knowing for whoever picks up phase 5: `PORTAL_SITE_URL` has to be configured per environment on Cloudflare, or no sitemap ships. That is deliberate, but it does mean the file is absent until then.

### Screenshots

Captura de Tela 2026-08-31 às 10 37 01

### Related issue(s)

Related to #2194 

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/README.md
+++ b/portal/README.md
@@ -71,17 +71,17 @@ VITE_PROXY_OVM_L1_CROSS_DOMAIN_MESSENGER=<address>
 VITE_PROXY_OVM_L1_STANDARD_BRIDGE=<address>
 # Use it to enable wallet connect
 VITE_WALLET_CONNECT_PROJECT_ID=<wallet-connect-id>
-# Error reporting with Sentry
-VITE_SENTRY_DSN=<dsn> # Sentry DSN. Also what enables the build-time plugin
-VITE_SENTRY_FILTER_KEY_ID=<string> # Application key used to tell first-party frames from third-party ones. The filtering is skipped when it does not reach the build
-VITE_SENTRY_RELEASE=<string> # Release name, in the "portal@yyyymmdd_sequence" format. Envelopes not matching it are rewritten
-VITE_TRACES_SAMPLE_RATE=<number> # Ratio of transactions sampled for tracing. Ignored when not a number
-# Read at build time by vite.config.ts, never shipped to the browser, hence no VITE_ prefix
-PORTAL_SITE_URL=<url> # Base URL the sitemap is built from. Without it no sitemap.xml is emitted
+# Sitemap
+PORTAL_SITE_URL=<url> # Base URL the sitemap is built from. Read at build time. Without it no sitemap.xml is emitted
+# Error reporting with Sentry. The unprefixed ones are read at build time and never reach the browser
 SENTRY_AUTH_TOKEN=<token> # Authorizes the source map upload
 SENTRY_ENVIRONMENT=<string> # Environment the release is deployed to. Together with VITE_SENTRY_RELEASE it is what names the release
 SENTRY_ORG=<string> # Sentry organization slug
 SENTRY_PROJECT=<string> # Sentry project slug
+VITE_SENTRY_DSN=<dsn> # Sentry DSN. Also what enables the build-time plugin
+VITE_SENTRY_FILTER_KEY_ID=<string> # Application key used to tell first-party frames from third-party ones. The filtering is skipped when it does not reach the build
+VITE_SENTRY_RELEASE=<string> # Release name, in the "portal@yyyymmdd_sequence" format. Envelopes not matching it are rewritten
+VITE_TRACES_SAMPLE_RATE=<number> # Ratio of transactions sampled for tracing. Ignored when not a number
 ```
 
 If not defined, the contracts addresses used will be the ones defined in [hemi-viem](https://github.com/hemilabs/hemi-viem).
--- a/portal/README.md
+++ b/portal/README.md
@@ -76,6 +76,12 @@ VITE_SENTRY_DSN=<dsn> # Sentry DSN. Also what enables the build-time plugin
 VITE_SENTRY_FILTER_KEY_ID=<string> # Application key used to tell first-party frames from third-party ones. The filtering is skipped when it does not reach the build
 VITE_SENTRY_RELEASE=<string> # Release name, in the "portal@yyyymmdd_sequence" format. Envelopes not matching it are rewritten
 VITE_TRACES_SAMPLE_RATE=<number> # Ratio of transactions sampled for tracing. Ignored when not a number
+# Read at build time by vite.config.ts, never shipped to the browser, hence no VITE_ prefix
+PORTAL_SITE_URL=<url> # Base URL the sitemap is built from. Without it no sitemap.xml is emitted
+SENTRY_AUTH_TOKEN=<token> # Authorizes the source map upload
+SENTRY_ENVIRONMENT=<string> # Environment the release is deployed to. Together with VITE_SENTRY_RELEASE it is what names the release
+SENTRY_ORG=<string> # Sentry organization slug
+SENTRY_PROJECT=<string> # Sentry project slug
 ```
 
 If not defined, the contracts addresses used will be the ones defined in [hemi-viem](https://github.com/hemilabs/hemi-viem).
--- a/portal/test/utils/sitemapRoutes.test.ts
+++ b/portal/test/utils/sitemapRoutes.test.ts
@@ -6,7 +6,6 @@ import { describe, expect, it } from 'vitest'
 // Named on purpose: the previous generator walked the tree and excluded things
 // with `!baseRoute.endsWith('demos')`, which is how `[shareAddress]` slipped in.
 const excluded: Record<string, string> = {
-  '/demos': 'redirects to /ecosystem',
   '/hemi-earn/pool/[shareAddress]': 'dynamic, one URL per pool address',
 }
 
--- a/portal/test/utils/sitemapRoutes.test.ts
+++ b/portal/test/utils/sitemapRoutes.test.ts
@@ -1,0 +1,43 @@
+import fs from 'node:fs'
+import path from 'node:path'
+import { sitemapRoutes } from 'utils/sitemapRoutes'
+import { describe, expect, it } from 'vitest'
+
+// Named on purpose: the previous generator walked the tree and excluded things
+// with `!baseRoute.endsWith('demos')`, which is how `[shareAddress]` slipped in.
+const excluded: Record<string, string> = {
+  '/demos': 'redirects to /ecosystem',
+  '/hemi-earn/pool/[shareAddress]': 'dynamic, one URL per pool address',
+}
+
+const localeDir = path.join(import.meta.dirname, '../../app/[locale]')
+
+const pageRoutes = function (dir: string, route = ''): string[] {
+  const entries = fs.readdirSync(dir, { withFileTypes: true })
+  const own = entries.some(entry => entry.isFile() && entry.name === 'page.tsx')
+  return [
+    ...(own && route ? [route] : []),
+    ...entries
+      .filter(entry => entry.isDirectory() && !entry.name.startsWith('_'))
+      .flatMap(entry =>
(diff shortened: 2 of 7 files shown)


portal/README.md
```diff
@@ -76,6 +76,12 @@ VITE_SENTRY_DSN=<dsn> # Sentry DSN. Also what enables the build-time plugin
 VITE_SENTRY_FILTER_KEY_ID=<string> # Application key used to tell first-party frames from third-party ones. The filtering is skipped when it does not reach the build
 VITE_SENTRY_RELEASE=<string> # Release name, in the "portal@yyyymmdd_sequence" format. Envelopes not matching it are rewritten
 VITE_TRACES_SAMPLE_RATE=<number> # Ratio of transactions sampled for tracing. Ignored when not a number
+# Read at build time by vite.config.ts, never shipped to the browser, hence no VITE_ prefix
+PORTAL_SITE_URL=<url> # Base URL the sitemap is built from. Without it no sitemap.xml is emitted
+SENTRY_AUTH_TOKEN=<token> # Authorizes the source map upload
+SENTRY_ENVIRONMENT=<string> # Environment the release is deployed to. Together with VITE_SENTRY_RELEASE it is what names the release
+SENTRY_ORG=<string> # Sentry organization slug
+SENTRY_PROJECT=<string> # Sentry project slug
```
gndelia: Let's sort these env vars alphabetically with the others in the list

portal/README.md
```diff
@@ -76,6 +76,12 @@ VITE_SENTRY_DSN=<dsn> # Sentry DSN. Also what enables the build-time plugin
 VITE_SENTRY_FILTER_KEY_ID=<string> # Application key used to tell first-party frames from third-party ones. The filtering is skipped when it does not reach the build
 VITE_SENTRY_RELEASE=<string> # Release name, in the "portal@yyyymmdd_sequence" format. Envelopes not matching it are rewritten
 VITE_TRACES_SAMPLE_RATE=<number> # Ratio of transactions sampled for tracing. Ignored when not a number
+# Read at build time by vite.config.ts, never shipped to the browser, hence no VITE_ prefix
```
gndelia:
```suggestion
```

Not needed

portal/utils/sitemap.ts
```diff
@@ -0,0 +1,51 @@
...
+  route: string
+}) => escapeXml(`${baseUrl.replace(/\/+$/, '')}/${locale}${route}`)
+
+export const buildSitemap = function ({ baseUrl, locales, routes }: Params) {
+  const entries = routes.flatMap(route =>
+    locales.map(function (locale) {
+      const alternates = locales
+        .map(
+          alternate =>
+            `    <xhtml:link rel="alternate" hreflang="${alternate}" href="${toUrl({ baseUrl, locale: alternate, route })}" />`,
```
gndelia: Is that space for anything in particular?

```suggestion
            `<xhtml:link rel="alternate" hreflang="${alternate}" href="${toUrl({ baseUrl, locale: alternate, route })}" />`,
```

portal/vite.config.ts
```diff
@@ -13,6 +15,15 @@ export default defineConfig(function ({ mode }) {
 
   const plugins: PluginOption[] = [react(), polyfills()]
 
+  if (env.PORTAL_SITE_URL) {
+    plugins.push(
+      sitemap({
+        baseUrl: env.PORTAL_SITE_URL,
+        includeHemiEarn: env.VITE_ENABLE_HEMI_EARN_PAGE === 'true',
```
gndelia, line 22: I think here we can read `featureFlags` from `app/featureFlags.ts`

portal/utils/sitemap.ts
```diff
@@ -0,0 +1,51 @@
...
+    /[&<>"']/g,
+    character =>
+      ({
+        '"': '&quot;',
+        '&': '&amp;',
+        "'": '&apos;',
+        '<': '&lt;',
+        '>': '&gt;',
+      })[character]!,
+  )
```
gndelia: IA says that we can use `new URL` and replace most of these

```
const toUrl = ({ baseUrl, locale, route }: {
  baseUrl: string
  locale: Locale
  route: string
}) => escapeXml(new URL(`/${locale}${route}`, baseUrl).toString())
```

Apparently, with `URL`, you only need to handle `&` - the rest are natively handled

portal/utils/sitemap.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const toUrl = ({
+  baseUrl,
+  locale,
+  route,
+}: {
+  baseUrl: string
+  locale: Locale
+  route: string
+}) => escapeXml(`${baseUrl.replace(/\/+$/, '')}/${locale}${route}`)
```
gndelia: Minor, but `baseUrl.replace(/\/+$/, '')` is done for every combination of routes + locales + alternates.

So technically, you can move that line out of the loop and do it once. Minor though

portal/test/utils/sitemapRoutes.test.ts
```diff
@@ -0,0 +1,43 @@
+import fs from 'node:fs'
+import path from 'node:path'
+import { sitemapRoutes } from 'utils/sitemapRoutes'
+import { describe, expect, it } from 'vitest'
+
+// Named on purpose: the previous generator walked the tree and excluded things
+// with `!baseRoute.endsWith('demos')`, which is how `[shareAddress]` slipped in.
+const excluded: Record<string, string> = {
+  '/demos': 'redirects to /ecosystem',
```
gndelia: I think there's no demos folder, so this line may not be needed

portal/README.md
ArturDolzan: Sorted in [`d41d3ae`](https://github.com/hemilabs/ui-monorepo/pull/2230/commits/d41d3aec). One small deviation: the list is grouped by topic with a `#` header and alphabetical inside each group, so the four `SENTRY_*` went into the Sentry group and `PORTAL_SITE_URL` got its own line, since it is about the sitemap rather than error reporting. Happy to flatten it if you meant one plain list.

portal/README.md
ArturDolzan: Gone in [`d41d3ae`](https://github.com/hemilabs/ui-monorepo/pull/2230/commits/d41d3aec). With the vars sorted into the existing groups the header had nothing left to introduce.

portal/utils/sitemap.ts
ArturDolzan: It was indenting the emitted XML, but nobody reads a sitemap by eye and it was the only place doing it. Dropped in [`d41d3ae`](https://github.com/hemilabs/ui-monorepo/pull/2230/commits/d41d3aec).

portal/vite.config.ts
ArturDolzan: Tried it and it does not work: `import.meta.env` is undefined in the config's Node context. I put a probe in with the variable set in the shell and got `import.meta.env.VITE_ENABLE_HEMI_EARN_PAGE = undefined`, so importing `featureFlags` there would read `false` every time and quietly drop `/hemi-earn` from the sitemap.

Same shape as `process.env` not seeing `.env` files, which is why the config reads through `loadEnv`. Happy to open an issue if you think there is a tidier way to share that flag between the two contexts.

portal/utils/sitemap.ts
ArturDolzan: Good call, done in [`d41d3ae`](https://github.com/hemilabs/ui-monorepo/pull/2230/commits/d41d3aec). Checked the cases before switching: `URL` percent-encodes `<`, `>` and `"`, leaves `&` alone (so that one still needs `&amp;`), and leaves `'` alone too, which is fine in XML. It also normalizes a trailing slash on the base, so it took the manual strip with it. The generator lost 11 lines.

portal/utils/sitemap.ts
ArturDolzan: This one solved itself with the `URL` switch above: it normalizes the trailing slash on its own, so the `replace` is gone entirely rather than hoisted.

portal/test/utils/sitemapRoutes.test.ts
ArturDolzan: You are right, removed in [`d41d3ae`](https://github.com/hemilabs/ui-monorepo/pull/2230/commits/d41d3aec). `/demos` is a redirect declared in `app.tsx` with no page on disk, and since that test walks the filesystem the exclusion was never reached.

gndelia (approved):

gabmontes (approved):

arielmergen (approved):

PR #2231, opened 2026-09-01
Restore the security headers with a Cloudflare Worker

ArturDolzan:
### Description

This PR restores the security headers the migration dropped, now served from a Cloudflare Worker instead of an .htaccess.

The generator that emitted the .htaccess went out with the Next build and nothing replaced it, so the branch currently ships no CSP, no HSTS and no frame-ancestors at all. On a UI that signs wallet transactions that is worth fixing before anything else in the deploy phase, and it is the one piece that can be written without Cloudflare being up yet.

I ported the policy exactly as production serves it today rather than adopting the stricter one from vetro's web app. The criterion for a migration is "does it serve the same headers?", and starting from default-src 'none' on an app that runs on 'self' would only surface what was missing once it was already live. The two dead headers (Expect-CT, X-XSS-Protection) are kept for the same reason. Hardening deserves its own PR.

Verified by diffing against the live site: the 11 CSP directives are the same set, 9 of the 10 headers are byte identical, and the only differences are inside connect-src and script-src, in exactly the origins that come from env vars (production has portal-api.hemi.xyz, umami and Sentry; locally those are localhost and switched off).

Worth knowing while reviewing: the Worker only runs because run_worker_first is set, otherwise Cloudflare serves the assets itself and nothing carries the headers. And `pnpm dev` deliberately gets no headers, since upgrade-insecure-requests would rewrite every request to https and break `dev:wifi` on a phone.

Broken down into:
- `375722f9` - Add a security headers builder
- `07411fe9` - Serve the headers from a Cloudflare Worker

## Notes

There is no worker-src directive on purpose. It falls back to default-src 'self' and the five web workers are same-origin, so they keep running. Setting it to 'none', which is what an app without web workers would do, silently stops tunnel history from syncing. That is written down next to the policy so the next person hardening it does not walk into it.

wrangler now needs a build first: the build is what tells it where the assets are, and without it every wrangler command fails asking for a directory property, with no hint that a build is the missing step. Noted in the README and in wrangler.jsonc.

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2194

### Checklist

- [x] Manual testing passed.


portal/package.json
```diff
@@ -70,6 +70,8 @@
     "wagmi": "2.15.7"
   },
   "devDependencies": {
+    "@cloudflare/vite-plugin": "1.23.1",
```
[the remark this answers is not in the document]
gabmontes: @ArturDolzan was this resolved?

gabmontes:
@SocketSecurity ignore npm/workerd@1.20260205.0

portal/package.json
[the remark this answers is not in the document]
ArturDolzan: @gabmontes good catch, I had resolved this one by accident. My bad. It is fixed now in e5953f39.

portal/wrangler.jsonc
```diff
@@ -1,17 +1,16 @@
...
-  // because it has no web workers; copying that verbatim silently kills the
-  // five under portal/workers, and tunnel history simply stops syncing.
+  // No "directory" here: `pnpm build` writes the path wrangler reads, so every
+  // wrangler command needs a build first or it fails asking for this property.
   "assets": {
-    "directory": "./dist",
+    "binding": "ASSETS",
     "not_found_handling": "single-page-application",
+    // Without this Cloudflare serves the assets directly and the Worker never
+    // runs, so no response carries the security headers.
```
gndelia: We don't need to explain CF variables here - let's read the documentation if that's the case. This is AI comment generation to auto explain themselves their decision 😆 

```suggestion
```

portal/utils/securityHeaders.ts
```diff
@@ -0,0 +1,163 @@
...
+const parseUrl = function (url?: string) {
+  if (!url) {
+    return null
+  }
+  try {
+    return new URL(url)
+  } catch {
+    return null
+  }
+}
```
gndelia: 2 things

- the comment is not really useful
- You can try `URL.canParse` https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static instead of trry/catch

portal/server.ts
```diff
@@ -0,0 +1,40 @@
...
+    import.meta.env.VITE_CUSTOM_RPC_URL_SEPOLIA,
+  ],
+  portalApiUrl: import.meta.env.VITE_PORTAL_API_URL,
+  sentryDsn: import.meta.env.VITE_SENTRY_DSN,
+  vetroApiUrl: import.meta.env.VITE_VETRO_API_URL,
+})
+
+export default {
+  async fetch(request, env) {
+    const response = await env.ASSETS.fetch(request)
```
gndelia: You may want to confirm with Joshua, but isn't an option to serve static assets directly without a worker? I think that one allows you to add `_headers` file, so no need for a worker in that case

portal/utils/securityHeaders.ts
```diff
@@ -0,0 +1,163 @@
...
+  apiOrigins
+    .filter(origin => origin !== null)
+    .forEach(origin => domains.add(origin))
+
+  // Only the origin is allow-listed, so a custom RPC carrying a port or a path
+  // still matches the responses the client gets back.
+  customRpcUrls
+    .flatMap(urls => (urls ?? '').split('+'))
+    .map(getOrigin)
+    .filter(origin => origin !== null)
```
gndelia: If you do

```suggestion
    .filter(Boolean)
```

you can drop the `?? null` on both `getOrigin` and `getDomain` which are just replacing `undefined` with `null`

portal/utils/securityHeaders.ts
```diff
@@ -0,0 +1,163 @@
...
+  const fonts = fontDomains.join(' ')
+
+  return {
+    // There is no `worker-src` on purpose: it falls back to `default-src`, and
+    // the five under portal/workers are same-origin. Setting it to 'none', as
+    // an app without web workers would, stops tunnel history from syncing.
+    'Content-Security-Policy': `default-src 'self'; script-src 'self' 'unsafe-inline' ${scriptDomains}; style-src 'self' 'unsafe-inline'; img-src 'self' ${imageSrcUrls.join(' ')} blob: data:; connect-src 'self' ${fetchDomains}; frame-src 'self' ${frameSrcUrls.join(' ')}; frame-ancestors 'none'; block-all-mixed-content; upgrade-insecure-requests; font-src 'self' ${fonts}; style-src-elem 'self' 'unsafe-inline' ${fonts};`,
+    'Cross-Origin-Opener-Policy': 'same-origin-allow-popups',
+    // Expect-CT and X-XSS-Protection are dead in current browsers. They are
+    // kept so this serves exactly what the .htaccess it replaces served.
```
gndelia: If they are dead, let's drop them?

portal/package.json
```diff
@@ -70,6 +70,8 @@
     "wagmi": "2.15.7"
   },
   "devDependencies": {
+    "@cloudflare/vite-plugin": "1.54.0",
+    "@cloudflare/workers-types": "4.20260504.1",
```
gndelia: There's a v5. You may want to check which version better matches the current wrangler version.

Additionally, the docs of that package state

> Note
> We now recommend using the [Wrangler CLI](https://www.npmjs.com/package/wrangler) and the wrangler types command to generate types based on your compatibility date and compatibility flags. You can learn more about this, and how to migrate from @cloudflare/workers-types [here in our docs](https://developers.cloudflare.com/workers/languages/typescript/#generate-types).
> @cloudflare/workers-types will continue to be published on the same schedule.

You may also want to explore if the CLI is a better approach

portal/server.ts
```diff
@@ -0,0 +1,40 @@
+/// <reference types="@cloudflare/workers-types" />
```
gndelia: You may want to confirm, but this may be pulling into React's app

 IA recommended a fix (using a separate tsconfig + extra build command), but it may not be worth it. I think it does not affect bundle size, but the only "issue" is that it may expose types to the portal's code that are not really supposed to be there

portal/plugins/sitemap.ts
```diff
@@ -10,6 +10,7 @@ type Options = {
 }
 
 export const sitemap = ({ baseUrl, includeHemiEarn }: Options): Plugin => ({
+  applyToEnvironment: environment => environment.name === 'client',
```
gndelia, line 13: I think this check will create a new environment, and other plugins (Sentry, Node polyfills) may be included too. Do we need to add this check to those?

portal/wrangler.jsonc
ArturDolzan: Fair, both are gone now in 63d7a497. The one about `directory` I had kept because it documents our build writing that path rather than anything Cloudflare does, but the README already says it, so it was redundant.

portal/utils/securityHeaders.ts
ArturDolzan: Both done in 63d7a497, though I went with `URL.parse` instead: it returns `URL | null` directly, so there is no second parse just to build the object.

Worth knowing what I hit on the way: neither `parse` nor `canParse` exists on the Worker's `URL` in the v4 types. It was bumping to v5 (your other comment) that unlocked it.

portal/server.ts
ArturDolzan: You are right that it would work, and the code review flagged the same thing independently. Every CSP input is a build time constant, so the Worker really is just attaching a fixed header set, and `run_worker_first` bills an invocation per asset to do it.

I would still keep the Worker here: the migration issue asks for headers set per response, and it leaves the door open for anything per request later. Happy to open an issue so we can weigh it properly, since two reviewers landing on the same idea is a decent signal.

portal/utils/securityHeaders.ts
ArturDolzan: Dropped the `?? null` in 63d7a497, thanks. Had to use `.filter(origin => origin !== undefined)` rather than `Boolean` though: `filter(Boolean)` does not narrow the type in TS 5.8, so the compiler still saw `string | undefined` downstream. The predicate form gets inferred.

portal/utils/securityHeaders.ts
ArturDolzan: Agreed, both gone in 63d7a497. Took the chance to fix two more while the "copy it verbatim" rule was already broken: X-Frame-Options was SAMEORIGIN while the CSP said frame-ancestors 'none', and the referrer policy was sending full URLs (so pool and wallet addresses) to every third party.

portal/package.json
ArturDolzan: Both, in 63d7a497. Went with the CLI in the end and dropped the package entirely, since it also types Env from the bindings.

portal/server.ts
ArturDolzan: Confirmed, and it was worse than exposure: `Response.webSocket` compiled in any portal file, so the Worker's Response had replaced the DOM one across the whole app. Anything touching fetch was being checked against the wrong semantics.

Fixed in 63d7a497 by moving the entry to `portal/server/` with its own tsconfig. It had to be a real `tsconfig.json` in that folder, not a named one: the editor only discovers projects by that filename, and my first attempt left the file with no project at all.

portal/plugins/sitemap.ts
ArturDolzan: Good call on Sentry, it was running in both environments in 63d7a497's parent: creating the release twice and uploading the Worker bundle as app code. Scoped to the client now, which also takes ~900 bytes of dead debug id back out of the Worker.

Polyfills are fine as they are. I checked the Worker bundle and it is 7.8 kB with zero occurrences of stream-http, buffer or process: nodePolyfills only injects when something imports http/https, and the Worker does not.

portal/utils/securityHeaders.ts
gndelia: You can tell your Claude that  `@total-typescript` is installed, so `import '@total-typescript/ts-reset/filter-boolean'` can be added [here](https://github.com/hemilabs/ui-monorepo/blob/main/portal/reset.d.ts) so `filter(Boolean)` behaves as you expect

ArturDolzan: You are right, and thanks for pointing at the file. Applied in 6d5d8da3

gndelia (approved):
uACK

portal/README.md
```diff
@@ -13,6 +13,7 @@ Some relevant folders are:
 - [/hooks](./hooks/) folder, which contains reusable hooks to the entire app that are not tied to a specific page.
 - [/messages/](./messages/) folder, which contain a file per locale with all the translated resources.
 - [/public](./public/) folder, which is served verbatim at the site root.
+- [server.ts](./server.ts), the Cloudflare Worker that serves the built assets and sets the security headers on every response.
```
gabmontes: Stale link: `server.ts` was moved to `server/index.ts` in 63d7a497. This should point to `[server/index.ts](./server/index.ts)` instead.

Gabi 🤖

ArturDolzan: Good catch, thanks. Fixed in [`c115401b`](https://github.com/hemilabs/ui-monorepo/pull/2231/commits/c115401b).

gabmontes (approved):

gndelia (approved):

.gitignore
```diff
@@ -13,6 +13,7 @@ packages/*/dist
 # Vite build output and wrangler local state
 portal/.wrangler
 portal/dist
+portal/server/worker-configuration.d.ts
```
joshuasing: Cloudflare recommends committing generated type files, and check it is up-to-date in CI: https://developers.cloudflare.com/workers/languages/typescript/#5-update-your-scripts-and-ci-pipelines

portal/worker/index.ts
joshuasing, line 1: nit: renaming this directory to `portal/worker/` helps make it more obvious this is for a Cloudflare worker.

portal/server/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export default {
+  async fetch(request, env) {
+    const response = await env.ASSETS.fetch(request).catch(
+      () => new Response('Internal Server Error', { status: 500 }),
+    )
+
+    // `upgrade-insecure-requests` rewrites every request to https, which the
+    // dev server does not speak, so `dev:wifi` on a phone would never load.
+    if (import.meta.env.DEV) {
```
joshuasing: This also prevents testing other parts of the CSP when in development mode. I think it may be preferable to disable only known-incompatible parts of the CSP in development mode so the remaining parts can be tested and issues caught earlier.

portal/utils/securityHeaders.ts
```diff
@@ -0,0 +1,147 @@
...
+  const hosts = getThirdPartyHosts(config)
+  const fetchDomains = Array.from(buildFetchDomains(config, hosts)).join(' ')
+  const scriptDomains = Array.from(buildScriptDomains(hosts)).join(' ')
+  const fonts = fontDomains.join(' ')
+
+  return {
+    // No `worker-src` on purpose: it falls back to `script-src`, which allows
+    // 'self', and the five under portal/workers are same-origin. Dropping
+    // 'self' from `script-src` stops tunnel history from syncing.
+    'Content-Security-Policy': `default-src 'self'; script-src 'self' 'unsafe-inline' ${scriptDomains}; style-src 'self' 'unsafe-inline'; img-src 'self' ${imageSrcUrls.join(' ')} blob: data:; connect-src 'self' ${fetchDomains}; frame-src 'self' ${frameSrcUrls.join(' ')}; frame-ancestors 'none'; block-all-mixed-content; upgrade-insecure-requests; font-src 'self' ${fonts}; style-src-elem 'self' 'unsafe-inline' ${fonts};`,
```
joshuasing: I think it may be preferable to build this CSP in parts, similar to what is done here: https://github.com/vetro-protocol/vetro-monorepo/blob/master/web/src/index.ts

I think this would make the policy easier to read, and we may want to conditionally include/exclude parts depending on environment (see https://github.com/hemilabs/ui-monorepo/pull/2231/changes#r3929060505).

joshuasing (changes requested):
Looks good overall, few minor comments below.

There are a few other parts that may need slight modification, but can wait until after we have a worker to test against (should be early tomorrow). Thanks

.gitignore
ArturDolzan: Done in [`fea41f96`](https://github.com/hemilabs/ui-monorepo/pull/2231/commits/fea41f96). They're generated with `--env-file=/dev/null` so local `.env` files don't leak into the interface, and `build` now runs `wrangler types --check`, which puts the freshness check in CI without a separate job.

portal/worker/index.ts
ArturDolzan: Agreed, renamed in [`fea41f96`](https://github.com/hemilabs/ui-monorepo/pull/2231/commits/fea41f96).

portal/server/index.ts
ArturDolzan: You're right, that was throwing away too much. Development now only drops `upgrade-insecure-requests` and keeps the rest of the policy, in [`fea41f96`](https://github.com/hemilabs/ui-monorepo/pull/2231/commits/fea41f96).

portal/utils/securityHeaders.ts
ArturDolzan: Done in [`fea41f96`](https://github.com/hemilabs/ui-monorepo/pull/2231/commits/fea41f96), following the shape in vetro. That's what made the conditional in the other thread clean.

joshuasing (approved):
Thanks, looks good!

gndelia (approved):

Issue #2237, opened 2026-09-04
Fix tunnel so it works with SAFE multisig wallets


gndelia:
### Description

The tunnel reverts when trying to bridge/tunnel assets between Ethereum and Hemi. See this [tenderly simulation](https://dashboard.tenderly.co/public/safe/safe-apps/simulator/904fd0af-2f15-4f06-aa14-94f9fd5705ac?trace=0.1.7.0.0.3.0)

> StandardBridge: function can only be called from an EOA

This error comes from this [solidity modifier onlyEOA](https://github.com/ethereum-optimism/optimism/blob/4e631e337d85860d796ef3e1dad3a89c28811ebd/packages/contracts-bedrock/src/universal/StandardBridge.sol#L105) which is attached to the methods we call (For example, see here [depositETH](https://github.com/ethereum-optimism/optimism/blob/4e631e337d85860d796ef3e1dad3a89c28811ebd/packages/contracts-bedrock/src/L1/L1StandardBridge.sol#L144) which we call in [hemi-tunnel-actions](https://github.com/hemilabs/ui-monorepo/blob/7d67c00b270b9bfdf38ba782c4568200ee37c220/packages/hemi-tunnel-actions/src/depositEth.ts#L99).

In order to support multisig wallets (Note: It may also add support for delegated addresses - see https://github.com/hemilabs/ui-monorepo/issues/1464), we need to switch to the `*to` format. For example, [depositETHTo](https://github.com/ethereum-optimism/optimism/blob/4e631e337d85860d796ef3e1dad3a89c28811ebd/packages/contracts-bedrock/src/L1/L1StandardBridge.sol#L159). This function uses the same signature, with an additional `_to` parameter which is the receiver (in the other method, the `_to` is resolved with the signer value)

We should do this for the methods to bridge from/to Ethereum, both for ETH and ERC20 tokens

> [!NOTE]
> We need to confirm here that 1. addresses are the same for SAFE in different chains and 2. That addresses are deployed (which the user does from the Safe app

The Prove + claim methods can be called by anyone, so they don't face this issues. Note that _we may_ need to implement a similar "receive address" with this new issue  as in #2132

gndelia:
One note from this: As `receiver` may not be the signer, we may need to update the TX history to reflect this (as users will see the TX they signed, but not the tokens they receive from). Maybe this is ok, but maybe not.

gndelia:
[...]

PR #2240, opened 2026-09-08
Add Safe connector behind a feature flag

gndelia:
### Description

This PR adds the Safe connector. It's behind a feature flag because I need to test it, together with #2239 and other changes that are incoming.

I know this PR adds a new env variable, which will conflict with `vite-migration`; however, I need to test this with a real SAFE, and I cannot do it locally. I need to add staging as a custom dApp in the SAFE web, and for that, the branch needs to be deployed (which is not the case of `vite-migration`). If that branch is merged, I'll help Artur solve the conflicts

### Screenshots

N/A

### Related issue(s)

Related to #2237 

### Checklist

- [x] Manual testing passed.


portal/context/evmWalletContext.tsx
```diff
@@ -75,6 +76,7 @@ const connectors = connectorsForWallets(
 export const allEvmNetworksWalletConfig = createConfig({
   chains: allEvmNetworks,
   connectors: [
+    ...(featureFlags.enableSafeWallet ? [safe()] : []),
```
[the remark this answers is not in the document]
gndelia: Done here [bdbd99e](https://github.com/hemilabs/ui-monorepo/pull/2240/commits/bdbd99e5629eda46e6557dd2b1186fc30e4442c2) and I also added a `manifest.json` + svg so it can be read in Safe app

ArturDolzan (approved):

portal/components/chainLogo.tsx
```diff
@@ -5,23 +5,11 @@ import { hemiMainnet } from 'networks/hemiMainnet'
 import { hemiTestnet } from 'networks/hemiTestnet'
 import { mainnet } from 'networks/mainnet'
 import { sepolia } from 'networks/sepolia'
-import { orange600 } from 'styles'
+import Image from 'next/image'
```
gabmontes, line 8: `import/order` warning: `next/image` (external) sits between internal-alias imports. Run `eslint --fix`.

— Auto Gabi

portal/context/evmWalletContext.tsx
```diff
@@ -17,11 +17,12 @@ import {
   walletConnectWallet,
 } from '@rainbow-me/rainbowkit/wallets'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { featureFlags } from 'app/featureFlags'
```
gabmontes, line 20: `import/order` warning: `app/featureFlags` (internal) breaks up the external-imports group. Run `eslint --fix`.

— Gabi 🤖

gabmontes (commented):
These warnings didn't fail CI because `import/order` is set to `"warn"` and the lint script has no `--max-warnings 0`. ESLint only fails on errors, so warnings never block the pipeline.

— Auto Gabi

portal/context/evmWalletContext.tsx
gndelia: what? no

portal/components/chainLogo.tsx
gndelia: no

gndelia:
> These warnings didn't fail CI because `import/order` is set to `"warn"` and the lint script has no `--max-warnings 0`. ESLint only fails on errors, so warnings never block the pipeline.
> 
> — Auto Gabi

this is incorrect

gabmontes (approved):

Issue #2244, opened 2026-09-08
Add HemiStake general stats


gndelia:
These stats need to be added below the hero if the user is disconnected

<img width="908" height="150" alt="Image" src="https://github.com/user-attachments/assets/a2d4218b-0d2e-4e77-a622-f9f4c138dbdc" />

According to the Wireframes, if the user is connected, then these should be moved to the bottom. I am unsure if I like this idea. We can use this issue to implement the cards (UI + hooks if needed) and then use a different issue in relation on how to calcualte each of these values.

LEaving the artifact here (it's private): [claude artifact](https://claude.ai/code/artifact/c34c9ad6-a333-4163-bb4c-f23b334229f2?via=auto_preview)

<img width="1390" height="331" alt="Image" src="https://github.com/user-attachments/assets/c4261e3b-33e4-4630-b9b4-535b0c44c165" />

<img width="1443" height="356" alt="Image" src="https://github.com/user-attachments/assets/87306147-888c-46ab-80b2-057ceefb7c71" />

<img width="388" height="398" alt="Image" src="https://github.com/user-attachments/assets/bb4664c4-2e83-4ce3-be31-187c6fc18702" />

gndelia:
I actually think we should follow Hemi Earn's design here 

<img width="868" height="186" alt="Image" src="https://github.com/user-attachments/assets/dd47a74c-297e-4a9d-a111-41a1f247096b" />

ArturDolzan:
Done.

PR #2247, opened 2026-09-09
Add the Hemi Stake general stats

ArturDolzan:
### Description

This PR adds the stats row to the Hemi Stake page, four cards following the Hemi Earn card design as suggested in #2244. They take the place of the two voting power cards that sat there, so `CardInfo` goes with them, it had no other consumer.

The cards read the `/hemi-stake` endpoint the portal API gained in #2257 through a single fetch, and each one picks its slice with react-query's `select`, so they share one cache entry. That endpoint answers for total staked and wallets staking today. Average lockup and rewards paid show a dash until it covers them, and a pill only renders when there is real data behind it, so nothing on the row is a placeholder. The lockup count on the third card comes straight from the contract.

One thing to flag: the endpoint is not deployed yet, since #2257 skipped the API version bump on purpose. Until it ships, a deployed build reads all four cards as unavailable.

Split into four commits:
- `3de4e4a0` - Drop the voting power summary
- `18ea0f68` - Add the Hemi Stake stats cards
- `734c605f` - Drop the stats section heading
- `ed019aec` - Measure rewards by what the card can price

### Related issue(s)

Related to #2244

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-stake/_components/totalStaked.tsx
```diff
@@ -0,0 +1,44 @@
...
+  )
+
+  return (
+    <StakeStatCard
+      badge={
+        <>
+          <TrendUpIcon />
+          {t('total-staked-badge', {
+            percentage: formatPercentage(growth?.percentage ?? 0),
+          })}
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/totalStaked.tsx]: Good catch. The sign now comes from the value instead of the icon, and the arrow only renders when it's positive. Did the same for the wallets badge, which had a hardcoded `+` that would have rendered "+-12" on a negative week. Changed in [`35c2a856`](https://github.com/hemilabs/ui-monorepo/pull/2247/commits/35c2a85687e8999e213576e594d531860f85f73d)

portal/app/[locale]/hemi-stake/_hooks/useStakeStats.ts
```diff
@@ -0,0 +1,25 @@
...
+]
+
+export const useStakeStats = function <TData = StakeStats>(
+  select?: (data: StakeStats) => TData,
+) {
+  const chainId = useHemi().id
+
+  return useQuery({
+    queryFn: fetchStakeStats,
+    queryKey: getStakeStatsQueryKey(chainId),
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_hooks/useStakeStats.ts:20]: This is a fixture, the numbers are invented on both networks, so there is no real mainnet data leaking into testnet. Making the mock network aware would mean inventing testnet token addresses for the rewards too, and all of it gets replaced the moment the endpoint lands. The chain is in the query key, so nothing is reused across a switch.

portal/app/[locale]/hemi-stake/_hooks/useWeeklyDeltas.ts
```diff
@@ -0,0 +1,23 @@
...
+// records yet.
+export const useWeeklyDeltas = function () {
+  const [networkType] = useNetworkType()
+
+  return useQuery<{ staked: number; wallets: number }>({
+    queryFn: () =>
+      new Promise(resolve =>
+        setTimeout(() => resolve({ staked: 2.1, wallets: 312 }), 2000),
+      ),
+    queryKey: getWeeklyDeltasQueryKey(networkType),
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_hooks/useWeeklyDeltas.ts:21]: Same story as the stats fixture: these deltas are hardcoded placeholders, not mainnet readings, so there is nothing network specific to honor yet. `networkType` is in the key so switching does not reuse the entry.

portal/app/[locale]/hemi-stake/_icons/trendUpIcon.tsx
```diff
@@ -0,0 +1,11 @@
+export const TrendUpIcon = () => (
+  <svg
+    fill="none"
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_icons/trendUpIcon.tsx:4]: Done in [`cd54deba`](https://github.com/hemilabs/ui-monorepo/pull/2247/commits/cd54deba). The direction lives in the text now, so the triangle is purely decorative.

portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx
```diff
@@ -0,0 +1,55 @@
...
+    ]
+  })
+
+  return (
+    <StakeStatCard
+      badge={paid
+        .map(({ amount, symbol }) => `${amount} ${symbol}`)
+        .join(' + ')}
+      isError={(isError && data === undefined) || isPricesError}
+      isLoading={isPending || isPricesPending}
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx]: `portal/.env` ships a default for `VITE_PORTAL_API_URL`, so the disabled path is not really reachable, and chaining `isPending` like this is what `stake/page.tsx` and `stakeAssetsTable` already do. You did point at a real seam though: the badge only needs token amounts, no prices. Split them in [`8aa3eba9`](https://github.com/hemilabs/ui-monorepo/pull/2247/commits/8aa3eba9), so a price failure now renders `-` in the value and keeps the amounts in the badge.

portal/app/[locale]/hemi-stake/_components/averageLockup.tsx
```diff
@@ -0,0 +1,30 @@
...
+  averageLock: stats.averageLock,
+  locksCount: stats.locksCount,
+})
+
+export const AverageLockup = function () {
+  const t = useTranslations('hemi-stake.stats')
+  const { data, isError, isPending } = useStakeStats(selectLockup)
+
+  const years =
+    Math.round(((data?.averageLock ?? 0) / secondsPerYear) * 10) / 10
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/_components/averageLockup.tsx]: I think we should convert to the closest unit available. Note that users can lock into less than one year (months), so say the average is 6 months; this would render as `0.5` years. Let's better convert to years, months, weeks or days (Just becaues the min amount is 6 days) depending on what's the best value.

portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx
```diff
@@ -0,0 +1,59 @@
...
+          )
+        ) : isPricesError ? (
+          '-'
+        ) : (
+          <StatValueSkeleton />
+        )
+      }
+    />
+  )
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx:108]: On second thought, I wonder if we should use a similar pattern here as Hemi Earn and show the rewards paid on hover

Image

Like this ☝🏽, but listing the rewards. Depending on the amount, rewards may be large, and it won't fit on one line.

gndelia (commented):
Note that not all the stats are ready to be implemented (For example rewards depend on the updates of the rewards contract) so be aware that we may need to hide eventually some of these properties if we don't get to implement them on time

For example: There's a high change there's no value for the delta change of wallets staking

portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx
Comment (gndelia) [reply portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx:108]: If so, may be worth using EarnCard (I think it's that component)

portal/app/[locale]/hemi-stake/_components/stakeStatCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+          ) : isError ? (
+            '-'
+          ) : (
+            <StatValueSkeleton />
+          )}
+        </p>
+        <div className="flex min-h-4 items-center">
+          {!isLoading && !isError ? (
+            <span className="flex w-fit items-center gap-x-1 rounded-md border border-solid border-neutral-200 bg-neutral-100 px-1.5 py-0.5 text-xxs font-medium text-neutral-600">
+              {badge}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/_components/stakeStatCard.tsx]: AI suggestion:

> The badge span renders unconditionally; RewardsPaid can pass '' when the rewards array is empty or no address resolves, leaving a bordered empty pill. hemi-earn/_components/earnCard.tsx guards it with badge && ... empty:hidden.

It seems another strong reason to just use Hemi Earn card?

portal/app/[locale]/hemi-stake/_components/averageLockup.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/averageLockup.tsx]: Good catch, anything under a year was rendering like `0.5 yrs`. It now picks the largest unit that fits (years, months, weeks or days) in [`5bfbcf46`](https://github.com/hemilabs/ui-monorepo/pull/2247/commits/5bfbcf46).

portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx:108]: Done in [`5bfbcf46`](https://github.com/hemilabs/ui-monorepo/pull/2247/commits/5bfbcf46), same shape as `FromPoolsBadge`: the pill keeps the compact summary and the tooltip lists one row per reward token with its logo, USD value and amount. I kept the summary text in the pill since that's what the wireframe shows, happy to swap it for a count if you'd rather.

portal/app/[locale]/hemi-stake/_components/stakeStatCard.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/stakeStatCard.tsx]: The empty pill was real, thanks. The card no longer draws the pill, it renders whatever badge the caller passes, in [`5bfbcf46`](https://github.com/hemilabs/ui-monorepo/pull/2247/commits/5bfbcf46).

On reusing EarnCard I'd rather not do it here. We've got EarnCard, PoolCard, CardInfo and now this one, so I'd rather unify them in a separate issue where the refactor gets proper scope.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2247#pullrequestreview-5157653397):
> Note that not all the stats are ready to be implemented (For example rewards depend on the updates of the rewards contract) so be aware that we may need to hide eventually some of these properties if we don't get to implement them on time

Agreed. The wallets delta is the weakest one, it needs a periodic snapshot that nothing records today, so it's first in line to be hidden.

gndelia (approved):

portal/app/[locale]/hemi-stake/_hooks/usePositionsVotingPowerSum.ts
```diff
@@ -1,31 +1,4 @@
...
-  tokenIds: bigint[]
-}) => [
-  'positions-voting-power-sum',
-  chainId,
-  ownerAddress?.toLowerCase(),
-  tokenIds.map(String).sort().join(','),
-]
 
 /** Prefix for invalidation: matches all position sum queries for this user/chain. Requires ownerAddress so invalidation reliably matches cached keys. */
 export const getPositionsVotingPowerSumQueryKeyPrefix = ({
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_hooks/usePositionsVotingPowerSum.ts:4]: Both hooks are gone, but `useStake`, `useUnlock`, `useIncreaseAmount` and `useIncreaseUnlockTime` still invalidate these keys — 8 no-op call sites. It also orphans `getTotalVotingPower` and `getPositionsVotingPowerSum` in `packages/ve-hemi-actions`, whose only consumers left are their own tests (knip misses it, no entry for that workspace). Dropping the invalidations would make both key-builder files unused too.

— Gabi 🤖

portal/app/[locale]/hemi-stake/_hooks/useStakeStats.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const useStakeStats = function <TData = StakeStats>(
+  select?: (data: StakeStats) => TData,
+) {
+  const chainId = useHemi().id
+
+  return useQuery({
+    enabled: portalApiUrl !== undefined && isValidUrl(portalApiUrl),
+    queryFn: fetchStakeStats,
+    queryKey: getStakeStatsQueryKey(chainId),
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_hooks/useStakeStats.ts:24]: Nothing invalidates this key, so after a stake or unlock the cards show pre-transaction numbers for up to 5 minutes. Note an `onSettled` invalidation won't make it instant either: the backend wraps `/hemi-stake` in `pSwr` at 5 minutes, so the refetch gets the cached body and only triggers revalidation.

— Auto Gabi

portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const locale = useLocale()
+  const { data: prices, isError: isPricesError } = useTokenPrices()
+  const t = useTranslations('hemi-stake.stats')
+  const { data, fetchStatus, isPending, status } = useStakeStats(selectRewards)
+
+  const isUnavailable =
+    isDataUnavailable({ fetchStatus, status }) ||
+    (data !== undefined && data.length === 0)
+
+  const paid = (data ?? []).flatMap(function ({ address, amount, chainId }) {
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx:49]: Two silent under-reports once rewards arrive. If no address resolves, `flatMap` drops everything, `isUnavailable` stays false (it only checks `data.length === 0`) and the card renders a confident "$0" instead of unavailable; partial resolution loses tokens the same way. And `getTokenPrice` falls back to `'0'` when the symbol is missing from `/prices`, so that token keeps its amount in the badge but adds $0 to the headline.

Using `paid.length` for the unavailable state covers the first. For the second, skipping unpriced tokens or flagging the total as partial.

— Auto Gabi

portal/app/[locale]/hemi-stake/_hooks/usePositionsVotingPowerSum.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_hooks/usePositionsVotingPowerSum.ts:4]: Accurate on all three counts, and I'd checked the invalidations but not the package side. I'm keeping them: the voting power is coming back as its own card, we just have not settled where, and those two actions plus the key builders are what it will need. Worth knowing the knip blind spot is real either way, `.knip.json` has no entry for `packages/*`, so nothing there gets analysed.

portal/app/[locale]/hemi-stake/_hooks/useStakeStats.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_hooks/useStakeStats.ts:24]: Right on both halves, and the second one is why I'd rather not patch it here: the route sets `revalidate: 5 * 60 * 1000`, so a client invalidation buys a request and the same body. The fix belongs on the backend, and the endpoint is not even deployed yet. Happy to open an issue so we can pick the cache window with the numbers in front of us.

portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/rewardsPaid.tsx:49]: Good catch on the first one, the guard now counts `paid` instead of the raw list in [`ed019aec`](https://github.com/hemilabs/ui-monorepo/pull/2247/commits/ed019aec), so unresolved tokens read as unavailable rather than a confident $0. On the unpriced token I'd rather not decide in this PR: skipping it moves the total silently too, and the card has no way to say "partial" today. Happy to open an issue for that half.

gabmontes (approved):

gndelia (approved):

PR #2253, opened 2026-09-09
Rename the governance staking page to Hemi Stake

### LINKED ISSUE #2249: Update the "Governance Staking" page into "Hemi Stake" with a more prominent link
- [x] Rename "Governance Staking" into "Hemi Stake"

Image

Image

- [x] Make the primary link be `/hemi-stake`. Redirect from `/staking-dashboard/` to the new one

- [x] The Boost stake will now be just an individual link, preserving the old logo, below the Genesis Drop link

Image

[Figma](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=21363-223&m=dev)

Eventually, I think we may want to make it the main page upon entering "app.hemi.xyz" (not Hemi Earn, not the tunnel)

ArturDolzan:
### Description

This PR moves the governance staking page to `/hemi-stake`, renames it to Hemi Stake, and splits the nav accordion so the two staking products stop sharing an entry. The old URL redirects, keeping the locale and the query string.

It lands as one big commit plus two small ones because the types tie the rename together: message keys come from `en.json` and analytics event names from a const array, so the folder, the namespace and the event list can't land separately without breaking the build. Each commit was checked in isolation, not just the tip.

- `a1266edc` - Rename the governance staking page to Hemi Stake
- `56ade867` - Flatten the stake group in the sidebar
- `c148ad4a` - Update the docs for the Hemi Stake rename

#### Notes

Renaming the thirty analytics events breaks their history in Umami. Boost staking now paints orange when selected instead of grey, since a plain entry uses the standard treatment the accordion didn't.

The vocabulary inside the moved folder still says "staking dashboard" (`stakingDashboardContext` and friends). That's a deliberate follow-up, it's mechanical churn that deserves its own diff.

The landing page idea from the issue is not here. I had it working and pulled it back out: it's three places rather than one, and on testnet `/hemi-stake` renders `TestnetDisabled`, so entering with `?networkType=testnet` would land on an unavailable screen.

### Screenshots

https://github.com/user-attachments/assets/f4ab56a3-8905-4df4-a58e-eb1579428d83

### Related issue(s)

Closes #2249

### Checklist

- [x] Manual testing passed.


gndelia (commented):
As it's unlikely that Hemi Earn goes live soon, you may want to test as well that all looks good without Hemi Earn link

portal/app/[locale]/_components/navbar/_components/hemiStake.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <ItemLink
+      event="nav - hemi stake"
+      href="/hemi-stake"
+      icon={<HemiStakeIcon />}
+      rightSection={
+        <div className="ml-auto hidden md:flex md:items-center">
+          <span className="body-text-caption flex items-center rounded-md bg-orange-600 px-1.5 py-px text-white">
+            {t('new')}
+          </span>
```
Comment (gndelia) [inline portal/app/[locale]/_components/navbar/_components/hemiStake.tsx]: Come to think of it, it seems this is a reimplementation of the [badge](https://github.com/hemilabs/ui-monorepo/blob/main/portal/components/badge.tsx) component. Perhaps we just need a variant with a new background color and call it a day

portal/components/icons/hemiStakeIcon.tsx
```diff
@@ -0,0 +1,50 @@
+import { ComponentProps } from 'react'
+
+const coin =
+  'M5.25 0C4.56056 -1.02735e-08 3.87787 0.135795 3.24091 0.399632C2.60395 0.66347 2.0252 1.05018 1.53769 1.53769C1.05018 2.0252 0.66347 2.60395 0.399632 3.24091C0.135795 3.87787 0 4.56056 0 5.25C0 5.93944 0.135795 6.62213 0.399632 7.25909C0.66347 7.89605 1.05018 8.4748 1.53769 8.96231C2.0252 9.44982 2.60395 9.83653 3.24091 10.1004C3.87787 10.3642 4.56056 10.5 5.25 10.5C6.64239 10.5 7.97774 9.94688 8.96231 8.96231C9.94688 7.97774 10.5 6.64239 10.5 5.25C10.5 3.85761 9.94688 2.52226 8.96231 1.53769C7.97774 0.553123 6.64239 2.07482e-08 5.25 0ZM5.25 1.575L8.575 4.9H6.7375V8.925H3.7625V4.9H1.925L5.25 1.575Z'
+
+const sparkle =
+  'M5.5 0C6.05 3.11667 7.88333 4.95 11 5.5C7.88333 6.05 6.05 7.88333 5.5 11C4.95 7.88333 3.11667 6.05 0 5.5C3.11667 4.95 4.95 3.11667 5.5 0Z'
```
gndelia: I think this makes reading the SVG and updating it harder. IMHO the best option is just copy-paste from Figma as close as real svg as possible. This way it's easier, if needed, to copy the SVG into a svg viewer online page to review the icon

gndelia: It does not matter if these SVG values get duplicated imho

portal/app.tsx
```diff
@@ -35,12 +35,13 @@ import {
...
-// dashboard, and the folder shipped misspelled. Everything else is handled by
-// the fallback in the locale layout.
+// The entry paths that do not simply gain a locale: `/stake` pointed at the
+// dashboard, and the governance staking page moved, having also shipped
+// misspelled. Everything else is handled by the fallback in the locale layout.
 const pathAliases = {
   '/stake': '/stake/dashboard',
-  '/staking-dashbord': '/staking-dashboard',
+  '/staking-dashboard': '/hemi-stake',
+  '/staking-dashbord': '/hemi-stake',
```
gndelia: No, I don't think we should redirect from a misspelled URL. Let them go to a 404

gndelia:
On mobile, the Hemi Stake logo seems smaller than the others

image

On desktop, it seems the other way around - the hemi stake icon seems bigger than the others. 

image


Let's adjust them in both cases

portal/app/[locale]/hemi-stake/page.tsx
```diff
@@ -51,23 +49,20 @@ function StakingContent() {
...
-      <div className="flex flex-col">
-        <PageTitle title={t('heading', { symbol })} />
-        {isEnabled ? (
-          <StakingContent />
-        ) : (
-          <TestnetDisabled subtitle={t('switch-to-start-staking')} />
-        )}
-      </div>
-    </PageLayout>
+    <div className="flex flex-col">
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/page.tsx]: Was PageLayout here removed and dropped for any particular reason?  This changes the layout a bit - padding is removed in comparison to prod

Prod

Image

vs your branch

Image

Notice the difference on spacing

portal/app/[locale]/_components/navbar/_components/hemiStake.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/_components/navbar/_components/hemiStake.tsx]: You're right, and it did start as the badge. I pulled it out because the shared one fixes `h-4` while the design wants `px-1.5 py-px`, which lands at 18px, and the only other badge sits inside a button in the positions table so growing it there felt wrong. Added a `size` instead: `xSmall` keeps today's height, `small` is the new one. Back on the badge in [`eb943467`](https://github.com/hemilabs/ui-monorepo/pull/2253/commits/eb943467).

portal/components/icons/hemiStakeIcon.tsx
ArturDolzan: Fair enough, done in [`eb943467`](https://github.com/hemilabs/ui-monorepo/pull/2253/commits/eb943467). The paths are inline now, no constants and no group transforms, so the file pastes straight into a viewer. I baked the sparkle offsets into the coordinates, which duplicates the values the way you said is fine.

portal/app.tsx
ArturDolzan: Done in [`27e69e51`](https://github.com/hemilabs/ui-monorepo/pull/2253/commits/27e69e51). Just so it's on record, that alias predates this PR and I was only retargeting it, so this drops a redirect that had been around for a while.

portal/app/[locale]/hemi-stake/page.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/page.tsx]: Reverted in [`7c43c0ed`](https://github.com/hemilabs/ui-monorepo/pull/2253/commits/7c43c0ed). It came out of a code review note: the route layout already renders a PageLayout and the page rendered a second one, so the padding was landing twice. You're right that it changes spacing against prod though, and a rename PR isn't the place for it.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2253#pullrequestreview-5162024338):
> As it's unlikely that Hemi Earn goes live soon, you may want to test as well that all looks good without Hemi Earn link

Checked with the flag off. The sidebar reads Hemi Stake, Tunnel, DEXs, Genesis Drop, Boost staking, with no gap where Hemi Earn sits and nothing else shifting.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2253#issuecomment-5611863526):
> On mobile, the Hemi Stake logo seems smaller than the others (...) On desktop, it seems the other way around

Good catch. The cause was proportion rather than box size: the viewBox left the coin at 70% of it while every other icon fills its own, so it measured 14px against their 12px on desktop and 19.6px against 28px on mobile. Fixed in [`eb943467`](https://github.com/hemilabs/ui-monorepo/pull/2253/commits/eb943467), the coin now matches Tunnel at both. Before and after screenshots:

Desktop
hemi-stake-icone-antes-desktop
hemi-stake-icone-depois-desktop



Mobile
hemi-stake-icone-antes-mobile
hemi-stake-icone-depois-mobile

portal/app/analyticsEvents.ts
gabmontes, line 1: Let's validate this change with marketing or whoever is used to pull analytics data to be sure we are not altering their stats.

gndelia: 100% nobody is using it

gndelia (approved):

portal/app/analyticsEvents.ts
gabmontes: :agitate:

gabmontes (approved):

PR #2256, opened 2026-09-09
portal: setup CF workers, improve security headers

### LINKED ISSUE #1563: Add Preview deployments on PRs

As suggested by @ArturDolzan , it would be ideal to have Preview deployments on our PRs, so whenever changes in the UI are done, we get a static version deployed that we can easily test, and there's no need to locally start the environment.

As the portal is a static page on Hostinger, in order to avoid building something custom for Hostinger, I thought we could use Cloudflare (As we did in other apps) to just serve the static page. That should be straightforward.

There are some nuances with the APIs (portal-backend and subgraph-api), but for most cases we can get by using the current API for everything. We can later see if this becomes a problem, and a new environment for the API is needed.

@joshuasing told me he could help us set up the Cloudflare environment for the portal to be served.

joshuasing:
### Description

Prepares the Vite-based Portal for deployment through Cloudflare Workers.

- Configures production and staging Worker environments, observability, preview URLs, and source-map uploads.
- Adds per-request CSP nonces compatible with Cloudflare Challenges.
- Hardens CSP directives and completes allowlists for WalletConnect, Coinbase Wallet, Binance Wallet, analytics, and Sentry.
- Disables Coinbase inline telemetry, which is incompatible with the nonce-based CSP.
- Adds missing cross-origin, agent-cluster, DNS-prefetch, framing, and browser permissions policies.

### Screenshots

N/A

### Related issue(s)

Related to #2194
Closes #1563

### Checklist

- [x] Manual testing passed.


portal/utils/securityHeaders.ts
```diff
@@ -166,12 +255,15 @@ export const buildSecurityHeaders = function (config: SecurityHeadersConfig) {
       fonts: fontDomains,
       isDev: config.isDev ?? false,
       scriptDomains: Array.from(buildScriptDomains(hosts)),
+      scriptNonce: config.scriptNonce,
     }),
     'Cross-Origin-Opener-Policy': 'same-origin-allow-popups',
-    'Permissions-Policy': 'geolocation=(), microphone=()',
+    'Cross-Origin-Resource-Policy': 'same-origin',
+    'Origin-Agent-Cluster': '?1',
+    'Permissions-Policy': permissionsPolicy,
```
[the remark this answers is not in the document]
joshuasing: HSTS is handled as a security setting in Cloudflare, and is not required here.

portal/utils/securityHeaders.ts
```diff
@@ -69,23 +131,43 @@ const buildFetchDomains = function (
...
+    'https://notify.walletconnect.org',
+    'https://pulse.walletconnect.com',
+    'https://pulse.walletconnect.org',
+    'https://push.walletconnect.com',
+    'https://push.walletconnect.org',
+    'https://relay.walletconnect.com',
+    'https://relay.walletconnect.org',
+    'https://rpc.walletconnect.com',
+    'https://rpc.walletconnect.org',
+    'wss://relay.walletconnect.com',
```
gndelia, line 153: apparently we need to add `verify.walletconnect.org` here as well, so we can fetch to it

portal/wrangler.jsonc
```diff
@@ -1,12 +1,31 @@
...
+    },
+  },
   "assets": {
     "binding": "ASSETS",
     "not_found_handling": "single-page-application",
     "run_worker_first": true,
   },
+  "env": {
+    "production": {
+      "name": "hemi-portal",
```
gndelia, line 24: AI is telling me that running `wrangler deploy` and `wrangler deploy --env production` both end up deploying against `hemi-portal` (which would be production), but depending on the command is the value for `preview_urls` that is used. Minor, but I'm flagging it in case you want to rename here so it doesn't match the top value in line 3 (so the 2 commands deploy to different workers)

portal/utils/securityHeaders.ts
```diff
@@ -69,23 +131,43 @@ const buildFetchDomains = function (
...
     'https://bscrpc.com',
     'https://rpc.ankr.com/bsc',
     'wss://nbstream.binance.click',
     'wss://nbstream.binance.com',
     'wss://nbstream.binance.info',
+    'wss://nbstream.yshyqxx.com',
+    // Merkle
     'https://api.merkl.xyz',
+    // Cloudflare Web Analytics
+    'https://cloudflareinsights.com',
```
gndelia, line 173: note that this will add analytics on staging - I am unsure if we want that (That's why it was set on line 104 as `hosts.analytics` is only expected to be enabled on prod)


But I guess those analytics are not the same as these, which are the ones you use for checking health on CF - right?

portal/utils/securityHeaders.ts
```diff
@@ -15,21 +16,86 @@ const parseUrl = (url?: string) => (url ? URL.parse(url) : null)
...
+  'https://verify.walletconnect.com',
+  'https://verify.walletconnect.org',
+  'https://keys.coinbase.com',
+]
+
+const permissionsPolicy = [
+  'accelerometer',
+  'ambient-light-sensor',
+  'attribution-reporting',
+  'autoplay',
```
ArturDolzan: Perhaps it's worth checking `autoplay=()` against the navbar video (`videoAsset.tsx:22` uses `autoPlay muted`). An empty allowlist seems to disable it for self too, and Chrome's muted-autoplay exemption doesn't apply once the policy is off. Maybe `autoplay=(self)`?

portal/utils/securityHeaders.ts
joshuasing: Cloudflare Insights is enabled on staging and reports useful metrics like Core Web Vitals - because this is enabled at the proxy level, I think it is best to keep always-enabled here.

portal/wrangler.jsonc
joshuasing: With the way Worker Builds is configured, I don't think this would cause a problem - but I think it is best to default to staging, so I will update the root `name`. Thanks!

portal/utils/securityHeaders.ts
joshuasing: Good catch, fixed. Thanks!

portal/utils/securityHeaders.ts
joshuasing: Thanks, added

gndelia (approved):

ArturDolzan (approved):

Issue #2243, opened 2026-09-08
Add HEMI pricing track 


gndelia:
<img width="2390" height="319" alt="Image" src="https://github.com/user-attachments/assets/ed7adb5e-84a6-4fdc-9b70-73bf632e4e2f" />

Something like this, but using our design system.

Current price of $HEMI is returned from `prices` API - the only tricky part is to get the 24h % change. Perhaps the initial implementation can only list the price, and later add that extra value.

It could be done next to the hero, or next to the Wallet Connection Button, TBD. See https://github.com/hemilabs/ui-monorepo/issues/2243#issuecomment-5591343699 for implementation

gndelia:
I generated these mockups with Claude

I think they follow the styles from the Network Selector, so we can use that too - the screenshots are guidelines

<img width="1457" height="361" alt="Image" src="https://github.com/user-attachments/assets/5cc27a2c-88df-4834-98e5-f0ca32499efe" />

<img width="1372" height="535" alt="Image" src="https://github.com/user-attachments/assets/6cb8550b-7fb2-4ab6-8a81-cd1cd5d5e1b8" />

<img width="1331" height="304" alt="Image" src="https://github.com/user-attachments/assets/856aef79-4599-4d87-97a9-1c96e2f78d04" />

Leaving the [artifact link here](https://claude.ai/code/artifact/b52fc763-375f-47f4-b6cf-189c3028a08f?via=auto_preview) for my future reference but it's private

PR #2259, opened 2026-09-10
Add the token price pill

ArturDolzan:
### Description

This PR adds the price pill: token mark, price, and a link out. It's wired to nothing yet, it only goes into Storybook, so the four Figma states plus one for a missing price can be exercised before we decide where it lives.

The price runs through the crypto rounder rather than the fiat one, which fixes two decimals and would render HEMI's price as `$0.02`. Everything else reuses what's there: the shell is the theme's small shadow, the focus ring is the secondary button's, and the arrow is the same one the sidebar uses for external links.

The 24h change is out, as the issue itself suggests.

### Screenshots

https://github.com/user-attachments/assets/7e7f4fa2-9244-4c7a-9551-8c8d6fda68c8

### Related issue(s)

Related to #2243

### Checklist

- [x] Manual testing passed.


portal/components/tokenPricePill.tsx
```diff
@@ -0,0 +1,60 @@
...
+          containerClassName="flex items-center"
+          height={10}
+          width={44}
+        />
+      ) : (
+        <>
+          <span className="body-text-medium text-neutral-950">
+            <ErrorBoundary fallback="-">
+              {price === undefined ? '-' : `$${formatNumber(price)}`}
+            </ErrorBoundary>
```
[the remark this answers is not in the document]
ArturDolzan: You're right, the formatting ran in the parent's render so the boundary never saw it. Moved it into a child component, the same shape as `RenderFiatBalanceUnsafe`, in [`e11f49b6`](https://github.com/hemilabs/ui-monorepo/pull/2259/commits/e11f49b6). Checked it with a malformed price and the pill renders `-` now instead of throwing.

portal/components/tokenPricePill.tsx
```diff
@@ -0,0 +1,60 @@
...
+}: Props) {
+  const t = useTranslations('common')
+
+  return (
+    <ExternalLink
+      className="group/token-price-pill flex h-7 w-fit items-center gap-x-1.5 rounded-md bg-white px-2.5 shadow-sm transition-colors duration-300 hover:bg-neutral-50 focus-visible:shadow-button-secondary-focused focus-visible:outline-none forced-colors:focus-visible:outline"
+      href={href}
+      onClick={onClick}
+    >
+      <TokenLogo size="xSmall" token={token} />
```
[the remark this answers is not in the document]
ArturDolzan: Good catch, done in [`e11f49b6`](https://github.com/hemilabs/ui-monorepo/pull/2259/commits/e11f49b6). The logo sits in an `aria-hidden` wrapper now, so the link's accessible name is "HEMI price $0.0154" instead of repeating the token.

portal/components/tokenPricePill.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const t = useTranslations('common')
+
+  return (
+    <ExternalLink
+      className="group/token-price-pill flex h-7 w-fit items-center gap-x-1.5 rounded-md bg-white px-2.5 shadow-sm transition-colors duration-300 hover:bg-neutral-50 focus-visible:shadow-button-secondary-focused focus-visible:outline-none forced-colors:focus-visible:outline"
+      href={href}
+      onClick={onClick}
+    >
+      <span aria-hidden>
+        <TokenLogo size="xSmall" token={token} />
```
gndelia: I would use the L1 format version of the logo, as we're using the price of the hemi token for all chains

portal/components/tokenPricePill.tsx
```diff
@@ -0,0 +1,64 @@
...
+import Skeleton from 'react-loading-skeleton'
+import { type Token } from 'types/token'
+import { useTranslations } from 'use-intl'
+import { formatNumber } from 'utils/format'
+
+import { ErrorBoundary } from './errorBoundary'
+import { ExternalLink } from './externalLink'
+import { TokenLogo } from './tokenLogo'
+
+const Price = ({ price }: { price: string }) => <>{`$${formatNumber(price)}`}</>
```
gndelia: I think formatNumber is used to format crypto (up to 6 decimals) which for USD wise it's a lot. Maybe https://github.com/hemilabs/ui-monorepo/blob/94365d39d6f16a91e6a7c428048528d41310466a/portal/utils/format.ts#L28 ?

portal/components/tokenPricePill.tsx
ArturDolzan: Good point, done in [`484f4920`](https://github.com/hemilabs/ui-monorepo/pull/2259/commits/484f4920). Switched to the L1 logo, which drops the network badge the default asset carries.

token-price-pill-default

portal/components/tokenPricePill.tsx
ArturDolzan: Both are off for this: `formatNumber` truncates and drops a digit on big values, `formatFiatNumber` renders HEMI's price as `$0.02`. Added `formatTokenPrice` in [`6fdd00ff`](https://github.com/hemilabs/ui-monorepo/pull/2259/commits/6fdd00ff), `smartRound(6, 2, 8)`, so it's two decimals from a dollar up and keeps the digits below it. Binance shows HEMI as `$0.00723137`.

gndelia: ah, true, I guess it makes sense to show more than 2 digits now because price is below 0.001 😭

portal/components/tokenPricePill.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <ExternalLink
+      className="group/token-price-pill flex h-7 w-fit items-center gap-x-1.5 rounded-md bg-white px-2.5 shadow-sm transition-colors duration-300 hover:bg-neutral-50 focus-visible:shadow-button-secondary-focused focus-visible:outline-none forced-colors:focus-visible:outline"
+      href={href}
+      onClick={onClick}
+    >
+      <span aria-hidden>
+        <TokenLogo size="xSmall" token={token} version="L1" />
+      </span>
+      <span className="sr-only">
```
gndelia, line 42: Very nice, I did not know about `sr-only` !  Interesting

gndelia (approved):

gabmontes (approved):

PR #2261, opened 2026-09-10
Migrate from Next to Vite

### LINKED ISSUE #2194: Migrate the portal from Next.js to Vite and deploy it on Cloudflare

ArturDolzan:
### Description

As the title says.

### Screenshots

N/A.

### Related issue(s)

Closes #2194 

### Checklist

- [x] Manual testing passed.


gndelia:
It's happening

portal/wrangler.jsonc
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "env": {
+    "production": {
+      "name": "hemi-portal",
+      "preview_urls": false,
+    },
+    "staging": {
+      "name": "hemi-portal-staging",
+    },
+  },
```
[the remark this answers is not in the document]
gndelia: @gabmontes had agreed to remove the beta env, so I think we're ok for now?

portal/README.md
```diff
@@ -4,14 +4,16 @@ The Portal is a Web3 app that allows users to interact with Hemi, an L2 that int
 
 ## Structure
 
-The portal is a Next app that uses static building and the app router. As it is a static page, it does not rely with SSR features (Except on local development).
+The portal is being migrated from Next to Vite (see [#2194](https://github.com/hemilabs/ui-monorepo/issues/2194)). The build runs on Vite and produces a static bundle, and routing is now react-router: `app.tsx` at the root holds the route table, and `main.tsx` is the entry. It has never relied on SSR. Every page is wired in `app.tsx`.
```
gndelia: This is now stale 😛 

I guess we can send a different PR for this though...

ArturDolzan: Fixed in c16305401facbb1d2b98415919fa6292a5e949cb. Thanks!

gndelia (approved):

PR #2263, opened 2026-09-10
Upgrade pnpm to v11

joshuasing:
### Description

Upgrade the repository from pnpm 10.33.4 to pnpm 11.24.0.

- Move `saveExact` from `.npmrc` to `pnpm-workspace.yaml`
- Replace legacy build dependency settings with `allowBuilds`
- Update all `package-manager` pins
- Regenerate the lockfile with pnpm 11
- Add the pnpm lockfile to `.prettierignore`
- Update minimum Node version (`engines.node`) to `>=24`

### Screenshots

N/A - no UI changes

### Related issue(s)

N/A.

### Checklist

- [x] Manual testing passed.


pnpm-workspace.yaml
```diff
@@ -5,13 +5,18 @@ packages:
...
 blockExoticSubdeps: true
+saveExact: true
 
-ignoredBuiltDependencies:
-  - 'bufferutil' # optional native ws speedup; ws falls back to pure JS
-  - 'core-js' # only a sponsor-banner postinstall, no functional payload
-  - 'keccak' # has a pure-JS fallback; the native addon is never loaded in the browser bundle
-  - 'utf-8-validate' # optional native ws speedup; ws falls back to pure JS
-  - 'workerd' # wrangler dev fetches the runtime it needs on its own; vite build never uses it
+allowBuilds:
```
gndelia, line 1: Can we keep this file alphabetically sorted? This should go to the top and save-exact below

joshuasing: Sure, fixed. As this file becomes larger, I think it could be worth grouping parts but I'll leave that up to you.

gndelia (dismissed):

portal-backend/api/package.json
```diff
@@ -38,6 +38,6 @@
   "engines": {
     "node": ">=22"
```
ArturDolzan, line 39: Perhaps it's worth a quick look at the `engines.node` next to these pins. pnpm 11.24.0 needs Node >=22.13 (it exits with an error below that), so `">=20"` on the crons and `">=22"` on the api may not hold anymore. The images are node:24 so nothing breaks now, but pnpm runs at container start here.

joshuasing: Agreed, updated all `engines.node` values to `>=24` - a few packages already had this, so all of the values now match.

ArturDolzan (approved):

gndelia (approved):

PR #2273, opened 2026-09-11
Restore the error pages heading size

### LINKED ISSUE #2266: 404 page heading is larger after the Vite migration
### Description

The 404 page text sizes changed after the Vite migration (production on the left, current on the right).

The portal used to have two different 404 pages that never matched each other:

- `portal/app/not-found.tsx` — the one Next actually served for unmatched routes. Title in a bare `<h3>`, which `globals.css` renders at 17px, and description in a `<p>` at 16px. Strings hardcoded in English.
- `portal/app/[locale]/not-found.tsx` — only reachable from inside the locale segment, e.g. Hemi Earn with the feature flag off. Title at 36px, description at 14px. Strings translated.

The migration dropped the root file and pointed the catch-all route at the locale one, so the second set of sizes is what everyone gets now.

One other difference: the 404 now renders inside the app layout, so the navbar is there. Before it rendered on its own.

Image

### Steps to reproduce

Open any URL that does not match a route, for example `/en/this-does-not-exist`.

### Proposed fix

Bring back the sizes the served page had, in `portal/app/[locale]/not-found.tsx`:

- title: `text-mid-md font-semibold` instead of `text-4xl font-medium`
- description: drop `text-sm`

The translated strings stay as they are — only the sizes go back. The icon, background and button already match.

ArturDolzan:
### Description

This PR puts the 404 title back to the size production served before the Vite migration. The migration deleted the root `not-found.tsx`, which is the file Next actually rendered for unmatched routes, and routed the catch-all to the locale one instead. Those two never agreed, so the title jumped from 17px to 36px.

The description keeps its size, it just moves from an `h3` back to a `p` so it stops inheriting the display face from the heading rule. The 500 page shares that markup, so it gets the same treatment and the two error pages still match each other.

Split into two commits since the 500 is alignment rather than the reported bug:
- `b924b180` - Restore the 404 heading size
- `d7408eb0` - Apply the same heading size to the 500 page

### Screenshots

Captura de Tela 2026-09-11 às 15 13 31

### Related issue(s)

Closes #2266

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/not-found.tsx
+++ b/portal/app/[locale]/not-found.tsx
@@ -18,12 +18,12 @@ export const NotFound = function () {
       <div className="z-10 m-auto flex flex-col items-center gap-4">
         <ExclamationMark />
         <div className="text-center">
-          <h1 className="text-4xl font-medium text-neutral-950">
+          <h1 className="text-mid-md font-semibold text-neutral-950">
             {t('not-found.title')}
           </h1>
-          <h3 className="mt-1 text-sm font-medium text-neutral-500">
+          <p className="mt-1 font-medium text-neutral-500">
             {t('not-found.description')}
-          </h3>
+          </p>
         </div>
         <Link
           className="button--base button-primary button-small button-regular"
--- a/portal/components/error500/index.tsx
+++ b/portal/components/error500/index.tsx
@@ -26,10 +26,8 @@ const Error500 = ({ description, reset, title, tryAgainLabel }: Props) => (
     <div className="m-auto flex flex-col items-center gap-4">
       <ExclamationMark />
       <div className="w-96 text-center max-md:max-w-[80%]">
-        <h1 className="text-4xl font-medium">{title}</h1>
-        <h3 className="mt-1 text-sm font-medium text-neutral-500">
-          {description}
-        </h3>
+        <h1 className="text-mid-md font-semibold">{title}</h1>
+        <p className="mt-1 font-medium text-neutral-500">{description}</p>
       </div>
       <button
         className="button--base button-primary button-small button-regular"


portal/components/error500/index.tsx
```diff
@@ -26,10 +26,8 @@ const Error500 = ({ description, reset, title, tryAgainLabel }: Props) => (
     <div className="m-auto flex flex-col items-center gap-4">
       <ExclamationMark />
       <div className="w-96 text-center max-md:max-w-[80%]">
-        <h1 className="text-4xl font-medium">{title}</h1>
-        <h3 className="mt-1 text-sm font-medium text-neutral-500">
-          {description}
-        </h3>
+        <h1 className="text-mid-md font-semibold">{title}</h1>
```
gndelia, line 29: I think this originally used `text-4xl`

https://github.com/hemilabs/ui-monorepo/blob/d2a844550c20f289db7ff4d52a77fa5c21e9a0fb/portal/components/error500/index.tsx#L45

Can we double check?

ArturDolzan: You're right, it was `text-4xl` and still is on main. I changed it on purpose: the 404 went back to the 17px it had before the migration, and since the 500 shares the same markup I moved it along so the two error pages keep matching.

Happy to drop that commit and leave the 500 alone if you'd rather keep this scoped to the 404.

gndelia (dismissed):

portal/app/[locale]/not-found.tsx
```diff
@@ -18,12 +18,12 @@ export const NotFound = function () {
       <div className="z-10 m-auto flex flex-col items-center gap-4">
         <ExclamationMark />
         <div className="text-center">
-          <h1 className="text-4xl font-medium text-neutral-950">
+          <h1 className="text-mid-md font-semibold text-neutral-950">
```
Comment (gabmontes) [inline portal/app/[locale]/not-found.tsx]: Nit: `text-neutral-950` is redundant here — `globals.css` already applies it to `h1`–`h4`.

— Gabi 🤖

Comment (ArturDolzan) [reply portal/app/[locale]/not-found.tsx]: Good catch, removed in [`40f256a4`](https://github.com/hemilabs/ui-monorepo/pull/2273/commits/40f256a4934ac9b04b5d05aac84eb26020257771). Nice side effect, the two `h1` now read exactly the same in both error pages.

gndelia (approved):

arielmergen (approved):

PR #2277, opened 2026-09-11
Show the HEMI price in the header

ArturDolzan:
### Description

This PR wires the price pill that landed in #2259 without a consumer. It goes left of the wallet button in the header, so the HEMI price reads on every page. On small screens the wallet moves to the bottom bar, so the pill sits alone on the right of the top bar instead.

It hides rather than showing a placeholder, on testnet (where the token list carries a stand-in token with no quote) and whenever the feed answers nothing. The header sits outside the app's error boundary, so the pill carries its own, and I checked a throw inside it still leaves every route rendering. The price code loads as its own chunk, so the header doesn't pay for it up front.

Split into four commits:
- `427e479c` - Show the HEMI price in the header
- `b30bf9c4` - Show the price pill on small screens too
- `053ab5d8` - Load the price fetch as its own chunk
- `eb7f7935` - Cap token prices at six decimals

Lint and tests pass.

### Screenshots
price-pill-six-digits

### Related issue(s)

Related to #2243

### Checklist

- [x] Manual testing passed.

### DIFF
--- a/portal/app/[locale]/_components/header/hemiPricePill.tsx
+++ b/portal/app/[locale]/_components/header/hemiPricePill.tsx
@@ -1,0 +1,56 @@
+import { ErrorBoundary } from 'components/errorBoundary'
+import { TokenPricePill } from 'components/tokenPricePill'
+import { useHemiToken } from 'hooks/useHemiToken'
+import { useNetworkType } from 'hooks/useNetworkType'
+import { useTokenPrices } from 'hooks/useTokenPrices'
+import { useUmami } from 'hooks/useUmami'
+import { Suspense } from 'react'
+import { getTokenPrice } from 'utils/token'
+
+const hemiPriceUrl = 'https://coinmarketcap.com/currencies/hemi/'
+
+const Price = function () {
+  const { data: prices, isLoading } = useTokenPrices()
+  const token = useHemiToken()
+  const { enabled, track } = useUmami()
+
+  // getTokenPrice answers '0' for a symbol the feed has no entry for, and that
+  // would render as $0.00 instead of hiding the pill
+  const quoted = getTokenPrice(token, prices)
+  const price = quoted === '0' ? undefined : quoted
+
+  if (price === undefined && !isLoading) {
+    return null
+  }
+
+  return (
+    <TokenPricePill
+      href={hemiPriceUrl}
+      isLoading={isLoading}
+      onClick={enabled ? () => track('header - hemi price') : undefined}
+      price={price}
+      token={token}
+    />
+  )
+}
+
+const HemiPricePillImpl = function () {
+  const [networkType] = useNetworkType()
+
+  if (networkType === 'testnet') {
+    return null
+  }
+
+  return <Price />
+}
+
+// The header renders outside the app's error boundary, so anything thrown here
+// blanks every route. The fragment is deliberate: ErrorBoundary falls back to
+// GenericError on a nullish fallback, and a header is no place for an error card
+export const HemiPricePill = () => (
+  <ErrorBoundary fallback={<></>}>
+    <Suspense fallback={null}>
+      <HemiPricePillImpl />
+    </Suspense>
+  </ErrorBoundary>
+)
--- a/portal/app/[locale]/_components/header/index.tsx
+++ b/portal/app/[locale]/_components/header/index.tsx
@@ -8,14 +8,13 @@ import { TunnelTabs } from 'components/tunnelTabs'
 
 import { Badge } from '../badge'
 
+import { HemiPricePill } from './hemiPricePill'
 import { HomeLink } from './homeLink'
 
-const WalletConnection = lazyWithFallback(
-  () =>
-    import('components/connectWallets').then(mod => ({
-      default: mod.WalletConnection,
-    })),
-  <div className="ml-auto" />,
+const WalletConnection = lazyWithFallback(() =>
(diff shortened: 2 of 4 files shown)


gndelia:
It's missing in the smallest view

image

Perhaps in this case it could go in the top nav bar, at the right, so the menu at the bottom is not crowded?

I'm also ok if you prefer to keep it hidden - I'm just highlighting it in case you missed this

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2277#issuecomment-5641359667):
> It's missing in the smallest view

Good catch. It was hidden on purpose to keep the small screen from getting crowded, but you're right that the top bar is the spot, its right side sits empty below 640px. Moved it there in [`b30bf9c4`](https://github.com/hemilabs/ui-monorepo/pull/2277/commits/b30bf9c4), the bottom bar stays untouched.

price-pill-mobile-topbar

thoughts?

portal/app/[locale]/_components/header/hemiPricePill.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const HemiPricePillImpl = function () {
+  const [networkType] = useNetworkType()
+
+  if (networkType === 'testnet') {
+    return null
+  }
+
+  return <Price />
```
Comment (gndelia) [inline portal/app/[locale]/_components/header/hemiPricePill.tsx]: It seems to me than then Price could be dynamically loaded

portal/app/[locale]/_components/header/hemiPricePill.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// The header renders outside the app's error boundary, so anything thrown here
+// blanks every route. The fragment is deliberate: ErrorBoundary falls back to
+// GenericError on a nullish fallback, and a header is no place for an error card
```
Comment (gndelia) [inline portal/app/[locale]/_components/header/hemiPricePill.tsx]: I think we can remove all of this

gndelia:
I think we should try to match the number of digits CMC uses (6)

image

image

If not, It's just too many digits (The price diff is probably caching on the portal backend)

portal/app/[locale]/_components/header/hemiPricePill.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <TokenPricePill
+      href={hemiPriceUrl}
+      isLoading={isLoading}
+      onClick={enabled ? () => track('header - hemi price') : undefined}
+      price={price}
+      token={token}
+    />
+  )
+}
```
Comment (gndelia) [inline portal/app/[locale]/_components/header/hemiPricePill.tsx]: `usePrices` returning `error` is not handled here

portal/app/[locale]/_components/header/hemiPricePill.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return <Price />
+}
+
+// The header renders outside the app's error boundary, so anything thrown here
+// blanks every route. The fragment is deliberate: ErrorBoundary falls back to
+// GenericError on a nullish fallback, and a header is no place for an error card
+export const HemiPricePill = () => (
+  <ErrorBoundary fallback={<></>}>
+    <Suspense fallback={null}>
```
Comment (gndelia) [inline portal/app/[locale]/_components/header/hemiPricePill.tsx]: I think this is redundant so it can be omitted

```suggestion
    <Suspense>
```

portal/app/[locale]/_components/header/hemiPricePill.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/_components/header/hemiPricePill.tsx]: Good idea, done in [`053ab5d8`](https://github.com/hemilabs/ui-monorepo/pull/2277/commits/053ab5d8).

portal/app/[locale]/_components/header/hemiPricePill.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/_components/header/hemiPricePill.tsx]: Removed in [`053ab5d8`](https://github.com/hemilabs/ui-monorepo/pull/2277/commits/053ab5d8). Flagging one thing so it doesn't bite later: the `<></>` fallback is load-bearing, ErrorBoundary does `fallback ?? <GenericError />`, so a `null` there drops a half-screen error card inside the header. Did you mean the boundary too, or just the comment?

portal/app/[locale]/_components/header/hemiPricePill.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/_components/header/hemiPricePill.tsx]: It was already covered, just not visibly: a failed query leaves the price undefined and the pill returns null. I checked with the endpoint answering 500 and it does disappear. Fair point that nothing said "error" though, so it goes through `isDataUnavailable` now in [`053ab5d8`](https://github.com/hemilabs/ui-monorepo/pull/2277/commits/053ab5d8), same as fiatBalance.

portal/app/[locale]/_components/header/hemiPricePill.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/_components/header/hemiPricePill.tsx]: You're right, dropped it in [`053ab5d8`](https://github.com/hemilabs/ui-monorepo/pull/2277/commits/053ab5d8).

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2277#issuecomment-5665037521):
> I think we should try to match the number of digits CMC uses (6)

Agreed, capped it at six in [`eb7f7935`](https://github.com/hemilabs/ui-monorepo/pull/2277/commits/eb7f7935). Reads `$0.006962` now.

gndelia (approved):

arielmergen (approved):

PR #2279, opened 2026-09-14
Add Truncate chain names

### LINKED ISSUE #2254: Transaction history table Chain names overlap in the From and To columns
### Description

In the tunnel transaction history, the chain names in the `From` and `To`
columns don't fit in their columns, so they run over the next one instead of
being cut off. The row becomes unreadable.

It shows up on Testnet, where the names are longer: "Hemi Sepolia" runs into the
`To` column and "Testnet Bitcoin" runs into `Tx Hash`.

Image

On Mainnet the names are shorter ("Hemi", "Bitcoin") so it mostly goes
unnoticed, but "Ethereum" already leaves almost no room.

### Steps to reproduce

Set the network type selector to Testnet and navigate to the Tunnel transaction
history, with at least one Bitcoin withdrawal in the list. The `From` and `To`
columns will overlap the columns next to them.

### Proposed fix

Names that don't fit should be cut off with the full name available on hover, or
the columns should be wide enough for the longest name in every language.

arielmergen:
### Description

The chain name overflowed the From and To columns instead of being cut off, so it ran over into the next column. It is now truncated, with the full name in a tooltip, and the columns fit the longest name: From and To go from 100px to
125px, and Status drops from 185 to 170 — six over the longest status string across the three locales. That also narrows the gap between Status and Action.

### Related issue(s)

Closes #2254  
Fixes #
Related to #

### Checklist

- [X] Manual testing passed.
- [X] Documentation updated, or N/A.
- [X] Environment variables set in CI, or N/A.


arielmergen:
Accurate, but no consumer of `Tooltip` passes a focusable trigger, and `focus` is in its default trigger. This belongs in the shared component, not in this cell. Leaving it as is.

gndelia:
This is not bad for the worst-case scenario, but it seems there's a lot of space between `status` and `action` column. Perhaps we need to tweak the columns' width?

arielmergen:
From and To go to 125px, which fits the longest chain name, and Status to 170,
so there's less width left to absorb.

arielmergen:
Good Catch, From and To go to 125px, which fits the longest chain name, and Status to 170, so there's less width left to absorb.

gndelia (approved):

gabmontes (approved):

Issue #2246, opened 2026-09-08
Add component to select lockup periods


gndelia:
We need a custom component that allows us to select Lockup periods

Screenshots 

<img width="608" height="187" alt="Image" src="https://github.com/user-attachments/assets/8fd46a50-9e0f-4672-aa10-d928b4947d0b" />

<img width="499" height="201" alt="Image" src="https://github.com/user-attachments/assets/62cb3299-d4b4-4286-85a7-24dbc1deff2d" />

<img width="1415" height="435" alt="Image" src="https://github.com/user-attachments/assets/e1aace92-a04d-4b46-b59b-a09272cbcdb0" />

Leaving the [private claude  artifact](https://claude.ai/code/artifact/1de649b4-3f13-4063-9fbd-4500cf803e4a?via=auto_preview)

ArturDolzan:
Done.

PR #2282, opened 2026-09-14
Replace the lockup slider with a preset row

ArturDolzan:
### Description

This PR turns the lockup selector into four presets. Picking a period meant dragging a slider, which is precise but not what most people want: the slider is still there, behind a Custom toggle.

Leaving Custom resets to the default preset, so the presets view never sits on a value none of them describe. The shortcut labels under the slider moved with it, they only show in Custom now. APRs under each label are supported by the component but nothing passes them yet, since `useCalculateApr` needs a `tokenId` from a position that does not exist while you are creating one.

### Screenshots

https://github.com/user-attachments/assets/809e83be-c702-40eb-b951-32472b626274

https://github.com/user-attachments/assets/744b420d-eee3-469b-8468-9f46f1654e6b

https://github.com/user-attachments/assets/ac478627-6b38-41cc-9891-ddca2e0aa82a

### Related issue(s)

Related to #2246

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
```diff
@@ -280,63 +308,91 @@ export function Lockup({
...
+                        ? 'cursor-default text-neutral-400'
+                        : 'cursor-pointer hover:text-neutral-950'
+                    }`}
+                    key={days}
+                    onClick={() => handleStepClick(days)}
+                  >
+                    {label}
+                  </span>
+                )
+              })}
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx:384]: These spans are unchanged from the base, they only show up in the diff because I re-indented them when the row moved into the Custom view (`git diff -w` comes back empty for that block). The keyboard gap is real but pre-existing. What was wrong is my description, which claimed they had been removed, fixed now.

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
```diff
@@ -267,6 +274,27 @@ export function Lockup({
...
 
+  const presets = [
+    { days: sixMonths, label: t('form.months', { months: 6 }) },
+    { days: oneYear, label: t('form.years', { years: 1 }) },
+    { days: twoYears, label: t('form.years', { years: 2 }) },
+    { days: maxDays, label: t('form.years', { years: 4 }) },
+  ]
+
+  const matchesPreset = presets.some(({ days }) => days === lockupDays)
+  const showDaysInput = showSlider || !matchesPreset
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/lockup/index.tsx:285]: `showDaysInput` derives from the current value, so the input unmounts while the user types in it. In the extend flow `minLocked` is a rounded 6-day value, so the input renders; `handleInputChange` writes `lockupDays` per keystroke without clamping to `minLocked`. Typing a preset below `minLocked` (e.g. `180` when `minLocked` is `516`) leaves an invalid value with the input gone, the CTA disabled, and `touched` blocking `updateInputDaysByMinLocked` — only the Custom toggle recovers it.

Better to drive visibility from explicit state rather than from whether the value happens to match a preset.

— Auto Gabi

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
```diff
@@ -267,6 +274,27 @@ export function Lockup({
...
+
+  const defaultPreset =
+    presets.find(({ days }) => days >= Math.max(twoYears, minLocked ?? 0))
+      ?.days ?? maxDays
+
+  function handleCustomClick() {
+    if (showSlider) {
+      handleStepClick(defaultPreset)
+    }
+    setShowSlider(!showSlider)
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/lockup/index.tsx:295]: Resetting unconditionally discards a value a preset already describes: slide to exactly 1 year, toggle Custom off, selection jumps to 2 years. Gating on `!matchesPreset` keeps the invariant and the user's choice.

— Gabi 🤖

portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx
```diff
@@ -0,0 +1,64 @@
...
+          disabled={disabled}
+          key={days}
+          onClick={() => onSelect(days)}
+          type="button"
+        >
+          <span className="body-text-semibold">{label}</span>
+          {sublabel !== undefined && (
+            <span
+              className={`text-xs font-medium ${
+                selected && !disabled ? '' : disabled ? '' : 'text-neutral-500'
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx:54]: Both branches of the inner ternary return `''`, so this collapses to `selected || disabled ? '' : 'text-neutral-500'`.

— Auto Gabi

portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx
```diff
@@ -0,0 +1,64 @@
...
+  labelledBy,
+  minLocked,
+  onSelect,
+  options,
+  value,
+}: Props) => (
+  <div
+    aria-labelledby={labelledBy}
+    className="grid w-full grid-cols-2 rounded-lg bg-white p-1 shadow-lockup-input-default xs:grid-cols-4"
+    role="group"
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx:25]: Minor: single-choice set announced as four independent toggles, so no "1 of 4" and no arrow-key navigation. `role="radiogroup"` + `role="radio"`/`aria-checked` fits better.

— Gabi 🤖

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
```diff
@@ -21,9 +22,13 @@ import {
 } from '../../_utils/lockCreationTimes'
 import { sanitizeLockup } from '../../_utils/sanitizeLockup'
 
+import { LockupPresets } from './lockupPresets'
 import { RangeSlider } from './rangeSlider'
 import { WarningMessage } from './warningMessage'
 
+const oneYear = epochsPerYear * step
+const sixMonths = Math.floor(epochsPerYear / 2) * step
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/_components/lockup/index.tsx:30]: Do these belong to `_utils/lockCreationTimes.ts` ?

gndelia:
MAybe it's me

image


but it looks _Very_ tall. Should we reduce the padding a bit, to make it more consistent with the rest of the app?

gndelia:
When clicking "Custom", entering a custom value, and then closing the Custom lockup section, it always selects "2 years"

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: You're right, gating it keeps both. Sliding to exactly 1 year and toggling off stays on 1 year now, and anything else snaps to the closest preset instead of a fixed default. [`fed791e9`](https://github.com/hemilabs/ui-monorepo/pull/2282/commits/fed791e9)

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: Good catch, and the extend flow was the sharp edge. Visibility is plain state now, seeded from whether the opening value is a selectable preset, so typing can't unmount the field under you. Fixing that also made the extend drawer open on the presets instead of hiding them behind Custom. [`fed791e9`](https://github.com/hemilabs/ui-monorepo/pull/2282/commits/fed791e9)

portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx]: Right, collapsed in [`fed791e9`](https://github.com/hemilabs/ui-monorepo/pull/2282/commits/fed791e9).

portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx]: Agreed on the semantics. I'd hold off here because half a radiogroup is worse than buttons, the role promises arrow key navigation with a roving tabindex, and without that a screen reader announces "1 of 4" while the arrows do nothing. There's a related gap you might have a view on: `disabled` drops the presets below the floor out of the tab order entirely, so `aria-disabled` plus a no-op handler would keep them discoverable. Happy to open an issue and do the whole pattern in one go.

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: They do, moved them over in [`fed791e9`](https://github.com/hemilabs/ui-monorepo/pull/2282/commits/fed791e9). They keep the derivation from `epochsPerYear`, which reads better sitting next to it anyway.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2282#issuecomment-5673792506):
> When clicking "Custom", entering a custom value, and then closing the Custom lockup section, it always selects "2 years"

Fixed in [`fed791e9`](https://github.com/hemilabs/ui-monorepo/pull/2282/commits/fed791e9), it only resets when the value isn't one of the presets, and then it picks the closest one. Picking 12 days and closing Custom lands on 6 months now, not 2 years.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2282#issuecomment-5673786479):
> but it looks _Very_ tall. Should we reduce the padding a bit, to make it more consistent with the rest of the app?

Not just you. The row was holding 48px so the APR line could land later without shifting the form. It follows the labels now, 36px with one line and back to 48 when the APR shows up. The block around it reserves a fixed height instead, so toggling Custom doesn't move anything, on mobile either. [`fed791e9`](https://github.com/hemilabs/ui-monorepo/pull/2282/commits/fed791e9). Screenshots/videos updated

gndelia (approved):

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
```diff
@@ -270,72 +304,109 @@ export function Lockup({
   useEffect(
     function updateInputDaysByMinLocked() {
       if (minLocked && Number(inputDays) < minLocked && !touched) {
-        updateInputDays(String(minLocked))
-        updateLockupDays(minLocked)
+        updateInputDays(String(openingDays))
+        updateLockupDays(openingDays)
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: `actionCell` seeds `lockupDays` at `minDays`, so this effect fires on open and writes `openingDays` — the smallest preset `>= minLocked` — where it used to write `minLocked`. A position with ~1 year left opens preselected at 2 years; anything over 732 days left opens at 4 years, the max, and the only enabled preset, with submit already enabled. Intended?

— Gabi 🤖

portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export const LockupPresets = function ({
+  labelledBy,
+  minLocked,
+  onSelect,
+  options,
+  value,
+}: Props) {
+  const hasSublabels = options.some(({ sublabel }) => sublabel !== undefined)
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx:20]: `sublabel` and the `h-12`/`h-9` branch it drives have no caller outside the story, since the APR needs a `tokenId` that doesn't exist during creation. Worth dropping both and bringing them back with the APR work.

— Auto Gabi

portal/app/[locale]/hemi-stake/_utils/lockCreationTimes.ts
```diff
@@ -16,6 +16,29 @@ export const twoYears = 732
...
+
+export const getNearestPreset = function ({
+  days,
+  minLocked,
+  presets,
+}: GetNearestPresetProps) {
+  const selectable = presets.filter(preset => !minLocked || preset >= minLocked)
+
+  return selectable.length === 0
+    ? maxDays
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_utils/lockCreationTimes.ts]: Nit: `maxDays` is itself a preset, so this branch runs only when `minLocked > maxDays`, and then `maxDays` is below the floor the filter just applied — an invalid lock. The caller writes it into `lockupDays` without `handleStepClick`'s guard. Unreachable today (`minLocked` is capped at `maxDays`), but the test below pins it as the expected answer. Clamping to `minLocked`, or returning nothing, would read truer.

— Gabi 🤖

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: Intended, yes. Opening at the floor proposed an extension of zero days, so it now opens on the shortest preset that clears it. Small correction though: submit was already enabled before, the floor is a valid value, so what changed is what it submits and not the button. If the 4 year case feels too aggressive, happy to open an issue and rethink the default.

portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/lockupPresets.tsx:20]: Fair point on the tokenId, I hadn't framed it that way. I'd lean toward keeping it, the sublabel and the height switch went in on purpose so the APR line can land without reworking the row. If you feel strongly, happy to open an issue so we can decide it alongside the APR work.

portal/app/[locale]/hemi-stake/_utils/lockCreationTimes.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_utils/lockCreationTimes.ts]: Went with the clamp, and the test now expects the floor instead of the max. The bit about the effect writing it without handleStepClick's guard is what sold it, I'd only looked at the click path. [`193fcddc`](https://github.com/hemilabs/ui-monorepo/pull/2282/commits/193fcddc)

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
Comment (gabmontes) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: @ArturDolzan discuss this behavior with @matt-lam and @max-sanchez. Let's them decide the proper UX here.

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: Just a correction to what I said above: the drawer never opened on a zero day extension. `minLocked` is already the first valid date past your current unlock, so it proposed the smallest possible extension, you just had to confirm. That makes your point stronger, not weaker.

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: Matt landed on dropping the presets when extending, so it goes back to opening on the Custom slider, in 7f63aed2.

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
```diff
@@ -202,8 +206,25 @@ export function Lockup({
...
+  const presets = [
+    { days: sixMonths, label: t('form.months', { months: 6 }) },
+    { days: oneYear, label: t('form.years', { years: 1 }) },
+    { days: twoYears, label: t('form.years', { years: 2 }) },
+    { days: maxDays, label: t('form.years', { years: 4 }) },
+  ]
+
+  const presetDays = presets.map(preset => preset.days)
+
+  const canUsePresets = minLocked === undefined
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/lockup/index.tsx:218]: Nit: this reads `minLocked === undefined` as "creating a lock", but the extend drawer passes `nearest ? nearest.maxValue! : undefined` and `maxValue` is `null` for a position at or below `minDays`. So "no floor" has two encodings here, and only one of them should select presets — today `null !== undefined` keeps the slider by luck, and whoever later fixes that `!` to `?? undefined` flips the extend drawer to presets. An explicit prop for the mode would decouple the two.

— Auto Gabi

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx:218]: Good catch on the two encodings. I fixed it at the source rather than adding a mode prop: the drawer now passes `nearest?.maxValue ?? minDays`, so minLocked is always a number when extending, and the `!` that was smuggling null through a `number | undefined` prop is gone. That also gives a position with 12 days or less a real floor, it was getting null and skipping every `minLocked &&` guard inside Lockup. [`e1bbd472`](https://github.com/hemilabs/ui-monorepo/pull/2282/commits/e1bbd472)

gabmontes (approved):

Issue #2286, opened 2026-09-15
Track Hemi supply historically


gndelia:
Track the Hemi supply historically, aggregating staked + circulating + non-circulating so we can map them on a stacked bar chart

 Additionally, use this to return current `/circulating` endpoint from the same source

gndelia:
This was completed - closing

PR #2289, opened 2026-09-15
Improve realtime indexing by skipping old headers

gndelia:
### Description

Follow-up to #2288 - the "Realtime" handler was indexing from `start_block` just like the historic data handler. By configuring this through an environment variable, I hope to skip _lots_ of block handlers, making the sync process faster. When running locally, everything works well, but on Envio, I'm seeing some performance issues. Taken from [Different Historical and Realtime Intervals recipe](https://docs.envio.dev/docs/HyperIndex/block-handlers#different-historical-and-realtime-intervals)

Variables were already configured on Envio, with blocks from 5 days ago

### Screenshots

N/A

### Related issue(s)

Related to https://github.com/hemilabs/ui-monorepo/issues/2286

### Checklist

- [x] Manual testing passed.


subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -340,7 +346,14 @@ indexer.onBlock(
   {
     name: 'supply-realtime',
     where: ({ chain }) => ({
-      block: { number: { _every: realtimeStride[chain.id] } },
+      block: {
+        number: {
+          _every: realtimeStride[chain.id],
+          ...(realtimeStartBlock[chain.id] && {
+            _gte: Number(realtimeStartBlock[chain.id]),
```
ArturDolzan, line 353: Might be worth a quick look at what happens if one of the new start block vars ends up non numeric. `Number()` would give NaN there, and I wonder if the realtime handler would then just stop firing without any log.

gndelia: I don't know, but vars are already configured so not happening

ArturDolzan (approved):

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
gndelia: I checked and we will get an error, so the logs would tell us to fix it

PR #2293, opened 2026-09-15
Add RPC fallbacks and logs to supply snapshots

gndelia:
### Description

Follow-up to #2289 and (apparently) the last part of the Envio changes.

This PR

- adds logging so we can check the logs for troubleshooting on Envio
- Adds several RPCs per chain, with `viem` fallback 

### Screenshots

With this, and setting several RPCs in Envio, indexing seems to finally sync completely without errors

image

### Related issue(s)

Related to #2286

### Checklist

- [x] Manual testing passed.


subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -86,25 +96,82 @@ export const endOfDay = (date: string) =>
 // Multicall3 lives at the same address on every chain
 const multicallAddress = '0xcA11bde05977b3631167028862bE2a173976CA11'
 
-const rpcUrls: Record<Chain['id'], string | undefined> = {
-  [bsc.id]: process.env.ENVIO_RPC_URL_BNB,
-  [hemi.id]: process.env.ENVIO_RPC_URL_HEMI,
-  [mainnet.id]: process.env.ENVIO_RPC_URL_ETH,
+export const toRpcUrls = (value = '') =>
+  value.split('+').filter(url => URL.canParse(url))
```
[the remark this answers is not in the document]
gndelia: RPCs will be configured correctly. Don't worry.

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -86,25 +96,82 @@ export const endOfDay = (date: string) =>
...
+            `Chain ${chainId} RPC ${new URL(url).host} answered ${response.status} ${response.statusText}`,
+          )
+        }
+      },
+      retryCount: 0,
+      timeout: 30000,
+    }),
+  )
+  return transports.length > 1
+    ? fallback(transports, { retryCount: 0 })
```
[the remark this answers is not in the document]
gndelia: I tested it locally and it's not the case

subgraphs/hemi-earn-requests-subgraph/README.md
```diff
@@ -51,7 +51,7 @@ This indexer also takes the HEMI supply snapshots that the portal-backend serves
 
 These optional env variables configure the snapshots:
 
-- `ENVIO_RPC_URL_BNB`, `ENVIO_RPC_URL_ETH`, `ENVIO_RPC_URL_HEMI` — the RPC URL used to read the balances on each chain. When unset, the public RPC of the chain is used.
+- `ENVIO_RPC_URL_BNB`, `ENVIO_RPC_URL_ETH`, `ENVIO_RPC_URL_HEMI` — the RPC URLs used to read the balances on each chain. Join several URLs with `+` (for example `https://a.example+https://b.example`) use them as viem fallbacks. The public RPCs can reject the historical reads, so set an archive URL on each chain.
```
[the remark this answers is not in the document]
gndelia: Fixed in [b26231a2](https://github.com/hemilabs/ui-monorepo/pull/2293/commits/b26231a2a8831bfaa48b7d42d01f1f722586fccd).

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -86,25 +96,82 @@ export const endOfDay = (date: string) =>
...
+        }
+      },
+      onFetchResponse(response) {
+        if (!response.ok) {
+          console.warn(
+            `Chain ${chainId} RPC ${new URL(url).host} answered ${response.status} ${response.statusText}`,
+          )
+        }
+      },
+      retryCount: 0,
```
gabmontes: `retryCount: 0` disables viem's default 3 retries with backoff (which cover 429/408/5xx). That makes sense when there's a fallback list to move on to, but lines 110 and 132 mean the unset (public RPC) and single-URL cases now have no transport-level retry at all, and the outer `withRetry` also went from 3 backoff retries to 1 after a flat 30s wait. Those are the endpoints most likely to throttle — worth keeping viem's retries when `transports.length === 1`.

Gabi 🤖

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -86,25 +96,82 @@ export const endOfDay = (date: string) =>
...
+  fn: () => Promise<T>,
+) {
+  const start = Date.now()
+  log.info(`${label} started`)
+  try {
+    const result = await fn()
+    log.info(`${label} finished`, { ms: Date.now() - start })
+    return result
+  } catch (error) {
+    log.info(`${label} failed`, {
```
gabmontes: Failures are logged at `info`, so they can't be filtered by level in the Envio logs — which is what this PR is adding logging for. `log.warn` (or `log.error`) seems a better fit here.

Auto Gabi

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -86,25 +96,82 @@ export const endOfDay = (date: string) =>
...
+const toTransport = function (chainId: Chain['id'], urls: string[]) {
+  if (urls.length === 0) {
+    return http(undefined, { batch: true, retryCount: 0, timeout: 30000 })
+  }
+  const transports = urls.map((url, index) =>
+    http(url, {
+      batch: true,
+      onFetchRequest() {
+        if (index > 0) {
+          console.info(`Chain ${chainId} falls back to ${new URL(url).host}`)
```
gabmontes: These transport-level logs are the most valuable for troubleshooting RPC issues, but `console.*` bypasses envio's `Logger`, which is the one documented as showing up in the Envio Hosted Service (and it forces the file-wide `no-console` disable). Did you confirm raw stdout shows up there too?

Gabi 🤖

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -115,6 +182,12 @@ const llamaChainByChain: Record<Chain['id'], string> = {
 // Shared by the chains. The public RPCs reject faster reads.
 const rateLimit = { calls: 2, per: 'second' } as const
 
+const dayEndRateLimit: Record<Chain['id'], RateLimit> = {
```
gabmontes, line 191: A short comment on why BNB needs 1 call per 3s (unlike the shared `rateLimit` right above, which does explain itself) would help the next reader. Also, the realtime `readSnapshot` effect still shares a single 2/s limit across all three chains while the historical one is now per chain — intentional?

Auto Gabi

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
gndelia: Fixed in [0fb3a6d5](https://github.com/hemilabs/ui-monorepo/pull/2293/commits/0fb3a6d5385f59ae8a2d5e18df822d5b777e981a). Only a transport inside a fallback list keeps `retryCount: 0`; the public RPC and a lone URL now use viem's retries.

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
gndelia: Fixed in [582ab714](https://github.com/hemilabs/ui-monorepo/pull/2293/commits/582ab714e2c00d1fa2473276dfd8421cdd7919b5). A failed attempt is now `log.warn`, so it is filterable even when the retry then succeeds.

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
gndelia: Good catch, and no — I did not confirm that raw stdout reaches the Hosted Service. envio logs through pino with the ECS strategies (`src/Logging.res.mjs`), so a `console.*` line is unstructured next to those JSON records and carries no level.

Fixed in [3b80d6f1](https://github.com/hemilabs/ui-monorepo/pull/2293/commits/3b80d6f1510c65c126304574fd9732df80c8b69b): each read builds its client, so `toTransport` gets `context.log` and both callbacks use the envio `Logger`. The file-wide `no-console` disable is gone. Building the client per read does not break batching: viem keys its batch scheduler by URL in a module-level cache, and measured on 2.28.4 the concurrent `getBlock` and `multicall` still leave as one POST with two calls.

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
gndelia: Comment restored in [7d44b417](https://github.com/hemilabs/ui-monorepo/pull/2293/commits/7d44b4173eec80a8283cc15a5ec0d5e4293d1a48).

On the realtime effect: yes, sharing one limit is deliberate. `readSnapshot` runs once per stride, about every 5 minutes per chain, so the three chains together make roughly 3 reads per 5 minutes against the 2-per-second cap. The day-end effect is the one that runs back to back during the historical sync, so it is the only one that needed a per-chain limit.

gabmontes (dismissed):

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -86,25 +93,92 @@ export const endOfDay = (date: string) =>
...
+const toTransport = function (
+  chainId: Chain['id'],
+  urls: string[],
+  log: Logger,
+) {
+  if (urls.length === 0) {
+    return http(undefined, { batch: true, timeout: 30000 })
+  }
+  // disallow retry for urls with fallbacks
+  const retryCount = urls.length > 1 ? 0 : undefined
```
ArturDolzan: Not sure, but maybe a single `hasFallbacks` const would read a bit clearer here? `urls.length > 1` and `transports.length > 1` are the same check, and naming it ties the `retryCount: 0` to the fallback list.

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
```diff
@@ -86,25 +93,92 @@ export const endOfDay = (date: string) =>
...
+) {
+  const start = Date.now()
+  log.info(`${label} started`)
+  try {
+    const result = await fn()
+    log.info(`${label} finished`, { ms: Date.now() - start })
+    return result
+  } catch (error) {
+    log.warn(`${label} failed`, {
+      error: (error as Error).message,
```
ArturDolzan, line 164: Perhaps it's worth checking what `error.message` holds here. viem puts `URL: <full url>` in its `HttpRequestError`/`TimeoutError` messages, so a paid endpoint's API key could end up in the Envio logs. Maybe `shortMessage` + `details` would be enough?

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
gndelia: Fixed in [57a3fcdd](https://github.com/hemilabs/ui-monorepo/pull/2293/commits/57a3fcdd). A `hasFallbacks` const now drives both the `retryCount` and the `fallback` wrapper, which also made the comment above it unnecessary.

subgraphs/hemi-earn-requests-subgraph/src/mappings/supplyHandlers.ts
gndelia: True, but we don't have secret keys now and these logs are not public - only seen by us (admins) on Envio. I'd rather keep it as is instead of complicating the error handling.

ArturDolzan (approved):

gabmontes (approved):

PR #2297, opened 2026-09-16
Add the HEMI supply chart to the Analytics tab

ArturDolzan:
### Description

This PR fills the Analytics tab with the HEMI supply over time: four cards and a stacked bar chart splitting it into circulating, staked and the rest, over a week, a month or three, priced in HEMI or in dollars at each day's rate.

The endpoint from #2286 isn't live yet, so the fetcher returns a snapshot of what it will answer. It sits behind `_fetchers/fetchHemiSupplyHistory.ts` so swapping it for the real call touches one file.

### Screenshots

https://github.com/user-attachments/assets/05870491-76d1-4fb8-aed5-000bcb2fd57b

https://github.com/user-attachments/assets/4e4ddd85-d022-465b-a475-0f8d12f20b75

Error state:

analytics-error

### Related issue(s)

Related to #2286

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-stake/analytics/_components/changeIndicator.tsx
```diff
@@ -0,0 +1,16 @@
...
+type Props = {
+  children: string
+  isUp: boolean
+}
+
+export const ChangeIndicator = ({ children, isUp }: Props) => (
+  <span className="flex items-center gap-x-0.5">
+    <Arrow
+      aria-hidden
+      className={`h-2.5 w-2 [&>path]:fill-neutral-600 ${isUp ? '' : 'rotate-180'}`}
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_components/changeIndicator.tsx]: Good catch, that was a regression from swapping the sign for the arrow. There's an sr-only span next to it now carrying the direction, so a screen reader hears "decreased by $16.24M". [`57974fe9`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/57974fe9)

portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx
```diff
@@ -0,0 +1,308 @@
...
+                    locale={locale}
+                    rows={rows}
+                    totalLabel={t('total')}
+                  />
+                }
+                pointerLength={0}
+              />
+            }
+            labels={() => ' '}
+            voronoiDimension="x"
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx:247]: Fair, and it's a real gap. I'd rather not fix it only here though: the Hemi Earn historical chart has the same tooltip-only pattern, so solving it in this one chart leaves the other half standing and gives us two different chart interactions to maintain. It's worth doing once, for both.

portal/app/[locale]/hemi-stake/analytics/_fetchers/fetchHemiSupplyHistory.ts
```diff
@@ -0,0 +1,10 @@
+import { type SupplyPoint } from '../_utils/supplyHistory'
+
+// TODO - drop this file when the endpoint tracked in
+// https://github.com/hemilabs/ui-monorepo/issues/2286 is available
+import sample from './hemiSupplyHistory.json'
+
+// TODO read the history from the endpoint tracked in
+// https://github.com/hemilabs/ui-monorepo/issues/2286. Until it ships, this is
+// a snapshot of what that endpoint returns, not generated data.
+export const fetchHemiSupplyHistory = async () => sample as SupplyPoint[]
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_fetchers/fetchHemiSupplyHistory.ts]: True as it stands, but this JSON is on its way out: the endpoint from #2286 should land before this merges, and then the fetcher does a real request instead of importing anything. Code splitting a file that's about to be deleted would be work with an expiry date.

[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_fetchers/fetchHemiSupplyHistory.ts]: Same reason as the bundling one. The snapshot is temporary and the real endpoint is chain aware, so the fetcher gets what it needs once it starts making a request. The query key already carries the chainId.

portal/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.ts
```diff
@@ -0,0 +1,162 @@
...
+  Number(formatUnits(BigInt(wei), decimals))
+
+// Throws on malformed values on purpose: it runs where the query can turn it
+// into an error state, rather than during render.
+export const parseSupplyPoints = (points: SupplyPoint[], decimals: number) =>
+  points.map(point => ({
+    circulating: toTokens(point.circulating, decimals),
+    nonCirculating: toTokens(point.nonCirculating, decimals),
+    priceUsd: Number(point.priceUsd),
+    staked: toTokens(point.staked, decimals),
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.ts]: You're right, and the parser was claiming otherwise in a comment right above it. Every conversion goes through a finite check now, amounts as well as the price, with tests for "not-a-number" and "Infinity". [`57974fe9`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/57974fe9)

gndelia (commented):
Image

vertical padding is odd, there's more space at the bottom than at the top.

portal/messages/en.json
```diff
@@ -452,6 +452,19 @@
...
+      "decreased-by": "decreased by",
+      "hemi-price": "HEMI price",
+      "increased-by": "increased by",
+      "month": "{count, plural, one {# month} other {# months}}",
+      "non-circulating": "Non-circulating",
+      "share-of-supply": "{share} of supply",
+      "staked": "Staked",
+      "title": "HEMI supply",
+      "total": "Total"
+    },
```
gndelia, line 467: HEMI should probably be read from the token (passed as symbol) instead of hardcoded here

gndelia (commented):
are the loading and error states in the chart?
loading should be a few skeleton bars (ideally, if they can be vertical bars). For errors, you can check this pattern of showing a refresh button if the query failed

https://github.com/vetro-protocol/vetro-monorepo/blob/0aacd370c9dbf36b0c40f557a99899445645f1d9/web/src/components/apyHistoryCard.tsx#L94

(It could also be combined with an ErrorBoundary for unexpected errors)

portal/test/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.test.ts
```diff
@@ -0,0 +1,154 @@
+import {
+  getSupplySummary,
+  parseSupplyPoints,
+  sliceByPeriod,
+  toChartSeries,
+} from 'app/[locale]/hemi-stake/analytics/_utils/supplyHistory'
+import { describe, expect, it } from 'vitest'
+
+const wei = (amount: number) => (BigInt(amount) * BigInt(1e18)).toString()
```
Comment (gndelia) [inline portal/test/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.test.ts]: FYI you can use [parseUnits](https://viem.sh/docs/utilities/parseUnits) from viem that does this

portal/test/utils/sitemapRoutes.test.ts
```diff
@@ -7,7 +7,8 @@ import { describe, expect, it } from 'vitest'
 // with `!baseRoute.endsWith('demos')`, which is how `[shareAddress]` slipped in.
 const excluded: Record<string, string> = {
   '/hemi-earn/pool/[shareAddress]': 'dynamic, one URL per pool address',
-  '/hemi-stake/analytics': 'placeholder until the analytics charts land',
+  '/hemi-stake/analytics':
+    'serving a bundled snapshot until the supply endpoint ships',
```
gndelia: sorry to ask, but why was this excluded? It will land prod if the feature toggle is enabled (which will be) once `hemi-staking` branch is merged

portal/test/app/[locale]/hemi-stake/analytics/_utils/formatSupplyValue.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    expect(
+      formatSupplyValue({ locale, symbol, unit: 'usd', value: 1_500_000 }),
+    ).toMatch(/^\$/)
+  })
+
+  it('should keep the sign ahead of the currency symbol', function () {
+    expect(
+      formatSupplyValue({ locale, symbol, unit: 'usd', value: -1_500_000 }),
+    ).toMatch(/^-\$/)
+  })
```
Comment (gndelia) [inline portal/test/app/[locale]/hemi-stake/analytics/_utils/formatSupplyValue.test.ts:36]: perhaps we could be a bit more strict with these tests, and assert the exact values. Are we using `contains` and regex for any particular reason?

portal/app/[locale]/hemi-stake/analytics/_hooks/useHemiSupplyHistory.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  return useQuery({
+    queryFn: async () =>
+      parseSupplyPoints(await fetchHemiSupplyHistory(), decimals),
+    queryKey: getHemiSupplyHistoryQueryKey(chainId),
+    refetchInterval: 1000 * 60 * 5,
+    select,
+    staleTime: 1000 * 60 * 5,
+  })
+}
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/analytics/_hooks/useHemiSupplyHistory.ts]: I know `decimals` is static, so it won't change, but any automatic tool or people reviewing may think that `decimals` should belong to the key. However, I don't like that approach either; it would be annoying if we need to consume the key elsewhere. Perhaps we can add a small comment explaining why `decimals` is not part of the key even though used inside `queryFn`

portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx
```diff
@@ -0,0 +1,308 @@
...
+  unit,
+}: Props) {
+  const locale = useLocale()
+  const t = useTranslations('hemi-stake.analytics')
+  const { width: windowWidth } = useWindowSize()
+  const chartWidth =
+    widthByBreakpoint.find(([minWidth]) => windowWidth >= minWidth)?.[1] ??
+    fallbackChartWidth
+
+  const formatValue = (value: number) =>
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx:204]: `formatValue` is the axis formatter (compact, 2 fraction digits) and the tooltip reuses it. On the bundled snapshot in HEMI mode, 165/179 consecutive days render an identical circulating value, 168/179 for non-circulating, and Total reads "10B" every single day — hovering says almost nothing. USD mode is fine because the price moves. `historicalMetricsChart` already splits this with `precision: 'full'`; the same split would work here.

— Gabi 🤖

portal/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+  return parsed
+}
+
+const toTokens = (wei: string, decimals: number) =>
+  toFiniteNumber(formatUnits(BigInt(wei), decimals), 'amount')
+
+// Throws on malformed values on purpose: it runs where the query can turn it
+// into an error state, rather than during render.
+export const parseSupplyPoints = (points: SupplyPoint[], decimals: number) =>
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.ts:86]: `sliceByPeriod` treats `at(-1)` as the newest point and `getSupplySummary` treats `at(0)`/`at(-1)` as the range ends, but nothing sorts. The snapshot is ascending and #2286 doesn't pin an order, so a newest-first response would invert the window and every change badge silently instead of failing. Sorting by timestamp here makes the rest of the pipeline order-independent.

— Auto Gabi

portal/app/[locale]/hemi-stake/analytics/_components/supplyChartCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+          <span className="body-text-medium text-neutral-500">
+            {t('title')}
+          </span>
+          <div className="flex items-center gap-3">
+            <div className="hidden items-center gap-2 lg:flex">
+              {periodControls('')}
+            </div>
+            <div className="hidden h-3 w-0.5 bg-neutral-200 lg:block" />
+            <div className="flex items-center gap-2">
+              <SegmentedControlItem
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/analytics/_components/supplyChartCard.tsx:70]: (anchored here since the rename itself has no diff line) `SegmentedControlItem` conveys selection by background colour only, so assistive tech reads these as identical buttons with no pressed state. `aria-pressed={selected}` on the button in `components/segmentedControlItem.tsx` closes it, and now that the component is shared it covers Earn too.

— Gabi 🤖

portal/messages/en.json
ArturDolzan: Good call, it reads from the token symbol now in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5). Both the chart title and the price card take it from `useHemiToken`.

portal/test/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.test.ts
Comment (ArturDolzan) [reply portal/test/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.test.ts]: Thanks, swapped the local helper for `parseUnits` in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5).

portal/test/utils/sitemapRoutes.test.ts
ArturDolzan: Fair question. That exclusion dates back to when the page was an empty placeholder. Now that the charts are here it belongs in the sitemap like `/hemi-stake` already does, so I added the route and dropped the exclusion in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5).

portal/test/app/[locale]/hemi-stake/analytics/_utils/formatSupplyValue.test.ts
Comment (ArturDolzan) [reply portal/test/app/[locale]/hemi-stake/analytics/_utils/formatSupplyValue.test.ts:36]: No good reason, just leftovers from when I was sketching the formatter. They assert exact strings now, plus two cases that were missing, in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5).

portal/app/[locale]/hemi-stake/analytics/_hooks/useHemiSupplyHistory.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_hooks/useHemiSupplyHistory.ts]: Done in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5), short comment explaining why it stays out of the key.

portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx:204]: Good catch, the tooltip formats at full precision now in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5). I tried more decimals on the compact form first and it wasn't enough, non-circulating still repeated 154 of 179 days at 5 places and Total never moved at all.

portal/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.ts:86]: Agreed, `parseSupplyPoints` sorts by timestamp now so the rest of the pipeline stops depending on the response order. In [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5).

portal/app/[locale]/hemi-stake/analytics/_components/supplyChartCard.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_components/supplyChartCard.tsx:70]: Good one, added in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5). It covers Earn too now that the component is shared.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2297#pullrequestreview-5227053226):
> vertical padding is odd, there's more space at the bottom than at the top.

You were right, roughly 7 above the rows and 14 below. Measured both in the browser and it's 6.96 and 6.79 now, in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5).

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2297#pullrequestreview-5227182613):
> are the loading and error states in the chart?

Done in [`794cd7a5`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/794cd7a5).

portal/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.ts
```diff
@@ -0,0 +1,193 @@
...
+const oneDayMs = 24 * 60 * 60 * 1000
+
+export const getPeriodDurationMs = (period: SupplyPeriod) =>
+  (daysPerPeriod[period] - 1) * oneDayMs
+
+// The endpoint dates a point by the day it covers, but a full ISO timestamp
+// would parse just as well, so only the date half is read.
+const toTimestamp = function (date: string) {
+  const timestamp = new Date(`${date.slice(0, 10)}T00:00:00Z`).getTime()
+  if (Number.isNaN(timestamp)) {
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_utils/supplyHistory.ts]: Good catch, and it's worse than the example: `2026-04-31` also slid to May 1. Added a round trip check against the original `YYYY-MM-DD` plus tests in [`4b3cdd63`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/4b3cdd63).

portal/app/[locale]/hemi-stake/analytics/_utils/formatSupplyValue.ts
```diff
@@ -0,0 +1,37 @@
...
+  unit: SupplyUnit
+  value: number
+}) {
+  const sign = value < 0 ? '-' : ''
+  const amount = Math.abs(value)
+
+  if (precision === 'full') {
+    return unit === 'usd'
+      ? `${sign}$${formatFiatNumber(amount)}`
+      : `${sign}${formatNumber(amount)} ${symbol}`
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_utils/formatSupplyValue.ts:30]: Worth flagging, though one correction: `smart-round` defaults to `en-US`, not the browser locale, so it's deterministic rather than environment dependent. Keeping it as is for now, `formatNumber` and `formatFiatNumber` don't take a locale anywhere in the portal, so fixing it only here would leave this one chart formatting differently from the rest of the app.

portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx
```diff
@@ -0,0 +1,340 @@
...
+            >
+              <Skeleton height="100%" />
+            </div>
+          ))}
+        </div>
+      </div>
+    )
+  }
+
+  return (
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx:325]: This `return` is the catch-all for every non-pending state, so it also covers two cases that aren't errors. USD with a missing price is the notable one: `toChartSeries` returns `undefined` by design when any point in the window has `priceUsd: null` (tested in `supplyHistory.test.ts`), but here it renders as a retryable error — and refetching returns the same nulls, so the button can never clear it. The slice cards render "-" for that exact state. A successful-but-empty history lands here too. Worth splitting the genuine query error from "no data to plot".

— Auto Gabi

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/analytics/_components/supplyChart.tsx:325]: Gonzalo asked for the refresh in the USD case, so I'd rather keep it. And it won't be a dead button forever: the snapshot we ship has prices on all 180 days, so that state only turns up once the real endpoint lands, and there a retry can actually bring them back. You're right about the cards though, they say "-" for the same thing. Thoughts?

gabmontes (approved):
uACK

gndelia:
since #2302  was merged, we can drop the hardcoded data and use the real endpoint directly (note that I'm getting that deployed, which may take up to 2 hours)

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2297#issuecomment-5716816189):
> since #2302 was merged, we can drop the hardcoded data and use the real endpoint directly

Done in [`4238a783`](https://github.com/hemilabs/ui-monorepo/pull/2297/commits/4238a783). Dropped the bundled JSON and the client side slicing since the endpoint already slices by period.

gndelia (approved):
uACK

PR #2298, opened 2026-09-16
Show the staked total in dollars

ArturDolzan:
### Description

This PR adds the dollar value next to the staked total on the Total staked card.

### Screenshots

Captura de Tela 2026-09-16 às 15 38 00

### Related issue(s)

Related to #2244

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-stake/_components/totalStaked.tsx
```diff
@@ -1,36 +1,50 @@
...
+  const usd =
+    data === undefined || quoted === '0'
+      ? undefined
+      : Big(staked).times(quoted).toNumber()
 
   return (
     <StakeStatCard
+      badge={
+        usd === undefined ? undefined : (
+          <StatBadge>{formatCompactFiat(usd, locale, 2)}</StatBadge>
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/_components/totalStaked.tsx]: I think we already have a component that does part of what's here: [RenderFiatBalance](https://github.com/hemilabs/ui-monorepo/blob/main/portal/components/fiatBalance.tsx). Perhaps you can use it. IT's a component, so maybe `StatBadge` needs to accept `ReactNode`

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/totalStaked.tsx]: Good idea, switched to `RenderFiatBalance` in [`bbaf26d4`](https://github.com/hemilabs/ui-monorepo/pull/2298/commits/bbaf26d4).

portal/app/[locale]/hemi-stake/_components/totalStaked.tsx
```diff
@@ -1,36 +1,54 @@
...
 
+  const formatBadge = (amount: string) =>
+    Number(amount) > 0 ? formatCompactFiat(Number(amount), locale, 2) : '-'
+
   return (
     <StakeStatCard
+      badge={
+        <StatBadge>
+          <RenderFiatBalance
+            balance={staked}
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/totalStaked.tsx]: `staked` is `BigInt(data ?? 0)`, so on error/idle `RenderFiatBalance` receives a defined `0n` and takes its success path — the `> 0` guard in `formatBadge` is the only thing turning that into `-`. Passing `data !== undefined ? BigInt(data) : undefined` moves that back into the component, lets the guard go away, and makes a genuine `$0.00` renderable.

— Auto Gabi

portal/app/[locale]/hemi-stake/_components/totalStaked.tsx
```diff
@@ -1,36 +1,54 @@
...
+    Number(formatUnits(staked, token.decimals)),
     locale,
   )
 
+  const formatBadge = (amount: string) =>
+    Number(amount) > 0 ? formatCompactFiat(Number(amount), locale, 2) : '-'
+
   return (
     <StakeStatCard
+      badge={
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/totalStaked.tsx:37]: On error this card shows `-` as the value plus a badge chip containing just `-`; `AverageLockup` and `RewardsPaid` render no badge instead. Suggest gating on the same `isUnavailable && data === undefined` expression already used for `isError` — gating on `data === undefined` alone would also drop the badge skeleton during the initial load.

— Gabi 🤖

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/totalStaked.tsx:37]: Good catch, gated it on the same expression the card uses for `isError` in [`f30a37a2`](https://github.com/hemilabs/ui-monorepo/pull/2298/commits/f30a37a2). The skeleton still shows on the first load.

portal/app/[locale]/hemi-stake/_components/totalStaked.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/totalStaked.tsx]: Right about the success path, but the guard isn't for missing data, it's for a missing price. `getTokenPrice` answers `'0'` and that multiplies out to a convincing `$0.00` whether `balance` is `0n` or `undefined`. Gating the badge on the error state (other thread) covers what you're after, so I left the balance alone.

Comment (gabmontes) [reply portal/app/[locale]/hemi-stake/_components/totalStaked.tsx]: But it looks like "-" means both "price missing" and "genuinely zero". Telling them apart needs the price, not the balance. Is that the intended behavior/logic?

Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/totalStaked.tsx]: Yeah, intended. Both mean "no dollar figure here", and I'd rather collapse them than let the badge claim `$0` on a day the feed is down. Splitting them would mean reading the price in the card again, which is what moving to `RenderFiatBalance` just got rid of, and the only case it separates is a zero total that doesn't happen on mainnet. Thoughts?

gndelia (approved):

gabmontes (approved):

PR #2299, opened 2026-09-16
Store the daily token price history

gndelia:
### Description

This PR updates `token-prices` to store the historical prices of some tokens (up to 1 year). This will be used for calculating historical prices of Hemi Supply. The values to track will be added through Env variables, though initially we just need HEMI.

The stored values look like this

```json
{ 
"daily-prices:BTC": 
  {
      "2025-09-16": "116843.40493587981",
      "2025-09-17": "116467.42724651704"
  }
}
```

This will be read by the API (in the next PR) to return the hemi supply info

The values stored look like this

### Related issue(s)

Related to #2286

### Checklist

- [x] Manual testing passed.


portal-backend/cron/token-prices/src/refresh-history.js
```diff
@@ -0,0 +1,121 @@
...
+  const stored = await client.get(key)
+  const history = stored ? JSON.parse(stored) : {}
+  const lastDate = toDate(Date.now() - dayMs)
+  const newest = newestDate(history)
+  if (newest === lastDate) {
+    return 0
+  }
+  const windowStart = toDate(startOfDay(lastDate) - (historyDays - 1) * dayMs)
+  const firstDate =
+    newest > windowStart ? toDate(startOfDay(newest) + dayMs) : windowStart
```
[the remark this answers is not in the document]
gndelia: If a price was not found it's probably because there was no price to begin with. We can revisit if not, but I think the current implementation is correct

portal-backend/cron/token-prices/src/config.js
```diff
@@ -6,11 +6,17 @@ const config = {
   cacheExpirationMin: Number(env('CACHE_EXPIRATION_MIN', '60')),
   coinMarketCap: {
     apiKey: env('COIN_MARKET_CAP_API_KEY', ''),
+    ids: Object.fromEntries(
+      env('COIN_MARKET_CAP_IDS', 'HEMI:38159')
+        .split(',')
+        .map(pair => pair.split(':')),
```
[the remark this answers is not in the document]
gndelia: Fixed in [8dc3d64a](https://github.com/hemilabs/ui-monorepo/pull/2299/commits/8dc3d64a5a7a480460174a68bb0e21c1607f9f3e). Both variables are now in the token price cron configuration table, with their defaults and the `SYMBOL:id` format.

portal-backend/cron/token-prices/src/refresh-history.js
```diff
@@ -0,0 +1,121 @@
...
+
+const toDate = time => new Date(time).toISOString().slice(0, 10)
+
+const startOfDay = date => Date.parse(`${date}T00:00:00Z`)
+
+// A daily quote stamped at midnight closes the day before it.
+const closingDate = time => toDate(Math.round(time / dayMs) * dayMs - 1)
+
+// Kept in sync by hand with api/src/coin-market-cap.ts. Once this cron is
+// written in TypeScript, both can read it from a shared package.
```
ArturDolzan: Perhaps it's worth double-checking the `api/src/coin-market-cap.ts` reference here. I couldn't find that file anywhere in the repo, so I wondered if we meant `./refresh-prices.js` (which shares `hasPrice` and the CMC headers).

portal-backend/README.md
```diff
@@ -265,8 +265,10 @@ These environment variables control how the `cron` job behaves:
 | ----------------------- | ----------------------------------------------------------------------------------- | ------------------------ |
 | CACHE_EXPIRATION_MIN    | How long the prices will be kept in the cache.                                      | 3600                     |
 | COIN_MARKET_CAP_API_KEY | The CoinMarketCap API key.                                                          |                          |
+| COIN_MARKET_CAP_IDS     | Comma separated `SYMBOL:id` pairs whose daily price history is kept.                | HEMI:38159               |
 | COIN_MARKET_CAP_SLUGS   | String of comma separated token slugs. I.e. "bitcoin,ethereum"                      | bitcoin                  |
 | REDIS_URL               | The URL of the Redis database.                                                      | `redis://localhost:6379` |
+| REFRESH_HISTORY_MIN     | How frequently the price history is completed. If 0, it runs once and exits.        | 5                        |
```
ArturDolzan: Not sure, but maybe the "if 0, it runs once and exits" wording is a bit off now that two intervals share the process. With `REFRESH_HISTORY_MIN=0` and prices still on a timer, the process keeps running. Wonder if "runs only once at start" is closer?

portal-backend/README.md
ArturDolzan, line 1: Might be worth adding the new `daily-prices:` prefix to the "Stored data" section, the way the supply cron documents `supply:`. Would help whoever reads it from the API next.

portal-backend/cron/token-prices/server.js
```diff
@@ -19,9 +21,25 @@ async function run() {
...
 } else {
   run()
 }
+
+const refreshHistoryMin = config.get('refreshHistoryMin')
+if (refreshHistoryMin > 0) {
+  startInterval(runHistory, refreshHistoryMin * 60 * 1000)
+} else {
+  runHistory()
+}
```
gabmontes: I would merge this with the prior `if`. That way, the server either run once or in a loop. I would also drop `refreshHistoryMin`.

portal-backend/cron/token-prices/src/refresh-history.js
gndelia: you can't find it because it comes in the next PR which is all about backend changes. However,  I changed my mind of the approach, and there won't be such code in backend, and I was dropping this in that (incoming) PR, but since there are changes, I'm doing so here 😅

portal-backend/README.md
gndelia: Coming in the next PR ! (I already committed it)

portal-backend/cron/token-prices/src/refresh-history.js
gndelia: Fixed in [2ce0edd4](https://github.com/hemilabs/ui-monorepo/pull/2299/commits/2ce0edd4b135ff3127ae22925d7f00fe9aef8ce6). The comment is dropped, as the backend file it pointed at will not be added.

portal-backend/README.md
gndelia: Good catch. Addressed in [f219c585](https://github.com/hemilabs/ui-monorepo/pull/2299/commits/f219c585e8357f78a7a04679ba50563c50dbcbd9), which removes `REFRESH_HISTORY_MIN` and its row: the price history now runs on the same interval as the prices, so the mixed state you describe can no longer happen.

portal-backend/cron/token-prices/server.js
gndelia: Fixed in [f219c585](https://github.com/hemilabs/ui-monorepo/pull/2299/commits/f219c585e8357f78a7a04679ba50563c50dbcbd9). One interval now drives both refreshes, and `refreshHistoryMin` is gone from the config, the README and the compose file.

ArturDolzan (approved):

portal-backend/README.md
```diff
@@ -265,6 +265,7 @@ These environment variables control how the `cron` job behaves:
 | ----------------------- | ----------------------------------------------------------------------------------- | ------------------------ |
 | CACHE_EXPIRATION_MIN    | How long the prices will be kept in the cache.                                      | 3600                     |
 | COIN_MARKET_CAP_API_KEY | The CoinMarketCap API key.                                                          |                          |
+| COIN_MARKET_CAP_IDS     | Comma separated `SYMBOL:id` pairs whose daily price history is kept.                | HEMI:38159               |
```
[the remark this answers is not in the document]
gndelia: incoming in the following PR

gabmontes (approved):

PR #2300, opened 2026-09-16
Show the APY on each lockup preset

ArturDolzan:
### Description

This PR fills the sublabel on each lockup preset with its APY. The numbers come from the tokenomics sheet and are hardcoded for now. It also adds a tooltip providing clarification on the APY estimate.

### Screenshots

https://github.com/user-attachments/assets/c42af98b-ba7a-4a66-8aee-76335b6ca6fa

### Related issue(s)

Related to #2246

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
```diff
@@ -206,12 +207,19 @@ export function Lockup({
 
   const amount = parseTokenUnits(input, token)
 
+  const toSublabel = function (days: number) {
+    const percentage = formatNumber(lockupApy[days])
+    return days === maxDays
+      ? t('form.up-to', { percentage })
+      : `~${percentage}%`
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: Good point, moved it to `form.approximate` in all three locales in [`32f6067a`](https://github.com/hemilabs/ui-monorepo/pull/2300/commits/32f6067a). Kept the same `~{percentage}%` value in each for now, the tilde reads fine in es and pt and a spelled out "aprox." would not fit the 2x2 grid on small screens.

gndelia:
LEt's hold off merging this one until we get confirmation these numbers are ok (prob tomorrow?)

ArturDolzan:
> LEt's hold off merging this one until we get confirmation these numbers are ok (prob tomorrow?)

Absolutely!

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
```diff
@@ -311,9 +321,21 @@ export function Lockup({
...
+            <Tooltip
+              borderRadius="12px"
+              id="lockup-apy-estimate"
+              text={t('apy-estimate')}
+              variant="info"
+            >
+              <div className="group/icon flex items-center">
+                <InfoIcon className="[&>g>path]:transition-colors [&>g>path]:duration-200 group-hover/icon:[&>g>path]:fill-neutral-950" />
+              </div>
+            </Tooltip>
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: The tooltip renders next to the label in every mode, but the APY sublabels only exist in preset mode. In custom mode (`showSlider`) and in the extend-lock flow (`minLocked` set, so `canUsePresets` is false) it explains percentages that aren't on screen. Consider rendering it only when `canUsePresets && !showSlider`.

Gabi 🤖

portal/app/[locale]/hemi-stake/_utils/lockupApy.ts
```diff
@@ -0,0 +1,8 @@
+import { maxDays, oneYear, sixMonths, twoYears } from './lockCreationTimes'
+
+export const lockupApy: Record<number, number> = {
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_utils/lockupApy.ts:3]: `Record<number, number>` hides missing keys: if a preset is added or changed without updating this map, `lockupApy[days]` is `undefined` and `formatNumber` throws (`input should be a valid number`), crashing the form instead of just dropping the sublabel. Colocating the APY with each preset entry, or guarding the lookup before formatting, would fail more gracefully.

Auto Gabi

portal/stories/lockupPresets.stories.tsx
```diff
@@ -49,10 +49,10 @@ export const Hover: Story = {
 export const WithApr: Story = {
```
gabmontes, line 49: Nit: the story is still named `WithApr` while it now shows APY values. Worth double-checking the terminology mix too — the stake table renders live `{percentage}% APR` while these presets show APY estimates on the same page.

Gabi 🤖

portal/app/[locale]/hemi-stake/_components/lockup/index.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/lockup/index.tsx]: Right, the tooltip was sitting outside the condition that gates the presets. Pulled both behind a single `showPresets` flag in [`ac382beb`](https://github.com/hemilabs/ui-monorepo/pull/2300/commits/ac382beb), so it can only show up next to the sublabels it describes.

portal/app/[locale]/hemi-stake/_utils/lockupApy.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_utils/lockupApy.ts:3]: Fair point. Guarded the lookup in [`ac382beb`](https://github.com/hemilabs/ui-monorepo/pull/2300/commits/ac382beb), a missing key now just drops the sublabel instead of throwing. I kept the map separate since it reads as the tokenomics source of truth, and the preset keys are computed so there's no literal union to make the Record exhaustive at the type level anyway.

portal/stories/lockupPresets.stories.tsx
ArturDolzan: Good catch on both. The APR is coming out of the stake table in a separate change, so the mix on that page sorts itself out. I left the story name alone for now, happy to rename it to `WithApy` if you'd rather not wait.

gndelia (dismissed):

gndelia (approved):

PR #2302, opened 2026-09-17
Add endpoint that returns Hemi Supply data

gndelia:
### Description

This PR updates `/circulating` to read from the `Envio` indexer, plus adds the endpoint `/supply-history/:period` which lists the hemi supply info for a given period.

The example data was already shown on https://github.com/hemilabs/ui-monorepo/pull/2288

Do note that `priceUSD` may be null if CMC or Redis fails somehow, so the FE should handle that (which I think it does)

### Related issue(s)

Related to #2286

### Checklist

- [x] Manual testing passed.


portal-backend/api/config/default.json
```diff
@@ -52,6 +52,10 @@
       "testnet": "HFj92373tN3eEdEtYqsWmVftpwhshRGzfwgzg9eG3W7S"
     }
   },
+  "supply": {
+    "correction": "350000000000000000000000000",
```
ArturDolzan, line 56: Could be worth confirming the deployed `SUPPLY_CORRECTION` really is `350000000000000000000000000`. The cron still reads it from the env (defaulting to `0`), so if they ever disagree `/circulating` would shift by that amount with no way to fix it without a release. Maybe map it in `custom-environment-variables.json` too?

portal-backend/README.md
```diff
@@ -80,6 +82,17 @@ $ curl http://localhost:3006/prices
 {"prices":{"BTC":"94514.79193898945","M-BTC":"95894.52612269788","PUMPBTC":"95990.74415080296","WBTC":"95797.80677773379"},"time":"2025-02-17T23:12:35.803Z"}
 ```
 
+#### `GET /supply-history/:period`
+
+Returns the daily HEMI supply of the last `1w`, `1m`, `3m`, `6m` or `1y`, ending yesterday. Any other period returns `400 Bad Request`.
+
+```console
+$ curl http://localhost:3006/supply-history/1m
+[{"circulating":"1052500000000000000000000000","date":"2026-09-10","nonCirculating":"7459107944000000000000000000","priceUsd":"0.00683108945841876","staked":"1488392055000000000000000000","totalSupply":"10000000000000000000000000000"}]
```
ArturDolzan: Not sure, but the sample numbers seem to come out 1 HEMI short of `totalSupply`, right under the line saying they add up. Perhaps worth nudging one of them so the example matches the invariant?

portal-backend/api/config/default.json
gndelia: It is - it is configured as Env variable, though imho it doesn't make any sense to read as env variable

portal-backend/README.md
gndelia: Fixed in [21c49d31](https://github.com/hemilabs/ui-monorepo/pull/2302/commits/21c49d313f4e7f216ad35d4c2fae2b9fbeb70a44). `staked` is now `1488392056…`, so the three values add up to `totalSupply`.

ArturDolzan (approved):

gabmontes (approved):

PR #2315, opened 2026-09-18
Show the Hemi Stake table on any connected chain

ArturDolzan:
### Description

This PR drops the "You're not connected to a supported chain" card from the stake positions table, the same change #2252 made for the tunnel history. Positions are read-only and already keyed on Hemi, so the address alone is enough, and this was the last wrong-network card left in the portal.

Split into four commits, the first three being fixes that stand on their own:
- `2ef37841` - Read the stake data from the Hemi public client
- `64f9d27c` - Pin the stake gas estimates to the Hemi chain
- `b4392edf` - Guard the stake writes on the Hemi chain
- `044e5268` - Show the stake table on any connected chain

### Screenshots

N/A - No UI changes.

### Related issue(s)

Related to #2252

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-stake/_hooks/useEstimateCreateLockFees.ts
```diff
@@ -24,6 +24,7 @@ export const useEstimateCreateLockFees = function ({
   })
 
   const { data: gasUnits, isError } = useEstimateGas({
+    chainId: token.chainId,
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_hooks/useEstimateCreateLockFees.ts:27]: The stake estimates are pinned now, but `useEstimateApproveErc20Fees` still calls `useEstimateGas` without `chainId`, so it estimates on the *connected* chain while `useEstimateFees` prices the units on `token.chainId`. With the table reachable from any chain, the approve call hits a codeless address on e.g. Ethereum and returns intrinsic gas instead of erroring, so the "Network gas fee" total in `stake.tsx`, `reviewStake.tsx` and `increaseAmount/index.tsx` is silently understated until the user switches to Hemi. Worth pinning that hook too.

— Auto Gabi

portal/app/[locale]/hemi-stake/_components/stakeTable/unlock.tsx
```diff
@@ -78,7 +78,7 @@ export function Unlock({ operation, unlockDate }: Props) {
 
   const handleUnlock = function () {
     setOperationRunning('unlocking')
-    runUnlock()
+    runUnlock(undefined, { onError: () => setOperationRunning('failed') })
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_components/stakeTable/unlock.tsx]: Nit: `actionCell.tsx` resets to `'idle'` on this same early-rejection path (declined switch / no account) while here it goes to `'failed'`. Same behavior today since only `'unlocking'` is read, but `'idle'` would keep both flows consistent.

— Gabi 🤖

portal/app/[locale]/hemi-stake/_hooks/useCalculateApr.ts
```diff
@@ -54,7 +54,7 @@ export const useCalculateApr = function ({
   return useQuery({
     enabled:
       enabled &&
-      !!hemiWalletClient &&
+      !!hemiClient &&
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-stake/_hooks/useCalculateApr.ts]: Nit: `useHemiClient` returns `usePublicClient(...)!` and both Hemi chains are always in the wagmi config, so this is always true — leftover from the wallet-client version. Same in `useCalculateRewards.ts` and `useRewardTokensAddresses.ts`.

— Auto Gabi

portal/app/[locale]/hemi-stake/_hooks/useEstimateCreateLockFees.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_hooks/useEstimateCreateLockFees.ts:27]: You're right that the approve estimate isn't pinned, the chainId on line 33 is on useEstimateFees, not on the useEstimateGas above it.

I'd rather not touch it here though. That hook is shared by 10 call sites across the tunnel, hemi-earn and Boost, so pinning it would move gas numbers well outside the stake table and I can't test all of those in this PR.

portal/app/[locale]/hemi-stake/_components/stakeTable/unlock.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/stakeTable/unlock.tsx]: Good catch, switched to `'idle'` in [`7c14c19f`](https://github.com/hemilabs/ui-monorepo/pull/2315/commits/7c14c19f).

portal/app/[locale]/hemi-stake/_hooks/useCalculateApr.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_hooks/useCalculateApr.ts]: Yep, traced it: `useHemi()` only ever returns hemiMainnet or hemiTestnet and both are unconditional in `allEvmNetworks`, so it was always true. Dropped in all three in [`7c14c19f`](https://github.com/hemilabs/ui-monorepo/pull/2315/commits/7c14c19f).

gndelia (approved):

portal/app/[locale]/hemi-stake/_hooks/useEstimateCreateLockFees.ts
Comment (gabmontes) [reply portal/app/[locale]/hemi-stake/_hooks/useEstimateCreateLockFees.ts:27]: @ArturDolzan please open an issue to revise and fix as needed.

gabmontes (approved):

portal/app/[locale]/hemi-stake/_hooks/useEstimateCreateLockFees.ts
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_hooks/useEstimateCreateLockFees.ts:27]: Opened #2317 with the details.

PR #2330, opened 2026-09-21
Flip the burned positions to newest first

### LINKED ISSUE #2324: Sort veHemi positions by closest to unlock

Sort the position list so the ones closest to unlocking show up first.

The table renders whatever order the subgraph returns, so there's no way to spot the position that's about to be claimable. That's usually the one you opened the page for.

### Proposed implementation

Sort by unlock time ascending before rendering. Nothing sorts today, `useStakingPositions` passes the subgraph result straight through.

ArturDolzan:
### Description

This PR flips the Burned tab so the most recent position sits on top. The sorting itself moved to the API in #2333, which returns positions by unlock time, soonest first. That's what the Active tab wants, but Burned positions have all matured already, so reading that tab as history makes more sense.

**Depends on #2333 being deployed.** Without it the endpoint still returns positions by creation date, and reversing that is worse than leaving it alone.

One correction to what I wrote on the earlier version of this PR: I claimed the endpoint made no ordering promise. It does, the query has always had an `orderBy`, and gndelia was right to point it out.

### Screenshots

N/A

### Related issue(s)

Closes #2324

### Checklist

- [x] Manual testing passed.


gndelia (commented):
The list came with the sorting defined in [the subgraph query](https://github.com/hemilabs/ui-monorepo/blob/d659d290d4bd56ff651defa1a61f1ffb65b95116/portal-backend/api/src/subgraphs/subgraph.ts#L674)

So perhaps you can solve the sorting on the backend instead of doing so in the frontend - specially if you're not going to offer a sorting option in the table

Not for this PR, but we definitely should design a way to click a column and make an action (like sorting). This should be part of the design system. With that, while the BE keeps the sorting defined, the user can sort with other directions

gndelia:
Some findings by AI (You may want to double check)

1. The burned tab also contains positions, and these are affected by the sorting

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2330#pullrequestreview-5270764530):
> The list came with the sorting defined in the subgraph query. So perhaps you can solve the sorting on the backend instead of doing so in the frontend

You were right, and I was wrong in the old description: the query does define an order, I claimed it didn't. #2333 just changes that default to unlock time.

I had added orderBy and orderDirection params there too, but you and gabmontes both pointed out that's premature, so they're out. This PR is only the reverse for Burned, since the two tabs want opposite directions off a single request.

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2330#issuecomment-5766164845):
> 1. The burned tab also contains positions, and these are affected by the sorting

Yes, and that's intentional, Burned now reads newest first. Worth flagging what I found while checking it though: there's no record of when a position was actually burned. handleWithdraw in the subgraph only flips the status, so the tab is ordered by scheduled unlock time. For anything that matured normally that's the same thing, but a forfeitable position burned early will sort as if it were still pending. Happy to open an issue for a withdrawnAt field if you think it's worth it.

gabmontes (approved):

gndelia (approved):

PR #2331, opened 2026-09-21
Add VITE_VETRO_API_URL to commited file

gndelia:
### Description

As the title says. The URL is public, and we use the same API in all environments, so we don't need to set it in any other place other than the commited `.env` (ofc it can be overriden with a `.env.local`

### Screenshots

N/A

### Related issue(s)

N/A

### Checklist

- [x] Manual testing passed.


gabmontes (commented):
Was `VITE_VETRO_API_URL` removed from the CF environment already? If not, it would take precedence.

gndelia:
> Was `VITE_VETRO_API_URL` removed from the CF environment already? If not, it would take precedence.

no. It needs to be removed _After_ this is merged. They have the same value so this can be merged safely, and then, after it lands on prod, removed from every env

gabmontes (approved):

ArturDolzan (approved):

PR #2334, opened 2026-09-22
Show the unlock date and countdown per position

### LINKED ISSUE #2325: Show the unlock date and countdown in the Lockup column

The Lockup column should show the unlock date with a countdown under it, and the Unlock button should read just "Unlock".

The button carried the countdown at all times, which made it much wider than the other row actions. And the column showed the lock duration, which never changes and doesn't answer the question people open the page for.

Duration and countdown can't share the column: both round to their biggest unit, so a position created three days ago with a two month lock renders "2 months" on both lines. That happens on 22 of 27 positions in a real wallet, since the two only diverge once a meaningful part of the lock has elapsed.

### Proposed implementation

Show `formatDate(unlockDate)` with `InRelativeTime` below it in `lockupTime.tsx`, and drop the `<Badge>` from `unlockCta.tsx`. The duration moves into the column tooltip, which keeps it available without spending a line on it.

ArturDolzan:
### Description

This PR leads the Lockup column with the unlock date and counts down underneath it, and takes the countdown off the Unlock button so it reads just "Unlock". The column used to show the lock duration, which never changes and never answers the question people open the page for, while the always-on countdown made that button much wider than the other row actions.

Duration and countdown can't share the column. Both round to their biggest unit, so a lock opened three days ago for two months renders the same words on both lines, which holds for 22 of the 27 positions in a real wallet. The two only pull apart once a good part of the lock has elapsed, so the duration moves into the column tooltip instead.

The column goes from 80px to 125px. I measured the rendered text in the real font rather than guessing: the widest case is `Dentro de 40 minutos` at 109px. That pushes the table's minimum width from 745px to 790px, so the horizontal scroll it already had below `lg` starts a little earlier.

One new message, `lockup-tooltip`, in the three locales. `DurationTime` lost its only consumer, so its `formatDuration` is exported and the component wrapper is gone; the polyfill still ships with that module, so no dependency changed.

### Screenshots

https://github.com/user-attachments/assets/b424dda6-29dc-4aa7-95ce-ab6857a38726

### Related issue(s)

Closes #2325

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-stake/_components/stakeTable/lockupTime.tsx
```diff
@@ -1,11 +1,37 @@
...
+    >
+      <div className="flex flex-col">
+        <span className="text-neutral-950">
+          {formatDate(unlockDate, locale)}
+        </span>
+        <span className="body-text-caption whitespace-nowrap text-neutral-500 first-letter:uppercase">
+          <InRelativeTime timestamp={unlockTime} />
+        </span>
+      </div>
+    </Tooltip>
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/stakeTable/lockupTime.tsx:42]: Fair point, though I'd leave it as is. The duration is supplementary here: the unlock date and the countdown are both visible text, and the team's read is that the total lockup isn't information people need. I'd rather not put a button in a cell that has no action behind it.

portal/messages/en.json
```diff
@@ -587,6 +587,7 @@
       "get-started-burned": "Your burned positions will show up here once their lockup periods have ended.",
       "locked-amount": "Locked Amount",
       "lockup": "Lockup",
+      "lockup-tooltip": "{duration} lock, unlocks on {date}",
```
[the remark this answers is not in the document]
ArturDolzan: Right, fixed in [`fe9c8f83`](https://github.com/hemilabs/ui-monorepo/pull/2334/commits/fe9c8f83). Only the English one, "Bloqueio de 2 meses" and "Bloqueo de 2 meses" are already correct so I left those alone.

portal/app/[locale]/hemi-stake/_components/stakeTable/lockupTime.tsx
```diff
@@ -1,11 +1,37 @@
...
+        duration: formatDuration(Number(lockTime), locale),
+      })}
+      variant="simple"
+    >
+      <div className="flex flex-col">
+        <span className="text-neutral-950">
+          {formatDate(unlockDate, locale)}
+        </span>
+        <span className="body-text-caption whitespace-nowrap text-neutral-500 first-letter:uppercase">
+          <InRelativeTime timestamp={unlockTime} />
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/_components/stakeTable/lockupTime.tsx]: I think `lockupTime` component is also used in the Burned table. That will cause that `InRelativeTime` shows past time as a reason to unlock, which makese no sense. Probably may want to render this only for active positions

gndelia (changes requested):

portal/utils/duration.ts
```diff
@@ -42,11 +40,3 @@ const formatDuration = function (durationSeconds: number, locale: string) {
...
   }).format({ seconds: 0 })
 }
-
-export const DurationTime = function ({ seconds }: DurationTimeProps) {
-  const locale = useLocale()
-
-  const formattedDuration = formatDuration(seconds, locale)
-
-  return formattedDuration
-}
```
gndelia, line 52: It seems this file is no longer  a component, so it should not be on `components` folder

portal/app/[locale]/hemi-stake/_components/stakeTable/lockupTime.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/stakeTable/lockupTime.tsx]: Good catch, "3 months ago" on a burned position reads as a countdown that already ran out. Gated it on the status in [`d69c0700`](https://github.com/hemilabs/ui-monorepo/pull/2334/commits/d69c0700), so Burned keeps the date and drops the countdown.

portal/utils/duration.ts
ArturDolzan: Right, it stopped being a component when the wrapper went away. Moved to utils/duration.ts in [`d69c0700`](https://github.com/hemilabs/ui-monorepo/pull/2334/commits/d69c0700). The polyfill import travels with it, so the note about keeping the function next to it still holds.

gndelia (approved):

gabmontes (approved):
uACK

Issue #1610, opened 2025-10-31
Improve AI integration in Github


gndelia:
This is an open issue to make some improvements to Copilot integration. This is a list of features (not limited) to some changes I've seen we could apply in our workflow

- [ ] When assigning issues to Copilot, we can add a `copilot-setup-steps.yml` file so the LLM does not need to figure out how to set the environment and it's provided by us. See [docs](https://docs.github.com/en/enterprise-cloud@latest/copilot/tutorials/coding-agent/get-the-best-results#pre-installing-dependencies-in-github-copilots-environment). We could here just install the deps and provide the command to start up the application locally
- [ ] Review that the [current instructions](https://github.com/hemilabs/ui-monorepo/tree/main/.github/instructions) are being taken into consideration. See [this example PR](https://github.com/copilot/tasks/pull/PR_kwDOK8fHCs6w3dYW?session_id=0a51a190-a2d3-476d-ae86-ddb3da0cbf09) generated by Copilot, which seems to ignore all the instructions, or at least they're not being referenced (but it was able to figure out everything)
- [ ] A playwright MCP server is started as part of the flow, so maybe we could instruct the AI to take screenshots of the changes made. It may not be that important after https://github.com/hemilabs/ui-monorepo/issues/1563 - but we could use a [QA agent](https://docs.github.com/en/enterprise-cloud@latest/copilot/tutorials/coding-agent/get-the-best-results#creating-custom-agents) to test the change implemented looks correct

gndelia:
Let's also use this issue for AI improvements in general

PR #2337, opened 2026-09-22
Add a run-checks skill and align check scripts

gndelia:
### Description

Adds `.claude/skills/run-checks/SKILL.md`, a skill that tells Claude how to verify a change here: what to format and lint, which workspaces to typecheck and test, and when to run the whole CI matrix instead.

For that, every workspace now uses the same script names:

- `typecheck` only checks types. It writes nothing.
- `build` produces what we ship: `dist/` for the packages the API image uses, the Vite bundle for the portal, the WASM for the subgraphs. Where there is nothing to emit, `build` just calls `typecheck`.

Before, some workspaces called the type check `tsc:check` and others had no name for it, so no single command worked everywhere.

Also: `prebuild` in the subgraphs, so `build` always runs `codegen` first; and a CLAUDE.md fix, because the cron images build from the repo root, not from their own folders.

Skill inspired on a similar made for Vetro

### Screenshots

N/A

### Related issue(s)

Related to #1610

### Checklist

- [x] Manual testing passed.


subgraphs/hemi-earn-requests-subgraph/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "hemi-earn-requests-subgraph",
   "version": "1.0.0",
   "scripts": {
+    "prebuild": "pnpm run codegen",
```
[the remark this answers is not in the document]
gndelia: Fixed in [bc2e568](https://github.com/hemilabs/ui-monorepo/pull/2337/commits/bc2e56802be9f471f1b723dac89bea6a88812ad5). `build` now calls a new `typecheck` script (`tsc --noEmit`), as in the other workspaces that do not emit JS.

.claude/skills/run-checks/SKILL.md
```diff
@@ -0,0 +1,123 @@
...
+
+- **Only docs (`*.md`)**: run prettier on them. Nothing else.
+- **Root config, lockfile or a dependency bump in many workspaces** (for
+  example `pnpm-workspace.yaml`, `pnpm-lock.yaml`, `patches/`, eslint,
+  prettier or knip config): go to [§3](#3-big-change-run-the-ci-matrix).
+- **Anything else**: go to §2.
+
+## 2. Scoped checks
+
+Format the changed files, and lint the changed `.js`, `.ts` and `.tsx` files.
```
ArturDolzan: Perhaps it's worth checking whether `.md` needs to be in the lint list here. `.eslintrc.json` has a `*.md` override and lint-staged runs eslint on `*.{js,md,ts,tsx}`, so a docs-only change might still trip markdownlint in CI.

.claude/skills/run-checks/SKILL.md
```diff
@@ -0,0 +1,123 @@
...
+
+- **`portal-backend/cron/<name>`** — no scripts. Review by hand, and see the
+  Docker note below.
+
+- **`subgraphs/<name>`** — `build` runs `codegen` first. `test` exists only in
+  some subgraphs.
+
+  ```bash
+  pnpm --fail-if-no-match --filter <name> run build
+  pnpm --filter <name> run --if-present test
```
ArturDolzan: Might be worth a quick look at the `test` line right under the guarded `build` one. Both use `<name>`, but only one has `--fail-if-no-match`, so a typo there would pass silently.

.claude/skills/run-checks/SKILL.md
gndelia: Fixed in [f30645d](https://github.com/hemilabs/ui-monorepo/pull/2337/commits/f30645d9e34f3d533c150dee47a2c28ea0226345). The skill now formats and lints `.md` files too, also for docs-only changes.

.claude/skills/run-checks/SKILL.md
gndelia: Fixed in [8949730](https://github.com/hemilabs/ui-monorepo/pull/2337/commits/8949730925e536e932953ac85e10cce365931b29).

ArturDolzan (dismissed):

gabmontes (commented):
Some scripts in the `package.json` files are `pnpm <script-name>` while others are `pnpm run <script-name>`. Both seem to work but we should stick to a single format. IMO we should be explicit here and use the long command version (with "run").

gndelia:
> Some scripts in the `package.json` files are `pnpm <script-name>` while others are `pnpm run <script-name>`. Both seem to work but we should stick to a single format. IMO we should be explicit here and use the long command version (with "run").

done in [6b12928](https://github.com/hemilabs/ui-monorepo/commit/6b12928dcda08f1257ec6da4ade61aef8f16888f)

gabmontes (approved):

ArturDolzan (approved):

arielmergen (approved):

PR #2339, opened 2026-09-22
Show the wallet totals on Hemi Stake

### LINKED ISSUE #2335: Move the global stats to analytics and show wallet totals

Show two boxes at the top of Hemi Stake with the connected wallet's totals: HEMI staked across all positions, and tokens available to claim across all of them. The three global stats move to the analytics page.

The page is about your positions but leads with protocol wide numbers, and the claimable amount is only visible one drawer at a time.

### Proposed implementation

The boxes only render with a wallet connected, so the hero banner leads for visitors. Rewards are computed per position and per reward token, so the claimable total needs a batched read rather than a loop.

ArturDolzan:
### Description

This PR opens Hemi Stake with what the connected wallet actually holds: the HEMI staked across its positions, and what it has available to claim. The three global stats that used to sit there move to the analytics tab, where the protocol wide numbers already live. The boxes only render with a wallet connected, so a visitor gets the hero first.

Split into two commits so the relocation is reviewable without the new code:
- `78863077` - Move the stake stats to the analytics page
- `9f3c89dc` - Show the wallet totals on Hemi Stake

Three decisions the diff won't explain. Claimable reads every position rather than only the active ones, because a burned position keeps paying the epochs it earned once `captureAndWithdraw` recorded its class, which is what Claim all already assumes. A disabled query counts as waiting, not failing, since the section mounts while wagmi is still reconnecting. And a price feed that never answers reads as an error instead of a confident $0.

### Screenshots

https://github.com/user-attachments/assets/7ae81f5d-2582-461f-a888-b77f2187455b

### Related issue(s)

Closes #2335

### Checklist

- [x] Manual testing passed.


portal/app/[locale]/hemi-stake/_utils/walletStakedTotal.ts
```diff
@@ -0,0 +1,9 @@
+import {
+  StakingPositionStatus,
+  type StakingPosition,
+} from 'types/stakingDashboard'
+
+export const sumActiveStake = (positions: StakingPosition[]) =>
+  positions
+    .filter(({ status }) => status === StakingPositionStatus.ACTIVE)
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_utils/walletStakedTotal.ts]: Good catch, the query matches pastOwners too so an active position the wallet handed off was still counted. Fixed in [`e819819a`](https://github.com/hemilabs/ui-monorepo/pull/2339/commits/e819819a), with a transferred-away case in the test.

portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+import { RewardAmount } from './rewardsDisplay'
+import { StakeStatCard } from './stakeStatCard'
+import { StatBadge, StatBadgeSkeleton } from './statBadge'
+
+export const WalletClaimable = function () {
+  const locale = useLocale()
+  const t = useTranslations('hemi-stake.stats')
+
+  const positionsQuery = useStakingPositions()
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx:40]: You're right, and it's worse than the cards: the table already says "No positions staked" on a failed request today. The catch is shared and pre-existing, so I'd rather not change that behaviour from here. Tracked in #2341.

portal/app/[locale]/hemi-stake/_components/walletStaked.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { sumActiveStake } from '../_utils/walletStakedTotal'
+
+import { StakeStatCard } from './stakeStatCard'
+import { StatBadge, StatBadgeSkeleton } from './statBadge'
+
+export const WalletStaked = function () {
+  const locale = useLocale()
+  const t = useTranslations('hemi-stake.stats')
+  const token = useHemiToken()
+  const { data, fetchStatus, status } = useStakingPositions()
```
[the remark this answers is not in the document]
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/walletStaked.tsx:20]: You're right, and it's worse than the cards: the table already says "No positions staked" on a failed request today. The catch is shared and pre-existing, so I'd rather not change that behaviour from here. Tracked in #2341.

gabmontes (dismissed):
LGTM

portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx
```diff
@@ -0,0 +1,122 @@
...
+      isLoading={isLoading}
+      label={t('your-claimable')}
+      value={
+        <ErrorBoundary fallback="-">
+          {formatCompactFiat(
+            Number(calculateUsdValue(withBalance, pricesQuery.data ?? {})),
+            locale,
+            2,
+          )}
+        </ErrorBoundary>
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx:130]: I think this is wrongly used. ErrorBoundary are for component renders. For this to work, you'd have to use something like


```tsx
<ErrorBoundary fallback="-">
  <Format value={...}" />
</ErrorBoundary>
```

Additionally, don't we have components to render USD values that already have ErrorBoundaries? We do..

portal/messages/pt.json
```diff
@@ -569,8 +569,11 @@
     "stats": {
       "average-lockup": "Bloqueio médio",
       "average-lockup-badge": "Em {count, plural, one {# bloqueio} other {# bloqueios}}",
+      "claimable-tokens": "{count, plural, one {# token} other {# tokens}}",
       "total-staked": "Total em stake",
-      "wallets-staking": "Carteiras em stake"
+      "wallets-staking": "Carteiras em stake",
+      "your-claimable": "Disponível para resgatar",
```
gndelia: (Flagged by AI) - should it be 

```suggestion
      "your-claimable": "Disponível para reclamar",
```
?

gndelia (commented):
I think we should follow the example of data-first checked when using react-query queries to render data, as stated in [CLAUDE.md](https://github.com/hemilabs/ui-monorepo/blob/main/CLAUDE.md#frontend-rules)

```tsx
const { data, error } = useTodos()

if (data) {
  return <div>{data.map(renderTodo)}</div>
}
if (error) {
  return 'An error has occurred'
}

return 'Loading...'
```

portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx
```diff
@@ -0,0 +1,122 @@
...
+    fetchStatus: pricesQuery.fetchStatus,
+    status: pricesQuery.status,
+  })
+
+  const isError = [
+    positionsQuery.isError,
+    pricesUnavailable,
+    rewardsQuery.isError,
+    hasError,
+  ].some(Boolean)
```
Comment (gndelia) [inline portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx:76]: minor, but `pricesUnavailable` can be skipped if there's nothing to claim (We can just show 0 instead of `-`)

portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx:130]: You're right, it was catching nothing: the value got computed while building the prop, so the throw happened in the parent's render. Moved it into a child component so the boundary actually wraps it, in [`3362c5ea`](https://github.com/hemilabs/ui-monorepo/pull/2339/commits/3362c5ea).

On reusing what we have, `RenderFiatBalance` does have the boundary, but it prices one token at a time and this card sums across reward tokens. So I followed its shape instead of calling it.

portal/messages/pt.json
ArturDolzan: Good catch, "reclamar" is what the rest of pt.json uses... Both works well though. Applied in [`3362c5ea`](https://github.com/hemilabs/ui-monorepo/pull/2339/commits/3362c5ea).

portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-stake/_components/walletClaimable.tsx:76]: Makes sense, with nothing to claim the price feed doesn't matter and $0 is the honest answer. Gated it in [`3362c5ea`](https://github.com/hemilabs/ui-monorepo/pull/2339/commits/3362c5ea).

ArturDolzan:
> [@gndelia commented](https://github.com/hemilabs/ui-monorepo/pull/2339#pullrequestreview-5285207966):
> I think we should follow the example of data-first checked when using react-query queries to render data, as stated in CLAUDE.md

Agreed. The snag is `StakeStatCard` takes `isError`/`isLoading` as props and resolves them inside, so a caller can't check data first unless the card changes shape too. Doing it in just my two cards would leave two patterns sitting next to the other three, so I opened #2343 to change the card and move all five over at once.

gabmontes (approved):

gndelia (approved):

PR #2347, opened 2026-09-23
Apply the new stat card design and share one component

### LINKED ISSUE #2345: Update the stat card to the new design

Restyle the stat card to match the new design: [link](https://www.figma.com/design/4fVd9wneclsvYDYD95ApZ9/Hemi-Portal?node-id=21847-18596&t=EXS5rt9p2J6nP212-1)

The same card is implemented three times and the copies have already drifted: `StakeStatCard` (Hemi Stake and analytics, 7 cards), `PoolCard` (Hemi Earn pool, same props but `text-2xl` instead of `text-xl`) and the `Container`/`Points` pair on the old stake dashboard. Restyling one leaves the other two behind.

### Proposed implementation

Decide whether these collapse into a single shared component under `components/` before applying the design, otherwise we restyle one and the drift gets worse. The badge slot only exists on `StakeStatCard`, so the Figma should say whether it stays.

ArturDolzan:
### Description

This PR applies the new stat card design from Figma and collapses the three copies of that card into one shared component. Hemi Stake, its analytics page and Hemi Earn each rendered their own version, and they'd already drifted apart: two at 24px and one at 20px, one with a badge slot and two without.

The value now sits at 24px with the badge beside it. Two things deliberately don't match the Figma. The card keeps its 12px radius, since changing the shared `Card` would repaint 18 other files. And the badge wraps under the value when the two don't fit, which the four supply cards on analytics always hit, because there the badge alone is about as wide as the card.

Split into six commits so the moves read apart from the redesign:
- `d8b1e5d0` - Move the stat card to the components folder
- `8dcaa3be` - Apply the new design to the stat card
- `ed166813` - Use the shared stat card in Hemi Earn
- `9bd2563b` - Replace the stat badge with the shared badge
- `c86e6a0e` - Shrink the wallet icon on the stake card
- `8185be07` - Align the badge with the bottom of the value

### Screenshots

https://github.com/user-attachments/assets/62eb3b00-f730-40bd-a3e0-7ffec34c3646

### Related issue(s)

Closes #2345

### Checklist

- [x] Manual testing passed.


portal/components/statCard.tsx
```diff
@@ -35,9 +37,7 @@ export const EarnCard = ({
             <Skeleton className="h-7 w-20" />
```
gabmontes: `leading-none` makes the loaded value 24px tall, but the skeleton is still `h-7` (28px). Measuring the rendered markup, the value row is 28px while loading and 24px once data lands, so every stat card jumps 4px — and since the cards sit in stretching grids, the whole row moves with it. On `main` there was no shift (32px in both states). `h-6` matches the new value height exactly.

— Auto Gabi

portal/components/statCard.tsx
```diff
@@ -35,9 +37,7 @@ export const EarnCard = ({
...
           )}
         </p>
-        {!isLoading && !isError && badge && (
-          <div className="self-start empty:hidden">{badge}</div>
-        )}
+        {badge}
```
gabmontes: All seven call sites now repeat a "no badge on error" guard — `isError ? undefined : …`, or the equivalent `isUnavailable && data === undefined` re-derived inline in `totalStaked`, `hemiPriceCard` and `supplySliceCard`. Since the value already renders `-` in that state, a badge is never wanted next to it, so `{!isError && badge}` here would let those guards be dropped and stop a future caller pairing a badge with `-`.

— Gabi 🤖

portal/components/statCard.tsx
ArturDolzan: Good catch, the skeleton stayed at 28px while the value moved to 24px. Fixed with `h-6` in [`e2d93c12`](https://github.com/hemilabs/ui-monorepo/pull/2347/commits/e2d93c12).

portal/components/statCard.tsx
ArturDolzan: Makes sense, moved it into the card in [`e2d93c12`](https://github.com/hemilabs/ui-monorepo/pull/2347/commits/e2d93c12).

Only three call sites could drop it though: in `averageLockup`, `supplySliceCard` and `hemiPriceCard` the check is a null guard for `data.locksCount` / `data.share` / `data.change`, so TypeScript keeps it there.

portal/components/statCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+            '-'
+          ) : (
+            <Skeleton className="h-6 w-20" />
+          )}
+        </p>
+        {!isError && (
+          <span className="flex items-end [&>*]:flex [&>*]:items-end">
+            {badge}
+          </span>
+        )}
```
gabmontes: This wrapper is rendered even when there's no badge (`walletsStaking`, `earnedAmount`, `poolInfoCards`, `walletClaimable` with no rewards, `FromPoolsBadge` returning `null`). With `flex-wrap`, an empty item still takes part in line breaking: once the value comes within 8px of the row's width, the empty span wraps to a second line and `gap-2` adds 8px of height. Checked in a browser — a 150px row is 24px tall, at 138px it's 32px with the empty span and 24px without it. The removed `EarnCard` handled this with `empty:hidden`; adding it here would keep that behaviour.

— Auto Gabi

portal/app/[locale]/hemi-earn/_components/stakedBalance.tsx
```diff
@@ -1,25 +1,26 @@
...
   const { status } = useAccount()
   const t = useTranslations('hemi-earn')
   const isDisconnected = !walletIsConnected(status)
+  const hasError = isError || isDisconnected
 
   return (
-    <EarnCard
-      badge={<FromPoolsBadge />}
+    <StatCard
+      badge={hasError || isPending ? undefined : <FromPoolsBadge />}
```
Comment (gabmontes) [inline portal/app/[locale]/hemi-earn/_components/stakedBalance.tsx]: Now that `StatCard` hides the badge on `isError`, `hasError ||` isn't needed here — `isPending ? undefined : <FromPoolsBadge />` does the same thing.

— Gabi 🤖

portal/components/statCard.tsx
ArturDolzan: Good catch, and that one's on me: before the wrapper there was nothing to lay out when a card had no badge. Added `empty:hidden` in [`b2bd4364`](https://github.com/hemilabs/ui-monorepo/pull/2347/commits/b2bd4364).

Measured it after: the slot now computes to `display: none`, so it's out of the line box rather than just invisible.

portal/app/[locale]/hemi-earn/_components/stakedBalance.tsx
Comment (ArturDolzan) [reply portal/app/[locale]/hemi-earn/_components/stakedBalance.tsx]: Right, the card covers it now. Dropped in [`b2bd4364`](https://github.com/hemilabs/ui-monorepo/pull/2347/commits/b2bd4364).

gabmontes (approved):

gndelia (approved):

